PR #169, opened 2017-09-07
pagination tooltips, nav color adjustment, refactors

benjamincharity:
Changes:

- add tooltips to pagination buttons when active
- lighten the navigation bar background per discussion with shelton
- refactor a few `any`s
- move top level flexbox styles into the navigation component so the consumer doesn't need to set
- minor cleanup


benjamincharity:
Our generated docs make the line count and file count way larger than the changes really dictate. I'm working on a way to get the docs out of our git tracking. But for now since it's hosted with GitHub pages I don't have a way around it.

We really need an S3 bucket where the docs get pushed during the CI build. TBD

PR #189, opened 2017-09-20
#151208103 - QA css class name

coreyterminator:
Added css class names to all existing components to be used by QA/automation for writing tests. Also added documentation about what these classes are and what they are used on, which will appear in the auto generated docs.

ticket: https://www.pivotaltracker.com/story/show/151208103


.gitignore
```diff
@@ -13,3 +13,4 @@ src/lib/helpers.scss
 *.css.map
 e2e/**/*.js
 e2e/**/*.js.map
+yarn.lock
```
benjamincharity: Why are we adding this?

src/lib/src/datepicker/datepicker.component.html
```diff
@@ -9,10 +10,15 @@
     [mdDatepickerFilter]="dateFilter"
     [disabled]="isDisabled"
   >
-  <md-datepicker-toggle mdSuffix [for]="picker"></md-datepicker-toggle>
+  <md-datepicker-toggle 
+    class="qa-datepicker-toggle"
+    mdSuffix 
+    [for]="picker"
+  ></md-datepicker-toggle>
```
benjamincharity: ❤️  the formatting.. so much easier to scan

PR #195, opened 2017-10-04
fix(LoginForm): Verified autofill works for email/password

benjamincharity:
Verified:
screen shot 2017-10-04 at 8 11 55 am


benjamincharity:
@coreyterminator ping for review

PR #559, opened 2018-01-24
Table

benjamincharity:
**TODO:** Generate docs before merge.

### PR Overview

- Create table component
- Create data source class for table
- Create sort directive for column sorting
- Refactor `Pagination` to `Paginator`
- Add docs for
    - table
    - sort
    - minWidth
    - noWrap
- Update `ts-lint` fix with correct glob
- Added more testing utilities (could not import from `@angular/cdk/testing`)
- A `refactor` commit can now trigger a breaking change

---

⚠️ ⚠️ ⚠️ 

This release will include **BREAKING CHANGES** 

- `TsPagination` (rename)
- `TsSelect` (event emitters changed).

⚠️ ⚠️ ⚠️


benjamincharity:
@coreyterminator @atlwendy I'd love some extra eyes on this prior to merge. Especially on the markdown file denoting how to use the table. I want to make sure the table makes sense and is usable for you guys.

> NOTE: I do expect to create a higher-level component that the core team can use. This would have the table/paginator/sorting/etc all hooked up by default. I just would like to get v1 out first so I can move on some other high-priority items.

benjamincharity:
@coreyterminator @atlwendy I'm going ahead with this merge as it's got some items I need back in master. I would still love a review on the usage docs post-merge when you get time 🙏

src/lib/src/paginator/paginator.component.ts
```diff
@@ -53,11 +56,16 @@ import { TsPaginationMenuItem } from './../utilities/interfaces';
...
-  selector: 'ts-pagination',
-  templateUrl: './pagination.component.html',
-  styleUrls: ['./pagination.component.scss'],
+  selector: 'ts-paginator',
+  templateUrl: './paginator.component.html',
+  styleUrls: ['./paginator.component.scss'],
+  host: {
+    class: 'ts-paginator',
+  },
+  changeDetection: ChangeDetectionStrategy.OnPush,
```
atlwendy: Why is here `.OnPush` - CheckOnce not CheckAlways?

src/lib/src/utilities/types/sort.types.ts
```diff
@@ -0,0 +1,8 @@
+/**
+ * Define the allowed sort directions for {@link TsSort}
+ */
+export type TsSortDirection
+  = 'asc'
+  | 'desc'
+  | ''
```
atlwendy: Does `''` mean no sorting?

src/demo/src/app/components/select/select.component.html
```diff
@@ -5,10 +5,9 @@
     [blankChoice]="blank"
     multipleAllowed="false"
     [valueKey]="key"
-    [formControl]="getControl('myChoices1')"
-    (open)="isOpened($event)"
-    (close)="isClosed($event)"
-    (change)="isChanged($event)"
+    [formControl]="myForm.get('myChoices1')"
+    (openedChange)="openStateChange($event)"
+    (selectionChange)="isChanged($event)"
```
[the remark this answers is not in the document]
benjamincharity: Material removed/changed a few of their event emitters to make their API consistent back during their `2.0.0` stable push. We just never picked up those changes until now.

Part of https://github.com/angular/material2/issues/6677

src/demo/src/app/components/table/table.component.ts
```diff
@@ -0,0 +1,242 @@
...
+  ngAfterViewInit() {
+    this.exampleDatabase = new ExampleHttpDao(this.http);
+
+    // If the user changes the sort order, reset back to the first page.
+    this.sort.sortChange.subscribe(() => {
+      this.paginator.currentPageIndex = 0;
+    });
+
+    // Fetch new data anytime the sort is changed, the page is changed, or the records shown per
+    // page is changed
```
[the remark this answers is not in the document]
benjamincharity: If we don't fetch new data, then there is no page change at all. We aren't paging through tons of rows locally; we are getting exactly the page we request from the server. So if the user clicks the pagination to go to the next page, we pass that 'next page' index to the server and ask for those records.

src/lib/src/paginator/paginator.component.ts
benjamincharity: This basically tells Angular to only run change detection when an `@Input` value changes. You are in essence promising to not do anything to mutate the state internally. This is much more performant since the change detection cycle doesn't have to check everything each time.

The few times where we need to do something internally in a component that uses this `changeDetection` strategy we manually tell the change detector to check for changes using `changeDetectorRef.markForCheck()`.

src/lib/src/utilities/types/sort.types.ts
benjamincharity: Yep!

PR #729, opened 2018-03-08
fix(Select): tighten up getValueKey return type and update the example

atlwendy:
Instead of define `public getValueKey(item: object, valueKey?: string)` return type as `string`, be more specific that it can be either `string` or an object of `{ [key: string]: string }`

Also, added valueKey as input in the example, so it's easier for consumer to follow.


src/lib/src/select/select.component.ts
```diff
@@ -103,6 +104,7 @@ export class TsSelectComponent extends TsReactiveFormBaseComponent {
 
   /**
    * Define the key that represents the value item from the object
+   * If not provided, currently it defaults to key "name"
```
benjamincharity: Let's add a line break here.

When our docs are generated these two lines will be combined into a single line unless there is an empty line separating them (which is a good thing for something like lines 130+131).

benjamincharity:
Merging, but one last reminder: When you open the PR you should see an option to 'Connect to Issue' (just above). If you connect it, the issue will automatically move to the review pipeline and this PR will be attached 👍

PR #733, opened 2018-03-09
feat(Tooltip): Create tooltip extension that adds a dotted underline

atlwendy:
Provide an input to tooltip component named `hasUnderline`.
If it's true, dotted underline showing underneath the text.
It's default to false.

### DIFF
--- a/src/lib/src/tooltip/tooltip.component.html
+++ b/src/lib/src/tooltip/tooltip.component.html
@@ -2,6 +2,7 @@
   class="c-tooltip qa-tooltip"
   [matTooltip]="tooltipValue"
   [matTooltipPosition]="tooltipPosition"
+  [ngClass]="{'c-tooltip--underline': hasUnderline}"
 >
   <ng-content></ng-content>
 </span>
--- a/src/lib/src/tooltip/tooltip.component.scss
+++ b/src/lib/src/tooltip/tooltip.component.scss
@@ -1,3 +1,5 @@
+@import './../scss/helpers/cursors';
+
 //
 // @component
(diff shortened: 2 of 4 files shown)


src/lib/src/tooltip/tooltip.component.html
```diff
@@ -2,6 +2,7 @@
   class="c-tooltip qa-tooltip"
   [matTooltip]="tooltipValue"
   [matTooltipPosition]="tooltipPosition"
+  [ngClass]="{'qa-tooltip-dotted-underline': hasUnderline}"
```
benjamincharity: Classes that begin with `qa-` are only for allowing QA to query our DOM elements. We have a separation between classes we use and classes QA uses so that they don't have to update their tests each time we refactor.

Since this is a modifier class we should use the BEM modifier naming format to extend the base `.c-tooltip` class: `.c-tooltip--underlined`

src/lib/src/tooltip/tooltip.component.scss
```diff
@@ -8,3 +8,9 @@
   display: inline-block;
   // Top level styles should be nested here
 }
+
+.qa-tooltip-dotted-underline {
+  border-bottom: .1em dotted;
+  border-color: inherit;
+  cursor: help;
```
benjamincharity: We have a method or mixin for setting CSS cursors. It doesn't do much yet, but it does help us enforce specific cursor usage so we should use it here (don't forget to import the SCSS file).

https://github.com/GetTerminus/terminus-ui/blob/master/src/lib/src/scss/docs/cursor.md

PR #745, opened 2018-03-16
feat(Validator): add lower case validator

atlwendy:
Create a validator to enforce a specific amount of lowercase letters.
And added to the demo


src/lib/src/services/validators/validators/lowercase/lowercase.ts
```diff
@@ -0,0 +1,33 @@
...
+} from '@angular/forms';
+import { createContainsLowercaseRegex } from '@terminus/ngx-tools/regex';
+
+
+/**
+ * Return a validator function to verify that number of lowercase letters is satisfied
+ *
+ * @return The validator function
+ */
+export function lowercaseValidator(digits?: number): ValidatorFn {
```
benjamincharity: Let's rename `digits` to something like `lowercaseMin` or `minimum` etc

src/lib/src/services/validators/validators/lowercase/lowercase.ts
```diff
@@ -0,0 +1,33 @@
...
+    if (!control || !control.value) {
+      return null;
+    }
+
+    digits = digits ? digits : 1
+    const invalidResponse: ValidationErrors = {
+      lowercase: {
+        valid: false,
+        actual: control.value,
+        contains: digits,
```
benjamincharity: To match existing validator objects, we have been using the validator name as they object key that contains the 'rule' or 'limitation'.

Something like:

```javascript
{
  valid: false,
  lowercase: lowercaseMin,
  actual: control.value,
}
```

src/lib/src/services/validators/validators/lowercase/lowercase.spec.ts
```diff
@@ -0,0 +1,55 @@
...
+
+  });
+
+
+  describe(`if the number is missing`, () => {
+
+    test(`should return response as default number 1`, () => {
+      this.validatorFn2 = lowercaseValidator();
+      expect(this.validatorFn2(new FormControl('a'))).toEqual(null);
+    })
```
benjamincharity: Line 52 and 53 are missing semicolons. Let's also add a line break after the `test` block closes.

(I realized last week that things like semicolons aren't being fixed automatically by our linter - I will fix this when I change our toolchain this week)

src/lib/src/services/validation-message/validation-message.service.ts
```diff
@@ -46,6 +46,7 @@ export class TsValidationMessageService {
       noResults: `No results found.`,
       url: `'${validatorValue.actual}' is not a valid URL.`,
       equalToControl: `'${validatorValue.actual}' is not equal to '${validatorValue.compareValue}'`,
+      lowercase: `Must contain at least ${validatorValue.contains} lower case letters`,
```
benjamincharity: It looks like both `lower case` and `lowercase` are technically valid spellings. But for consistencies sake, we should change this to `lowercase` to match the phrasing elsewhere in the codebase.

src/lib/src/services/validators/validators/lowercase/lowercase.spec.ts
```diff
@@ -0,0 +1,55 @@
...
+      const val = 'abCD';
+      expect(this.validatorFn(new FormControl(val))).toEqual(null);
+    });
+
+  });
+
+  describe(`if string length is incorrect`, () => {
+
+    test(`should return error`, () => {
+      const val = 'abxCD';
```
benjamincharity: This will pass validation (causing test to fail). `aBC` would correctly fail.

src/lib/src/services/validators/validators.service.mock.ts
```diff
@@ -73,6 +73,16 @@ export class TsValidatorsServiceMock {
     });
   }
 
+  lowercase = () => {
+    return jest.fn().mockReturnValue(this.isValid ? null : {
+      lowercase: {
+        valid: false,
+        contains: 2,
+        actual: 6,
```
benjamincharity: a) `actual` should be a string of letters since it is the fake version of what might be typed into the input. (eg `actual: 'FOo'`)
b) As mentioned in another comment below: the key `contains` should be the validator name. So in this case it would be `lowercase: 2,`.

benjamincharity:
Also, don't forget to attach this PR to an issue using the ZenHub integration just above the comment box 👍

atlwendy:
@benjamincharity interestingly I did the "connect PR to issue" through zenHub integration last Friday, but for some reason it didn't show up. This time it did.

PR #813, opened 2018-04-04
810 package dependencies

benjamincharity:
- core no longer needs to install/import Material or CDK (including theme styles)
- lib no longer bundling any deps
- refactor button classes addition/removal
- add npm command to explore sourcemap
- all modules export `*` from the component to include any types/interfaces
- bump ngx-tools
- all material imports now reference full path (eg `@angular/material/checkbox`)
- all interfaces/types now live in their associated module


benjamincharity:
@atlwendy just wanted you aware of these changes - all interfaces/types have been moved into the associated module files.

So for example, rather than `TsPaginatorMenuItem` living in `src/utilities/interfaces/paginator.interface.ts` it is now in `src/paginator/paginator.component.ts`

PR #828, opened 2018-04-09
feat(Validator): add uppercase validator

atlwendy:


terminus-ui/src/services/validators/validators/uppercase/uppercase.ts
```diff
@@ -0,0 +1,33 @@
...
+  ValidatorFn,
+  AbstractControl,
+  ValidationErrors,
+} from '@angular/forms';
+import { createContainsUppercaseRegex } from '@terminus/ngx-tools/regex';
+
+
+/**
+ * Return a validator function to verify that number of uppercase letters is satisfied
+ *
```
benjamincharity: Missing JSDoc `@param`

benjamincharity:
@atlwendy I think  a couple test cases may have been swapped around. One is failing that shouldn't, and one isn't that should.

atlwendy:
@benjamincharity I don't know how that happened. I checked the tests before commit and was sure they were good. maybe it's because of `out-tsc` directory issue? Anyway, I removed that directory and should be good now. Please re review.

PR #898, opened 2018-05-18
fix(RadioGroup): hotfix error caused when consumer doesn't use a formControl

benjamincharity:


terminus-ui/src/radio-group/radio-group.component.ts
```diff
@@ -285,10 +285,13 @@ export class TsRadioGroupComponent extends TsReactiveFormBaseComponent implement
    * Update the change detector if the control value changes
    */
   public ngOnInit(): void {
-    this.formControl.valueChanges.subscribe((v: any) => {
-      this.writeValue(v);
-      this.changeDetectorRef.markForCheck();
-    });
+    // istanbul ignore else
+    if (this.formControl) {
+      this.formControl.valueChanges.subscribe((v: any) => {
```
bmalinconico: this needs an unsubscribe fyi

benjamincharity: Ahh thanks

PR #970, opened 2018-07-10
fix(RadioGroup): disabled radio button should not be clickable

atlwendy:


benjamincharity:
@atlwendy looks like we need to add a test for the method that was adjusted.

Issue #974, opened 2018-07-12
small option in radio group component


atlwendy:
Regardless of setting it either `small: true` or `small: false`, DOM treats it as `small`. The only way to have it normal is to remove `small` setting.

PR #975, opened 2018-07-12
fix(RadioGroup): small option true/false reflects DOM visual

atlwendy:
It closes issue #974

### DIFF
--- a/terminus-ui/src/radio-group/radio-group.component.ts
+++ b/terminus-ui/src/radio-group/radio-group.component.ts
@@ -277,7 +277,13 @@ export class TsRadioGroupComponent extends TsReactiveFormBaseComponent implement
    * Define if the visual style should be large or small
    */
   @Input()
-  public small: boolean = false;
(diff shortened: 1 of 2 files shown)


terminus-ui/src/radio-group/radio-group.component.ts
```diff
@@ -277,7 +277,14 @@ export class TsRadioGroupComponent extends TsReactiveFormBaseComponent implement
    * Define if the visual style should be large or small
    */
   @Input()
-  public small: boolean = false;
+  public set small(value: boolean) {
+    this._small = coerceBooleanProperty(value);
+    this.value = this._small;
```
benjamincharity: We don't want to set the value here. `this.value` is the underlying model for any form components where we implement our own `ControlValueAccessor`.

benjamincharity: And to the point you made earlier today about adding integration tests here next... they would have caught this 😺

Issue #1010, opened 2018-08-01
Select: Material select doesn't change when underlying form control is updated


benjamincharity:
Need to trigger change detection.

PR #1011, opened 2018-08-01
fix(Select): trigger change detection when the formControl changes

benjamincharity:
ISSUES CLOSED: #1010

### DIFF
--- a/terminus-ui/src/select/select.component.ts
+++ b/terminus-ui/src/select/select.component.ts
@@ -1,12 +1,14 @@
 import {
-  Component,
-  Input,
-  Output,
-  EventEmitter,
-  forwardRef,
   ChangeDetectionStrategy,
+  Component,
+  EventEmitter,
+  Input,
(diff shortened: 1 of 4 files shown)


terminus-ui/src/select/select.component.ts
```diff
@@ -204,6 +206,26 @@ export class TsSelectComponent extends TsReactiveFormBaseComponent {
...
+
+
+  /**
+   * Trigger change detection when the underlying form changes
+   */
+  public ngOnInit(): void {
+    // istanbul ignore else
+    if (this.formControl) {
+      this.formControl.valueChanges.subscribe((v) => {
+        this.changeDetectorRef.detectChanges();
```
bmalinconico: unsubscribe?

benjamincharity: Dammit. Knew I was going to forget something. Thanks for the catch.. fix incoming

PR #1038, opened 2018-08-23
972 datepicker manual entry validation

### LINKED ISSUE #453: Input: Add usage docs around ngModel vs formControl
Currently it's very unclear if you should use both or part of each etc. We should clearly define when it's useful to use `ngModel` and when it's not needed. (And why we have to handle reactive forms differently from standard Angular)

---

The bindings are both two way. The _only_ reason we have to pass the `formControl` is because we are using a library on another library. So the actual DOM input element is so far nested down inside multiple components that we have no way for the consumer (you) to attach your form control to that actual `input` element. So the UI library needs you to explicitly pass the control to the UI component.

If we removed all our libraries and simply used angular, it would be as simple as assigning the formControlName:

```
<form novalidate [formGroup]="myGroup">
  Name: <input type="text" formControlName="name">
  Location: <input type="text" formControlName="location">
</form>
```

### LINKED ISSUE #506: Input: Incorrect div gets focus during click
The selector `mat-input-infix mat-form-field-infix` seems to get focus when clicking is certain areas.  

- does angular material vanilla do this?

### LINKED ISSUE #917: DateRange: Input values are not validated when manually typed
Currently the component will not allow the user to select an invalid calendar date. However it is still possible for a user to manually enter an invalid date range.

We need to trigger group validation on the blur event for the actual date inputs.

---

## TODO

1. Start date selected or start date blur event:
    1. check for existing validators on endDateCtrl
    1. if none exist, add minDate validator and trigger validation
1. End date selected or end date blur event:
    1. check for existing validators on startDateCtrl
    1. if none exist, add maxDate validator and trigger validation

### LINKED ISSUE #921: Input: icon aligment is off for input prefix & suffix
34258204-eb10e22c-e62a-11e7-9b61-7b640135ec1e

### LINKED ISSUE #922: Input: Input validation is separate from control validation
Notice that according to the form control, this input value is invalid. But the input itself is still styled as though it's valid.

Should also handle #453 and #921 in this ticket (if rewriting).

---

- https://github.com/angular/material2/blob/master/src/lib/input/input.ts
- https://github.com/angular/material2/blob/master/src/lib/form-field/form-field.ts
- https://github.com/angular/material2/blob/master/src/cdk/text-field/autofill.ts
- Possible solution for #385, #352, #656:
  - https://github.com/angular/material2/blob/master/src/cdk/text-field/text-field.md

### LINKED ISSUE #925: Input: Fix possible overlap of hint and validation errors
Interesting note: The Angular library seems to be set up to only show one at a time:
https://github.com/angular/material2/blob/master/src/lib/form-field/form-field.ts#L344-L348

We should move them from absolute positioning to use flexbox so they stack on smaller screens.

### LINKED ISSUE #972: DatePicker: Manual input is not validated
Currently the dates are validated when selected from the calendar but the actual inputs allow any characters with no validation.

pasted_image_at_2018_07_11_01_21_pm

- Restrict to numbers & forward slashes
- ~May need to add a 'date' mask to the input masking directive.~ Tracked in #1015
- Validate on blur.

### LINKED ISSUE #1015: Mask: Need a date format mask
Currently there is no guidance for users manually entering a date into a TsDatepicker instance. We should create a date format input mask to give this guidance to users.

`__/__/____`

### LINKED ISSUE #1024: Input: Rebuild Epic
Primary needs:

- Unified validation
- Works with ngModel & formControl
- Has masks built in

---

Verify that pagination select lays out well after refactor. (it's using legacy styles)

benjamincharity:
Closes #1024
Closes #1015 
Fixes #453 
Fixes #506 
Fixes #917 
Fixes #921 
Fixes #922 
Fixes #925 
Fixes #972


benjamincharity:
Moving to a release branch: https://github.com/GetTerminus/terminus-ui/pull/1048

PR #1045, opened 2018-09-05
feat(fileUpload): Add ControlValueAccessor and image ratio validation

### LINKED ISSUE #1041: FileUpload: Update to use form control correctly
- [ ] Implement control value accessor like other form components.
- [ ] Set validations on form control
- [ ] Allow consumer to pass in `required` validator

atlwendy:
It closes #1041


terminus-ui/src/file-upload/file-upload.component.ts
```diff
@@ -51,12 +53,22 @@ import { TsDropProtectionService } from './drop-protection.service';
...
  * Unique ID for each instance
  */
 let nextUniqueId = 0;
 
+/**
+ * Custom control value accessor for our component.
+ * This allows our custom components to access the underlying form validation via our base class
+ */
+/* tslint:disable:no-use-before-declare */
+const CUSTOM_INPUT_CONTROL_VALUE_ACCESSOR: any = {
```
benjamincharity: We should change this name to better reflect where it's used.

terminus-ui/src/file-upload/file-upload.component.ts
```diff
@@ -233,17 +251,17 @@ export class TsFileUploadComponent implements OnInit , OnChanges, OnDestroy, Aft
   }
   private _acceptedTypes: TsFileAcceptedMimeTypes[] = TS_ACCEPTED_MIME_TYPES.slice();
 
-  /**
-   * Create a form control to manage validation messages
-   */
+  // /**
+  //  * Create a form control to manage validation messages
+  //  */
```
benjamincharity: Wonky comments

terminus-ui/src/file-upload/file-upload.component.ts
```diff
@@ -319,13 +350,18 @@ export class TsFileUploadComponent implements OnInit , OnChanges, OnDestroy, Aft
     this._seedFile = file;
 
     if (file) {
-      const newFile = new TsSelectedFile(file, this.dimensionConstraints, this.acceptedTypes, this.maximumKilobytesPerFile);
+      const newFile = new TsSelectedFile(
+        file,
+        this.dimensionConstraints,
+        this.acceptedTypes,
+        this.maximumKilobytesPerFile,
+        this.ratioConstraints);
```
benjamincharity: Style: let's drop the closing `)` down a line and add a trailing comma. It can be very helpful for scanning when you see the indentation come back out to the original level.

terminus-ui/src/file-upload/selected-file.ts
```diff
@@ -211,4 +217,19 @@ export class TsSelectedFile {
     return false;
   }
 
+  private validateImageRatio(constraints: Array<number> | undefined): boolean {
```
benjamincharity: I'm afraid we are going to have to make this a bit more complex.. Ratio's are most commonly defined as two numbers separated by a `:` eg `3:4`. So I think we'll need to support an array of strings rather than numbers, then use the `:` to split the string.

Which means we'll have to add some logic to convert the image dimensions into a ratio. I'm guessing we should probably round the ratios also. It looks like LinkedIn needs 2 decimal places so we can just go with that for now.

benjamincharity:
Oh, and it looks like our coverage went down a bit

terminus-ui/src/file-upload/file-upload.component.ts
```diff
@@ -676,8 +771,29 @@ export class TsFileUploadComponent implements OnInit , OnChanges, OnDestroy, Aft
...
+    }));
+    return parsedImageRatio;
+  }
+
+  private parseRatioToString(ratios: Array<ImageRatio> | undefined): Array<string> | undefined {
+    if (!ratios) {
+      return;
+    }
+    const parsedRatio: Array<string> = [];
+    ratios.map((r) => parsedRatio.push(r.widthRatio.toString() + ' : ' + r.heightRatio.toString()));
```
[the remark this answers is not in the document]
atlwendy: Haha, I checked both and purposely chose to add space. ❌

PR #1113, opened 2018-11-08
fix(TsInput): Add a check to ensure not to try to run change detection

### LINKED ISSUE #1095: Input: Possible for CDR to fire after component is destroyed
Core reporting this. When toggling an input via `ngIf`, the change detector can be fired after the component is destroyed. We need to manually track the destruction:

```
    // Note that in some cases this can end up being called after the component is destroyed.
    // Add a check to ensure that we don't try to run change detection on a destroyed view.
    if (!this._componentDestroyed) {
      this._changeDetectorRef.detectChanges();
    }
```

atlwendy:
Add a check to ensure that not to try to run change detection on a destroyed view.

close #1095

### DIFF
--- a/package.json
+++ b/package.json
@@ -140,7 +140,7 @@
     "@angular/platform-browser-dynamic": "^6.1.7",
     "@angular/router": "^6.1.7",
     "@terminus/ngx-tools": "^3.0.0",
-    "@terminus/ui": "^8.24.0",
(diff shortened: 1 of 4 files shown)


benjamincharity:
@atlwendy looks good, if you can, just rebase from master so that this branch is up to date (updating from the PR will add another commit)

Issue #1049, opened 2018-09-10
ValidationMessages: Verify built in validators show error messages


benjamincharity:
Currently, it seems that at least some built in validators are not showing error messages via the validation messages component. `Validators.min(15)` is an example of a built in validator that does not show a message.

- [ ] Add error messages for all [built in validators](https://angular.io/api/forms/Validators)

---

#### Info

`TsValidationMessages` gets passed a `FormControl`. It [loops through the form control errors](https://github.com/GetTerminus/terminus-ui/blob/master/terminus-ui/src/validation-messages/validation-messages.component.ts#L49) to get the correct message from the `TsValidationMessageService` which uses the [method `getValidatorErrorMessage` to return the appropriate error message](https://github.com/GetTerminus/terminus-ui/blob/master/terminus-ui/src/validation-messages/validation-message.service.ts#L40). 

Currently all of our custom validators are supported, but validators built into Angular don't have messages defined. We need to make sure that there are appropriate error messages defined for when a built in validator is used.

All built in validators: https://angular.io/api/forms/Validators

We can add tests using a test component like:

```typescript
import { FormControl, Validators } from '@angular/forms';

@Component({
  template: `
    <ts-validation-messages
      validateImmediately="true" // this will make the control validate without needing to be touched first
      [formControl]="myControl"
    ></ts-validation-messages>
  `,
})
export class AttrDisabled {
  myControl = new FormControl(null, Validators.required);
}
```

Then we can query the dom for the message and 'expect' a specific string to be the content.

PR #1125, opened 2018-11-28
fix(ValidationMessages): support for built-in Angular validators and …

shani-terminus:
…deprecation messages for duplic

deprecation schedule for 11.x

ISSUES CLOSED: #1049

### DIFF
--- a/terminus-ui/src/validation-messages/validation-message.service.ts
+++ b/terminus-ui/src/validation-messages/validation-message.service.ts
@@ -36,13 +36,20 @@ export class TsValidationMessageService {
    * @param validatorName - The name of the validator
    * @param validatorValue - The value of the validator
    * @return The error message
+   *
    */
   public getValidatorErrorMessage(validatorName: string, validatorValue?: any): string {
-    const config: {[key: string]: string} = {
+
+    const config: { [key: string]: string } = {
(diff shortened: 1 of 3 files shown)


terminus-ui/src/validation-messages/validation-message.service.ts
```diff
@@ -36,13 +36,20 @@ export class TsValidationMessageService {
    * @param validatorName - The name of the validator
    * @param validatorValue - The value of the validator
    * @return The error message
+   *
+   * @deprecated target 11.x
```
benjamincharity: We should move this deprecation notice to comments just above the items being deprecated (in this case, specific validators). This currently seems like we will be deprecating this entire method.

terminus-ui/src/validation-messages/validation-messages.component.ts
```diff
@@ -48,7 +48,7 @@ export class TsValidationMessagesComponent {
    */
   public get validationMessage(): string | null | undefined {
     if (this.control && this.control.errors) {
-      for (const propertyName in this.control.errors) {
+       for (const propertyName in this.control.errors) {
```
benjamincharity: Extra space indentation

Issue #1046, opened 2018-09-06
Loading Overlay: When used with a TsTable inside a TsCard the overlay is not full width


benjamincharity:
There is a CSS hack to force the nested table to display full-bleed in a card (by overriding padding/margin). The loading overlay layout styles need to take this into account:

![image.png](https://images.zenhubusercontent.com/59fa074d8a75884b9085aa07/e214bf5e-8354-4c79-b2e6-a8d2df10a6c9)

- The style hack should live in the card's scss rather than the overlay's

PR #1130, opened 2018-12-04
fix(Card): overlay bleeds to edges on a table in a card

shani-terminus:
ISSUES CLOSED: #1046

### DIFF
--- a/terminus-ui/src/card/card.component.scss
+++ b/terminus-ui/src/card/card.component.scss
@@ -158,3 +158,12 @@
   }
 }
 
+// Hack for managing overlay when a table is nested in a card
+.c-card__inner {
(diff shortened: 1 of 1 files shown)


terminus-ui/src/card/card.component.scss
```diff
@@ -146,3 +146,12 @@
   }
 }
 
+// Hack for managing overlay when a table is nested in a card
+.c-card__inner {
+  .ts-table ~ .ts-loading-overlay {
+    .c-loading-overlay {
+      left: -24px;
```
benjamincharity: Let's use our spacing mixin for these sizes. 24px would be `spacing(large)`

PR #1171, opened 2018-12-17
699 validator in range

### LINKED ISSUE #699: Validator: isInRange
Basically a combination of `min` and `max` validators.

- Should accept two numbers (min/max)
- Add validation message to the validation message service

shani-terminus:
Closes #699


terminus-ui/src/validators/validators/isInRange/isInRange.ts
```diff
@@ -0,0 +1,57 @@
...
+ * @param maximum - The minimum value
+ * @return The validator function
+ */
+export function isInRangeValidator(
+  minimum: number | AbstractControl = 0,
+  maximum: number | AbstractControl = 0):
+   ValidatorFn {
+  return (control: AbstractControl): ValidationErrors | null => {
+    // Allow optional controls by not validating empty values
+    if (!control || isNaN(control.value)) {
```
benjamincharity: We should use the `ngx-tools` `isNumber` function here for our check. This will allow us to work with a string number (eg `'2'`) as well as a true number (eg `2`). If it passes the check, then we can just call `Number` on it:

```
// Verify the control
if (!control) {
  return null;
}

const value: number | null = isNumber(control.value) ? Number(control.value) : null;

// Verify the value
if (value === null) {
  return null;
}
```

terminus-ui/src/validators/validators/isInRange/isInRange.ts
```diff
@@ -0,0 +1,57 @@
+import {
+  ValidatorFn,
+  AbstractControl,
+  ValidationErrors,
+} from '@angular/forms';
+import { coerceNumberProperty } from '@terminus/ngx-tools/coercion';
+import { isAbstractControl } from '../../../utilities/type-coercion/is-abstract-control';
```
benjamincharity: Let's add a line break between the 3rd party imports and local imports

benjamincharity:
Replacing with #1225

Issue #1210, opened 2019-01-08
Validator: inCollection and isInRange validators were added in 9.x and needs to be ported to 10.x


benjamincharity:
https://github.com/GetTerminus/terminus-ui/blob/8786366efe/terminus-ui/src/validators/validators/inCollection/inCollection.ts

PR #1223, opened 2019-01-15
chore(Validators): isInRange and inCollection added to 10.x

shani-terminus:
package.json dependency on terminus-ui was also updated to pick up the latest version

ISSUES CLOSED: #1210


shani-terminus:
The code for inCollection was already in 10, however, we saw errors in the demo because it was running on an older version of the library. Updating package.json fixed that, and isInRange was added.

Issue #1174, opened 2018-12-19
CSV: Should have ability to fill width with a single column


dpeet:
I am using the CSV Component.  By default it looks like this
![image](https://user-images.githubusercontent.com/4729779/50234759-65343f00-0384-11e9-8b32-f62428e21fd8.png)

I want the Component to fill the available space like this
![image](https://user-images.githubusercontent.com/4729779/50234803-79783c00-0384-11e9-8e62-b7d3f00f2001.png)

I used the below code to create that effect
```scss
.c-csv-entry .c-csv-entry__row {
  display: block !important;
}

.c-csv-entry__input {
  width: 100%;
}

ts-csv-entry {
  display: block;
  width: 100%;
} 
```

---

- [ ] Add `@Input` to force full width
  - `singleColumnMode: boolean`
  - Default to `false`

dpeet:
I think a single column mode would make sense.  Right now it's only being used in my repo, but this input matches well to the excel spreadsheets marketers are used to working in, and is useful for when we need that type of input

benjamincharity:
👍 I'll turn this ticket into a feature request for that ability via a simple `@Input`.

dpeet:
https://media.giphy.com/media/FxPMNutVvNrBm/giphy.gif

ahhh this is awesome :D

benjamincharity:
You should just need quotes around `true`. The demo has a working example with this code:

```html
  <div>
    <ts-csv-entry
      [columnValidators]="validators"
      [columnHeaders]="['Full Width']"
      (blobGenerated)="file($event)"
      fullWidth="true"
      columnCount=1
    ></ts-csv-entry>
  </div>
```

dpeet:
Made that fix and it still isn't full width

dpeet:
Also, should this example have quotes as well?

```
<ts-csv-entry
  id="my-id"
  maxRows="1000"
  columnCount="6"
  rowCount="12"
  fullWidth=false
  [columnValidators]="arrayOfValidators"
  (blobGenerated)="handleTheFileBlob($event)"
></ts-csv-entry>
```

benjamincharity:
Yes. That is a documentation mistake. Thanks for calling that out.

All data attributes should have quotes. It may be hit or miss on what the browser does when it sees an unquoted attribute, but it's not 'valid' so there is no guarantee it would work.

dpeet:
[...]

PR #1239, opened 2019-01-17
fix(CSVEntry): fullwidth works within flex layout = row

shani-terminus:
ISSUES CLOSED: #1174

### DIFF
--- a/terminus-ui/src/csv-entry/csv-entry.component.ts
+++ b/terminus-ui/src/csv-entry/csv-entry.component.ts
@@ -73,8 +73,10 @@ let nextUniqueId = 0;
  *              maxRows="1000"
  *              columnCount="6"
  *              rowCount="12"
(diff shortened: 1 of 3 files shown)


terminus-ui/src/csv-entry/csv-entry.component.ts
```diff
@@ -73,7 +73,6 @@ let nextUniqueId = 0;
  *              maxRows="1000"
  *              columnCount="6"
  *              rowCount="12"
- *              fullWidth=false
```
benjamincharity: IMO we should not remove this. Up till now, the primary JSDoc for a component has every possible input/output regardless of the default value.

shani-terminus: Added it back in, as well as other inputs that were not captured.

Issue #1139, opened 2018-12-11
Navigation: Update navigation item's font weight


dpeet:
Change from `300` to `500`.

Core is currently overriding this. Once updated we should let the core team know the override is no longer needed.

NOTE: This does *not* need to be back-ported to legacy versions.

shani-terminus:
![screen shot 2019-01-23 at 11 36 28 am](https://user-images.githubusercontent.com/44702601/51622590-e3947b80-1f04-11e9-84e3-32f66e59d69e.png)

shani-terminus:
This went out with #1240

PR #1247, opened 2019-01-22
feat(Navigation): navigation supports truncating welcome message and …

shani-terminus:
…name

- truncated welcome message and name appear in tooltip
- length is configurable
- minor style changes per #1139

### DIFF
--- a/terminus-ui/src/navigation/navigation.component.ts
+++ b/terminus-ui/src/navigation/navigation.component.ts
@@ -16,6 +16,7 @@ import {
 } from '@angular/core';
 import { BehaviorSubject } from 'rxjs';
 import { groupBy } from '@terminus/ngx-tools';
+import { TsTooltipPositionTypes } from '../tooltip/tooltip.module';
 
 
 /**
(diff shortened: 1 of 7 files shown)


terminus-ui/src/navigation/navigation.component.ts
```diff
@@ -217,11 +222,53 @@ export class TsNavigationComponent implements OnInit, AfterViewInit {
...
+   * Define the welcome message tooltip position
+   */
+  @Input()
+  public welcomeMsgTooltipPosition: TsTooltipPositionTypes = 'below';
+
+  /**
+   * Getter to return truncated user name
+   */
+  public get truncatedUserName(): string | undefined {
+    if (this.usersFullName && this.usersFullName.length > this.userNameLength) {
```
benjamincharity: We should probably tell our test coverage that we don't care about the 'else' path here:

```
  // istanbul ignore else
  if (...)
```

Same for the `if` in `truncatedWelcomeMsg` also

shani-terminus:
screen shot 2019-01-23 at 11 35 55 am
screen shot 2019-01-23 at 11 36 13 am
screen shot 2019-01-23 at 11 36 04 am

PR #1258, opened 2019-01-26
feat(Tabs): created a Tab component with demo

baigfaizan95:


benjamincharity:
🎉 Thanks for the effort here! 

We're currently heads down for the next few days getting v11 out as it's blocking a few items. But once that drops I'll circle back here to review 👍

PR #1274, opened 2019-01-29
fix(Select): Use formatUIFn function input for seeding autocomplete values

### LINKED ISSUE #1269: Select component cannot display selected object if it's not string

atlwendy:
closes #1269

### DIFF
--- a/terminus-ui/src/select/select.component.ts
+++ b/terminus-ui/src/select/select.component.ts
@@ -28,6 +28,7 @@ import {
 } from '@terminus/ngx-tools/coercion';
 import {
   hasRequiredControl,
+  isFunction,
   isString,
   TsDocumentService,
   untilComponentDestroyed,
@@ -222,6 +223,7 @@ let nextUniqueId = 0;
  *              placeholder="My placeholder!"
  *              showProgress="true"
  *              [sortComparator]="myComparator"
+ *              [chipFormatUIFn]="myUIFormatter"
  *              tabIndex="-1"
  *              theme="primary"
  *              validateOnChange="true"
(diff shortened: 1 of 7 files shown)


terminus-ui/src/select/select.component.ts
```diff
@@ -31,6 +31,7 @@ import {
   isString,
   TsDocumentService,
   untilComponentDestroyed,
+  isFunction,
```
benjamincharity: We've started keeping our imports somewhat alphabetical. Currently we are alphabetizing import collections like the above, but not every separate import statement.

This may make it easier: https://marketplace.visualstudio.com/items?itemName=Tyriar.sort-lines

terminus-ui/src/select/select.component.ts
```diff
@@ -685,6 +687,30 @@ export class TsSelectComponent implements
   }
   private _delimiter: string = DEFAULT_DELIMITER;
 
+  /**
+   * Define a function to retrieve the UI value for an option
+   */
+  @Input()
+  public set formatUIFn(value: TsSelectFormatFn) {
```
benjamincharity: Can we rename this to something more explicit? Maybe `chipFormatFn` or `chipFormatUIFn` etc just so it's clear this will only affect chips and not all selects.

terminus-ui/src/select/select.component.ts
```diff
@@ -954,6 +980,15 @@ export class TsSelectComponent implements
         this.autocompleteFormControl.setValue(this.ngControl.value);
         this.autocompleteSelections = this.ngControl.value;
       }
+      if (this.ngControl.valueChanges) {
+        this.ngControl.valueChanges
+          .pipe(untilComponentDestroyed(this))
+          .subscribe((newValue) => {
+            if (this.autocomplete && newValue) {
```
benjamincharity: we can add `// istanbul ignore else` above this if so coverage will ignore the non-existent 'else' path

PR #1276, opened 2019-01-29
feat(Table): support table column cell alignment

edwin-terminus:


terminus-ui/table/src/cell.ts
```diff
@@ -127,4 +152,10 @@ export class TsColumnDefDirective extends CdkColumnDef {
    */
   @Input()
   public minWidth: string | undefined;
+
+  /**
+   * Define an alignment type for the cell.
+   */
+  @Input()
+  public alignment: TsTableColumnAlignment | undefined;
```
bmalinconico: Consider setting a default alignment instead of allowing undefined. This removes the need to check if it is present all over.

edwin-terminus: Sounds good.  Either way should work.  Setting a default also means that all legacy code will now receive an alignment="left" inside tsCell.

edwin-terminus:
@benjamincharity @bmalinconico The last commit should address your concerns, please review and give me a thumbs up if all looks good.  thanks.

PR #1311, opened 2019-01-31
fix(Select): v10 bug fixes moved to master

benjamincharity:
ISSUES CLOSED: #1310

### DIFF
--- a/terminus-ui/select/src/select.component.spec.ts
+++ b/terminus-ui/select/src/select.component.spec.ts
@@ -43,6 +43,7 @@ import {
   getAllOptionInstances,
   getAutocompleteInput,
   getChipElement,
+  getChipElementDisplayValue,
   getFilterInputElement,
   getOptgroupElement,
   getOptionElement,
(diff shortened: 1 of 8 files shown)


terminus-ui/select/src/select.component.spec.ts
```diff
@@ -1012,6 +1056,26 @@ describe(`TsSelectComponent`, function() {
           expect(instance.autocompleteFormControl.value).toEqual(['Arkansas']);
         });
 
+
+        test(`should set single value 2`, () => {
```
shani-terminus: I was confused by "should set single value 2" when I read it in isolation. It does make sense within the suite. I tend to err toward more clarity though, so I'd be inclined to call the test "should set single value with a second selection" or something along the lines indicating a reset of sorts.

terminus-ui/select/src/select.component.spec.ts
```diff
@@ -1012,6 +1056,26 @@ describe(`TsSelectComponent`, function() {
...
+          const fixture = createComponent(testComponents.SeededAutocompleteWithFormatFn);
+          fixture.componentInstance.allowMultiple = false;
+          fixture.componentInstance.keepOpen = false;
+          fixture.componentInstance.myCtrl.setValue([{ name: 'Florida', population: '20.27M'}]);
+          fixture.detectChanges();
```
shani-terminus: Add an expect here to make sure it was selected correctly and the reset changes the value.

PR #1333, opened 2019-02-05
fix(Test): fix boolean coercion stuff. add isDisabled to card compone…

deanterm:
- Tests are now green with no ugly yellow warnings about boolean coercion! woooooo
- Added `isDisabled` to the card in preparation for removing `disabled` input.

### DIFF
--- a/demo/app/components/card/card.component.html
+++ b/demo/app/components/card/card.component.html
@@ -56,8 +56,8 @@
 
 
 <ts-card
-  supportsInteraction="true"
-  centeredContent="true"
+  [supportsInteraction]="true"
+  [centeredContent]="true"
   [utilityMenuTemplate]="myTemplate"
   aspectRatio="16:9"
   tsVerticalSpacing
@@ -67,7 +67,7 @@
 
 
 
-<ts-card centeredContent="true" tsVerticalSpacing>
+<ts-card [centeredContent]="true" tsVerticalSpacing>
   Centered content
 </ts-card>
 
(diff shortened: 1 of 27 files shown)


demo/app/components/card/card.component.html
```diff
@@ -108,12 +108,12 @@ <h3 tsCardTitle>Card with Title & aspect</h3>
   <h3 tsCardTitle tsTitleAccentBorder>Card with Title & accent border</h3>
 </ts-card>
 
-<ts-card disabled="true" supportsInteraction="true" tsVerticalSpacing>
+<ts-card [disabled]="true" [supportsInteraction]="true" tsVerticalSpacing>
```
shani-terminus: These three (lines 111, 116, 132) should also be updated to isDisabled

terminus-ui/card/src/card.component.ts
```diff
@@ -154,6 +154,22 @@ export class TsCardComponent {
   }
   public _disabled = false;
```
shani-terminus: This should also be isDisabled.

Reference in button component: 
https://github.com/GetTerminus/terminus-ui/blob/master/terminus-ui/button/src/button.component.ts#L207-L219

benjamincharity: We can't actually remove the `disabled` input yet since that would be a breaking change. Just updating internals and adding the new input so that apps can begin the upgrade. So for now, both inputs will exist.

shani-terminus: Ah, thank you for clarifying. Carry on Dean!

PR #1335, opened 2019-02-05
chore(FileUpload): strings are type safe

shani-terminus:


terminus-ui/file-upload/src/selected-file.ts
```diff
@@ -203,6 +216,15 @@ export class TsSelectedFile {
     this.fileReader.readAsDataURL(this.file);
   }
 
+  /**
+   * Validate string type for https://github.com/GetTerminus/terminus-ui/issues/1226
+   *
+   * @param result of the file blob
+   * @return true for string
+   */
+  private isString(x: any): x is string {
```
benjamincharity: We should use the version of this function that is in the ngx-tools repo https://github.com/GetTerminus/ngx-tools/blob/master/ngx-tools/src/verify-types/is-string/is-string.ts

PR #1340, opened 2019-02-05
fix(daterange): on initial load, date range isn't getting value and status set properly

### LINKED ISSUE #1339: daterange isn't populating value and status properly on first load

atlwendy:
closes #1339

### DIFF
--- a/terminus-ui/date-range/src/date-range.component.ts
+++ b/terminus-ui/date-range/src/date-range.component.ts
@@ -1,4 +1,5 @@
 import {
+  ChangeDetectorRef,
   Component,
   EventEmitter,
   Input,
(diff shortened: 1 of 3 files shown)


terminus-ui/date-range/src/date-range.component.ts
```diff
@@ -274,6 +278,13 @@ export class TsDateRangeComponent implements OnInit, OnDestroy {
       return;
     }
 
+    this.changeDetectorRef.detectChanges();
+
+    // This is to fix on an initial load, date range value isn't populating correctly.
```
benjamincharity: Can we prefix this comment with `// HACK: etc..`? Using a standard prefix helps call these specific items out (ie `TODO` etc)

terminus-ui/date-range/src/date-range.component.ts
```diff
@@ -7,6 +7,7 @@ import {
   OnInit,
   Output,
   ViewEncapsulation,
+  ChangeDetectorRef,
```
benjamincharity: We've started alphabetizing multi-imports. Looks like there is a vs code plugin that might be able to make it easy

PR #1354, opened 2019-02-12
fix(Select): show required validation message when it's required and not filled on blur

### LINKED ISSUE #1157: Select: Required validation message not appearing when closing without a selection
Repro:

- Create a single select with a required form control
- Click the select to open
- Click elsewhere to close the select
- Notice the select border is red indicating an error state, but the validation message doesn't appear
- Clicking the select _again_ makes the error appear.

selectrequiredbug

---

This may be related to the touched state of the control. We should look into a) FormControl's `markAsTouched`, `markAsDirty`, b) The validation message components `validateOnChange` input.

atlwendy:
closes #1157

### DIFF
--- a/terminus-ui/select/src/select.component.spec.ts
+++ b/terminus-ui/select/src/select.component.spec.ts
@@ -52,6 +52,7 @@ import {
   getSelectInstance,
   getSelectTriggerElement,
   getToggleAllElement,
+  getValidationMessageElement,
 } from '@terminus/ui/select/testing';
 
 import { TsSelectOptionComponent } from './option/option.component';
@@ -1174,6 +1175,33 @@ describe(`TsSelectComponent`, function() {
       jest.runAllTimers();
     });
 
+    describe(`required`, () => {
+
+      test(`should set required if the form control is required`, () => {
+        const fixture = createComponent(testComponents.ValidateOnChange);
+        fixture.detectChanges();
+        const component = getSelectElement(fixture);
+        const selectTrigger = getSelectTriggerElement(fixture);
+        dispatchMouseEvent(selectTrigger, 'click');
+        dispatchMouseEvent(component, 'click');
+        const validationMessage = getValidationMessageElement(fixture);
+
+        expect(validationMessage).toBeTruthy();
+      });
+
+      test(`should set required if the form control is required for autocomplete mode`, () => {
+        const fixture = createComponent(testComponents.AutocompleteRequired);
+        fixture.detectChanges();
+        const component = getAutocompleteInput(fixture);
+
+        const selectTrigger = getSelectTriggerElement(fixture);
+        dispatchMouseEvent(selectTrigger, 'click');
+        dispatchMouseEvent(component, 'click');
+        const validationMessage = getValidationMessageElement(fixture);
+
+        expect(validationMessage).toBeTruthy();
+      });
+    });
 
     describe(`panel`, function() {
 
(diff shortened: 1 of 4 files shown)


terminus-ui/select/src/select.component.spec.ts
```diff
@@ -1174,6 +1174,33 @@ describe(`TsSelectComponent`, function() {
...
+        const component = getSelectElement(fixture);
+        const selectTrigger = getSelectTriggerElement(fixture);
+        dispatchMouseEvent(selectTrigger, 'click');
+        dispatchMouseEvent(component, 'click');
+        const validationMessage = getValidationMessageElement(fixture);
+
+        expect(validationMessage).toBeTruthy();
+      });
+
+      test(`should set required if the form control is required`, () => {
```
shani-terminus: These two tests have the same description (lines 1177 + 1191), so clarify
`should set required if the form control is required with autocomplete`

terminus-ui/select/src/select.component.ts
```diff
@@ -2000,6 +2006,7 @@ export class TsSelectComponent implements
 
     // Mark this control as 'touched' to trigger any validations needed on blur
     this.onTouched();
+    this.updateValueAndValidity();
```
shani-terminus: Not sure about best practice, but would it be worthwhile to call this.updateValueAndValidity() within this.onTouched() method?

terminus-ui/select/testing/src/test-components.ts
```diff
@@ -407,6 +408,29 @@ export class Autocomplete {
...
+        *ngFor="let option of states"
+        [value]="option.slug"
+        [option]="option"
+      >
+        {{ option.foo }}
+      </ts-select-option>
+    </ts-select>
+  `,
+})
+
```
shani-terminus: remove line between component and its export

terminus-ui/select/testing/src/test-helpers.ts
```diff
@@ -96,3 +96,7 @@ export function getFilterInputElement(fixture: ComponentFixture<any>): HTMLInput
   return fixture.debugElement.query(By.css('.ts-select-panel__filter-input .c-input__text')).nativeElement as HTMLInputElement;
 }
 
+export function getValidationMessageElement(fixture: ComponentFixture<any>): HTMLInputElement {
```
shani-terminus: This is something we could leverage in other components! No changes, just noting it.

benjamincharity: @atlwendy, Shani and I chatted through this and for now you can just leave it as is. I'm going to open up a new ticket to create a top level testing endpoint (`@terminus/ui/testing`) to expose these more widely used utilities that cross components.

terminus-ui/select/src/select.component.ts
```diff
@@ -1509,6 +1509,12 @@ export class TsSelectComponent implements
     this.changeDetectorRef.markForCheck();
   }
 
+  private updateValueAndValidity() {
```
benjamincharity: Can we add a JSDOC comment here?

terminus-ui/select/src/select.component.ts
benjamincharity: We actually cannot do this. 😞  The `onTouched` method is something we have to define as part of the CVA implementation. Angular's core forms code will call our `registerOnTouched` method which will overwrite our `onTouched` method with theirs.

atlwendy:
Ah, habit from working at core to click on update branch button :)

PR #1366, opened 2019-02-15
feat(Validation): add domain validator

### LINKED ISSUE #1235: Validator: Domain validator
Currently, core has a need to validate domains but not full URLs in some cases.

Leverage a domain regex from the tools library.

- [ ] Create validator
- [ ] Add to demo
- [ ] Add to usage docs as needed
- [ ] Add to primary readme
- [ ] Add to demo

atlwendy:
closes #1235

### DIFF
--- a/README.md
+++ b/README.md
@@ -165,6 +165,7 @@ for each feature.
 | Name             | Purpose                                                       | Status             |
 |------------------|---------------------------------------------------------------|:------------------:|
 | `creditCard`     | A credit card number must be valid                            | :white_check_mark: |
+| `domain`         | A domain must be valid                                        | :white_check_mark: |
 | `email`          | An email address must be valid                                | :white_check_mark: |
 | `equalToControl` | A control's value must be equal to another control's value    | :white_check_mark: |
 | `greaterThan`    | A number must be greater than another value                   | :white_check_mark: |
--- a/demo/app/components/validation/validation.component.html
+++ b/demo/app/components/validation/validation.component.html
@@ -227,6 +227,35 @@ http:// shouldfail.com
     ></ts-input>
   </ts-card>
 
+  <ts-card tsVerticalSpacing>
+    <h3 tsCardTitle>Domain</h3>
+
+    <div fxLayout="row" [fxLayoutGap]="flexGap">
+<pre fxFlex>
+  // Valid domains:
+  foo.com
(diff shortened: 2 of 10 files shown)


package.json
```diff
@@ -144,7 +144,7 @@
     "@angular/platform-browser": "^7.2.2",
     "@angular/platform-browser-dynamic": "^7.2.2",
     "@angular/router": "^7.2.2",
-    "@terminus/ngx-tools": "latest",
+    "@terminus/ngx-tools": "^5.2.0",
```
shani-terminus: This should be "latest"
I understand their were some issues last week, but we now have a work-around.

benjamincharity: Once it's changed back to `latest` we should be able to run `yarn upgrade @terminus/ngx-tools` to update the lock file so that `5.2.0` is installed

README.md
```diff
@@ -165,6 +165,7 @@ for each feature.
 | Name             | Purpose                                                       | Status             |
 |------------------|---------------------------------------------------------------|:------------------:|
 | `creditCard`     | A credit card number must be valid                            | :white_check_mark: |
+| `domain`         | An domain  must be valid                                      | :white_check_mark: |
```
benjamincharity: Spelling `An` vs `A`. Plus extra space after `domain`

demo/app/components/validation/validation.component.html
```diff
@@ -197,28 +197,28 @@ <h3 tsCardTitle tsVerticalSpacing="small--0">Contains numbers</h3>
...
-// Invalid URLs:
-http://
-http://foo.bar?q=Spaces should be encoded
-foo.com
-rdar://1234
-http://1.1.1.1.1
-http://.www.foo.bar./
-http:// shouldfail.com
-</pre>
+      <pre fxFlex>
```
benjamincharity: We should undo this change.. I know it makes indentation super odd, but the `pre` element will always respect indentation. So this will actually output 3 indentions in the UI and we really want it to be at 0 indentation when output.

terminus-ui/validators/src/validators/domain/domain.ts
```diff
@@ -0,0 +1,31 @@
+import {
+  ValidatorFn,
+  AbstractControl,
+  ValidationErrors,
```
benjamincharity: Let's sort these while we're in here

Issue #1167, opened 2018-12-14
Input: Phone, currency and percentage masks not working correctly


benjamincharity:
CURRENTLY: every other character correctly masks. ie:
- (1
- 11
- (111
- 1111
- (111) 11
- etc

NOTE:
- Phone mask works with `ngModel`, but not with `FormControl`.
- ~If sanitize is turned off, it works perfectly. So~ the issue seems to be with our CVA.

Issue #1371, opened 2019-02-19
Input: Percentage mask loses % symbol when too many characters are typed


benjamincharity:
- use percentage mask
- set sanitize mask to true
- type in `1234.123`
- see the `%` disappear

PR #1372, opened 2019-02-19
fix(Input): sanitized mask value no longer overwriting UI

benjamincharity:
ISSUES CLOSED: #1167 

Note: There is still an outstanding mask issue with percentages: https://github.com/GetTerminus/terminus-ui/issues/1371

### DIFF
--- a/terminus-ui/input/src/input.component.ts
+++ b/terminus-ui/input/src/input.component.ts
@@ -1427,7 +1427,6 @@ export class TsInputComponent implements
     } else {
       const finalValue = !this.maskSanitizeValue ? value : this.cleanValue(value, this.currentMask.unmaskRegex);
       this.onChangeCallback(finalValue);
-      this.inputValueAccessor.value = finalValue;
     }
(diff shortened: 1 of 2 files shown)


terminus-ui/input/src/input.component.ts
```diff
@@ -1427,7 +1427,6 @@ export class TsInputComponent implements
     } else {
       const finalValue = !this.maskSanitizeValue ? value : this.cleanValue(value, this.currentMask.unmaskRegex);
       this.onChangeCallback(finalValue);
-      this.inputValueAccessor.value = finalValue;
```
atlwendy: why would this remove the % symbol?

benjamincharity: From my current understanding, assigning back to the CVA here triggers Angular's internal updates which then overwrite the UI value with the model value.

shani-terminus: Weird

PR #1373, opened 2019-02-19
feat(Validation): use date pipe

### LINKED ISSUE #1279: ValidationMessagesService: Should use our date pipe
Currently the service is using the date-fns API directly. We should convert this to use our date pipe so that any date formatting changes only need to happen in one place.

https://github.com/GetTerminus/terminus-ui/blob/master/terminus-ui/validation-messages/src/validation-messages.service.ts#L89-L97

We should tackle this before the next breaking change.

atlwendy:
closes #1279

### DIFF
--- a/terminus-ui/input/src/input.component.ts
+++ b/terminus-ui/input/src/input.component.ts
@@ -48,11 +48,11 @@ import {
   MAT_DATE_FORMATS,
 } from '@angular/material/core';
 import {
-  format as formatDate,
   isValid as isValidDate,
 } from 'date-fns';
 import { Subject } from 'rxjs';
 import { TS_SPACING } from '@terminus/ui/spacing';
+import { TsDatePipe } from '@terminus/ui/pipes';
 import { TsFormFieldControl } from '@terminus/ui/form-field';
 import {
   inputHasChanged,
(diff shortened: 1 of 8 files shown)


terminus-ui/input/src/input.component.ts
```diff
@@ -1528,7 +1528,7 @@ export class TsInputComponent implements
         let stringifiedDate: string | undefined;
 
         if (this.mask === 'date') {
-          stringifiedDate = this.isValidDateString(value) ? formatDate(value, 'MM-dd-yyyy') : value;
+          stringifiedDate = this.isValidDateString(value) ? new TsDatePipe().transform(value, 'short') : value;
```
benjamincharity: We should be relying on the Angular injector for these. If you add the `TsPipesModule` to the imports for this module, then you can just inject the pipe via the constructor:

```
constructor(private datePipe: TsDatePipe) {}
```

terminus-ui/validation-messages/src/validation-messages.service.ts
```diff
@@ -86,14 +86,15 @@ export class TsValidationMessagesService {
       inCollection: `${validatorValue.actual} is not an accepted item.`,
     };
 
-    // TODO: Convert to use our date pipe. Blocked by https://github.com/GetTerminus/terminus-ui/issues/753
+    const pipeClass: TsDatePipe = new TsDatePipe();
```
benjamincharity: Same note about injection via constructor

PR #1380, opened 2019-02-22
feat(Logo): logo component is now available

shani-terminus:


terminus-ui/logo/src/logo.component.spec.ts
```diff
@@ -0,0 +1,175 @@
...
+  configureTestBedWithoutReset(moduleDefinition);
+
+  beforeEach(() => {
+    // Reset parent component inputs
+    if (hostComponent) {
+      hostComponent.type = '';
+      fixture.detectChanges();
+    }
+
+    fixture = TestBed.createComponent(TestHostComponent);
```
atlwendy: probably want to use `createComponent` test helper that got released in ngx-tools 6.1.0, so you don't have to new up TestBed and all. Could also take a look at input component spec for example.

README.md
```diff
@@ -457,6 +461,7 @@ This project follows the [all-contributors](https://github.com/kentcdodds/all-co
 [icon-usage]:         http://uilibrary-docs.terminus.ninja/master/components/TsIconComponent.html#readme
 [input-usage]:        http://uilibrary-docs.terminus.ninja/master/components/TsInputComponent.html#readme
 [link-usage]:         http://uilibrary-docs.terminus.ninja/master/components/TsLinkComponent.html#readme
+[logo-usage]:         http://uilibrary-docs.terminus.ninja/master/components/TsLogo.html#readme
```
benjamincharity: I believe this link should be to `TsLogoComponent.html#readme`

terminus-ui/icon/src/custom-icons/logo.ts
```diff
@@ -1,6 +1,6 @@
...
-    </svg>
+  <svg height="24" viewBox="0 0 24 24" width="24" xmlns="http://www.w3.org/2000/svg">
```
benjamincharity: Since we are building a separate module, we should move the SVGs into this module. `terminus-ui/logo/src/svgs/my-logo.svg` That way they aren't bundled with the icon module

terminus-ui/icon/src/custom-icons/logo_color.ts
```diff
@@ -1,14 +1,17 @@
 // tslint:disable: max-line-length
```
benjamincharity: We should use hyphens to separate words in filenames rather than underscores.

terminus-ui/icon/src/custom-icons/logo.ts
benjamincharity: I actually see now that there are already logos in the logo module.. are the icon changes related to the logo module?

terminus-ui/logo/src/logo-types/full_account_hub.ts
```diff
@@ -0,0 +1,24 @@
+// tslint:disable: max-line-length
```
benjamincharity: Same comment about using hyphens in file names for the following files

terminus-ui/logo/src/logo.component.md
```diff
@@ -0,0 +1,67 @@
...
+```html
+<ts-logo type="full-solid"></ts-logo>
+<ts-logo type="mark-gradient"></ts-logo>
+```
+
+Search `TS_LOGO_TYPES` to see all allowed types.
+
+## Colors
+
+Solid logos are white by default, but can also be black (Terminus Dark) or gray (#44525E).
```
benjamincharity: I'd prefer to not reference the exact hex color here as there is a high chance it won't get updated if/when the logo color is.

terminus-ui/logo/src/logo.component.md
```diff
@@ -0,0 +1,67 @@
...
+
+```html
+<ts-logo type="full-solid" logoColor="gray"></ts-logo>
+<ts-logo type="mark-solid" logoColor="black"></ts-logo>
+```
+
+Search for `TS_LOGO_COLORS` to see all allowed logo colors.
```
benjamincharity: We should probably include a list of allowed colors here just to make it super easy for people.

terminus-ui/logo/src/logo.component.scss
```diff
@@ -0,0 +1,23 @@
...
+  @include reset;
+
+  // black:
+  &.ts-logo--black svg path {
+    fill: color(pure, dark);
+  }
+
+  // gray :
+  &.ts-logo--gray svg path {
+    fill: #44525e;
```
benjamincharity: Let's create an SCSS variable just inside the top style (`.ts-logo`) with a comment so we know why a custom color is used.

terminus-ui/logo/src/logo.component.spec.ts
```diff
@@ -0,0 +1,175 @@
...
+      fixture.detectChanges();
+
+      expect(window.console.warn).toHaveBeenCalled();
+    });
+
+    test(`should change type to full-account-hub on input`, () => {
+
+      hostComponent.type = 'full-account-hub';
+      fixture.detectChanges();
+      expect(logo2.type).toEqual('full-account-hub');
```
benjamincharity: We should probably be checking that the actual SVG output is what we expect here rather than just the TS type. We can add an ID inside the SVG code to make it easier to determine which SVG is loaded if it helps

terminus-ui/logo/src/logo.component.ts
```diff
@@ -0,0 +1,159 @@
...
+  encapsulation: ViewEncapsulation.None,
+  exportAs: 'tsLogo',
+})
+
+export class TsLogoComponent implements OnInit {
+  /**
+   * Name of the logo types
+   */
+  @Input()
+  public set type(value: TS_LOGO_TYPE | undefined) {
```
benjamincharity: Do we need to accept `undefined` here?

terminus-ui/logo/src/logo.component.ts
```diff
@@ -0,0 +1,159 @@
...
+    return this._type;
+  }
+  private _type: TS_LOGO_TYPE | undefined = 'full-gradient';
+
+
+  /**
+   * Name of the logo colors
+   */
+  @Input()
+  public set logoColor(value: TS_LOGO_COLOR | undefined) {
```
benjamincharity: Same question about needing `undefined`

terminus-ui/logo/src/logo.component.ts
```diff
@@ -0,0 +1,159 @@
...
+
+  public svg: SafeHtml = ``;
+
+
+  constructor(
+    private domSanitizer: DomSanitizer,
+  ) {}
+
+  ngOnInit() {
+    switch (this.type) {
```
benjamincharity: Let's add `public` before this and `:void` as the return type.

I'd also suggest moving this logic into another method and then calling that method from `ngOnInit`. In general we should be trying to keep logic out of lifecycle events.

terminus-ui/icon/src/custom-icons/logo.ts
shani-terminus: The icon changes were updated svg's from UX

terminus-ui/logo/src/logo.component.spec.ts
shani-terminus: Great idea!

terminus-ui/logo/src/logo.component.spec.ts
```diff
@@ -0,0 +1,160 @@
...
+})
+class TestHostComponent {
+  type = '';
+  logoColor = '';
+
+   @ViewChild('one')
+  one!: TsLogoComponent;
+
+   @ViewChild('two')
+  two!: TsLogoComponent;
```
benjamincharity: Some indentation wonkiness here

terminus-ui/logo/src/logo.component.spec.ts
```diff
@@ -0,0 +1,160 @@
...
+
+ describe(`TsLogoComponent`, function() {
+  let fixture: ComponentFixture<TestHostComponent>;
+  let hostComponent: TestHostComponent;
+  let logo1: TsLogoComponent;
+  let logo2: TsLogoComponent;
+
+  beforeEach(() => {
+    fixture = createComponent(TestHostComponent, undefined, [TsLogoModule]);
+    // fixture = TestBed.createComponent(TestHostComponent);
```
benjamincharity: Legacy comment

terminus-ui/logo/src/logo.component.ts
```diff
@@ -0,0 +1,163 @@
...
+        this.svg = this.domSanitizer.bypassSecurityTrustHtml(MARK_SOLID_LOGO);
+        break;
+      default:
+        this._logoColor = 'white';
+        this.svg = this.domSanitizer.bypassSecurityTrustHtml(FULL_GRADIENT_LOGO);
+        break;
+    }
+  }
+
+  public ngOnInit(): void {
```
benjamincharity: Let's move this above `setType`.

In general our order is usually:

```
properties
getters
inputs
outputs
constructor
lifecycle events
public methods
private methods
```

demo/app/app.module.ts
```diff
@@ -55,6 +55,7 @@ import { TsInputModule } from '@terminus/ui/input';
 import { TsLinkModule } from '@terminus/ui/link';
 import { TsLoadingOverlayModule } from '@terminus/ui/loading-overlay';
 import { TsLoginFormModule } from '@terminus/ui/login-form';
+import { TsLogoModule } from '@terminus/ui/logo';
```
atlwendy: I still get `ERROR in demo/app/app.module.ts(58,30): error TS2307: Cannot find module '@terminus/ui/logo'`here.  I think the reason is that logo component hasn't been released so by referencing it from library, it won't find the module. Once it's released, it should be resolved.

shani-terminus: That makes sense! Ben said he's only ever seen it in dev mode.

terminus-ui/logo/src/logo.component.ts
```diff
@@ -0,0 +1,163 @@
...
+
+
+  public svg: SafeHtml = ``;
+
+
+  constructor(
+    private domSanitizer: DomSanitizer,
+  ) {}
+
+  private setType() {
```
benjamincharity: This needs a return type of `: void`

terminus-ui/logo/src/logo.component.ts
benjamincharity: We lost a comment here..

Let's move this above `setType`.

In general our order is usually:

```
properties
getters
inputs
outputs
constructor
lifecycle events
public methods
private methods
````

File: terminus-ui/logo/src/logo.component.ts (hunk 17 above)
Comment (benjamincharity) [reply terminus-ui/logo/src/logo.component.ts]: Also needs a JSDOC comment

File: terminus-ui/logo/src/logo.component.ts [hunk 18]
```diff
@@ -0,0 +1,163 @@
...
+  host: {
+    class: 'ts-logo',
+    '[class.ts-logo--black]': 'logoColor === "black"',
+    '[class.ts-logo--gray]': 'logoColor === "gray"',
+  },
+  changeDetection: ChangeDetectionStrategy.OnPush,
+  encapsulation: ViewEncapsulation.None,
+  exportAs: 'tsLogo',
+})
+
```
benjamincharity: Unneeded line break

benjamincharity:
I accidentally hit `accept` when adding review comments..

Issue #1074, opened 2018-09-19
Move test helpers into the UI library


benjamincharity:
Core has created several test utility helpers. These should be moved into the UI library under the appropriate secondary entry `@terminus/ui/input/testing` etc

- [x] Move helpers from core into UI library.
- [x] Add any needed tests.
- [x] Add to usage docs as needed

PR #1381, opened 2019-02-25
fix(TestHelpers): moved all core test helpers into the library

benjamincharity:
ISSUES CLOSED: #1074 

- Moved test helpers from core into library
- Added JSDoc comments to all helpers
- Added all helpers to the primary component readmes
- Added a couple classes for easier targeting
- Removed unneeded import
- Remove unneeded ID setter

---

#### Replaced core helpers:

`toggleCheckbox` replaces `checkTsCheckbox` and `uncheckTsCheckbox`
`sendInput` replaces `setTsInputValue` and `setTsDatePicker`
`selectStandardRadio`/`selectVisualRadio` replaces `setTsRadioBoxValueByText`
`setDateRangeValues` replaces `setTsDateRangeValue`
`getAllOptionInstances` replaces `getOptions`
`openSelect` replaces `openSelector`
`selectOption` replaces `selectOption`

### DIFF
--- a/terminus-ui/date-range/src/date-range.component.spec.ts
+++ b/terminus-ui/date-range/src/date-range.component.spec.ts
@@ -9,7 +9,6 @@ import { By } from '@angular/platform-browser';
 import { typeInElement } from '@terminus/ngx-tools/testing';
 import * as testComponents from '@terminus/ui/date-range/testing';
 import {
-  getDebugRangeInputs,
   getRangeInputElements,
   getRangeInputInstances,
 } from '@terminus/ui/date-range/testing';
--- a/terminus-ui/date-range/testing/src/test-helpers.ts
+++ b/terminus-ui/date-range/testing/src/test-helpers.ts
@@ -7,10 +7,77 @@ import {
 } from '@angular/forms';
 import { DebugElement } from '@angular/core';
 import { TsInputComponent } from '@terminus/ui/input';
+import { TsDateRangeComponent } from '@terminus/ui/date-range';
 
 
+/**
+ * Get an array of all TsDateRangeComponent debug elements
+ *
+ * @param fixture - The component fixture
+ * @return The array of DebugElements
+ */
+export function getAllDateRangeDebugElements(fixture: ComponentFixture<any>): DebugElement[] {
+  const debugElements = fixture.debugElement.queryAll(By.css('ts-date-range'));
+  if (!debugElements || debugElements.length < 1) {
+    throw new Error(`'getAllDateRangeInstances' found no date range instances`);
+  }
+  return debugElements;
+}
 
+/**
+ * Get all TsDateRangeComponent instances
+ *
+ * @param fixture - The component fixture
+ * @return The array of TsDateRangeComponent instances
+ */
+export function getAllDateRangeInstances(fixture: ComponentFixture<any>): TsDateRangeComponent[] {
+  return getAllDateRangeDebugElements(fixture).map((i) => i.componentInstance);
+}
 
+/**
+ * Get an array of debug elements for TsInputComponents within a single TsDateRangeComponent
+ *
+ * @param fixture - The component fixture
+ * @param index - The index of the TsDateRangeComponent
+ * @return The array of DebugElements
+ */
+export function getDateRangeInputDebugElements(fixture: ComponentFixture<any>, index = 0): DebugElement[] {
+  const debugElement = getAllDateRangeDebugElements(fixture)[index];
+  if (!debugElement) {
+    throw new Error(`'getDebugRangeInputs' found no debug element at '${index}'`);
+  }
+  return debugElement.queryAll(By.css('ts-input'));
+}
+
+/**
+ * Get an array of TsInputComponent instances for a single TsDateRangeComponent
+ *
+ * @param fixture - The component fixture
+ * @param index - The index of the TsDateRangeComponent
+ * @return The array of TsInputComponent instances
+ */
+export function getRangeInputInstances(fixture: ComponentFixture<any>, index = 0): TsInputComponent[] {
+  return getDateRangeInputDebugElements(fixture, index).map((v) => v.componentInstance);
+}
+
+/**
+ * Get an array of HTMLInputElements for a single TsDateRangeComponent
+ *
+ * @param fixture - The component fixture
+ * @param index - The index of the TsDateRangeComponent
+ * @return The array of TsInputComponent instances
+ */
+export function getRangeInputElements(fixture: ComponentFixture<any>, index = 0): HTMLInputElement[] {
+  return getDateRangeInputDebugElements(fixture, index).map((v) => v.componentInstance.inputElement.nativeElement);
+}
+
+/**
+ * Create a date range form group with required controls
+ *
+ * @param startDate - The initial start date
+ * @param endDate - The initial end date
+ * @return The FormGroup
+ */
 export function createDateRangeGroup(startDate: null | Date = null, endDate: null | Date = null): FormGroup {
   const formBuilder = new FormBuilder();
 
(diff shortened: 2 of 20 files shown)


terminus-ui/select/src/select.component.spec.ts
```diff
@@ -52,13 +51,25 @@ import {
...
 
+function createComponent<T>(component: Type<T>): ComponentFixture<T> {
+  const moduleImports = [
+    FormsModule,
+    ReactiveFormsModule,
+    TsSelectModule,
+    NoopAnimationsModule,
+  ];
+
+  return createComponentInner(component, undefined, moduleImports);
```
shani-terminus: Though this is confusing, it's at least contained.

Is there ever a scenario where the "outer" component needs to be used for testing?

terminus-ui/date-range/src/date-range.component.spec.ts
```diff
@@ -9,7 +9,6 @@ import { By } from '@angular/platform-browser';
 import { typeInElement } from '@terminus/ngx-tools/testing';
 import * as testComponents from '@terminus/ui/date-range/testing';
 import {
-  getDebugRangeInputs,
```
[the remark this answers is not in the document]
benjamincharity: Yeah, I called it out in the PR comment - it was just an unused import that the linter automatically pulled out ¯\_(ツ)_/¯

terminus-ui/radio-group/testing/src/test-helpers.ts
```diff
@@ -0,0 +1,111 @@
...
+  if (!debugElement) {
+    throw new Error(`'selectStandardRadio' found no radio group`);
+  }
+  const labels: DebugElement[] = debugElement.queryAll(By.css('.mat-radio-label'));
+
+  labels.forEach((labelDebugEl) => {
+    const labelTextElement = labelDebugEl.query(By.css('.mat-radio-label-content')).nativeElement;
+
+    if (labelTextElement.textContent.trim() === radioValue) {
+      console.log('SELECTING');
```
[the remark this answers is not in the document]
benjamincharity: whoops.. yeah these were not supposed to be left in 🤦‍♂️

terminus-ui/select/testing/src/test-helpers.ts
```diff
@@ -18,80 +26,318 @@ export function createKeydownEvent(key: string, keyCode: number): KeyboardEvent
...
+ * Get all TsSelectOptionComponent instances for a TsSelectComponent
+ *
+ * @param fixture - The component fixture
+ * @param index - The index of the desired TsSelectComponent
+ * @return An array of TsSelectOptionComponents
+ */
+export function getAllOptionInstances(fixture: ComponentFixture<any>, index = 0): TsSelectOptionComponent[] {
+  const instance = getSelectInstance(fixture, index);
+  const options = instance.options.toArray();
+  if (options.length < 1) {
```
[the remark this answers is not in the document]
benjamincharity: Possibly

terminus-ui/select/src/select.component.spec.ts
benjamincharity: This is just a helper to make it easier in the test file. Since every usage of `createComponent` needs the module imports added I just created this helper function so that I didn't need to add the imports to every single test.

It returns the original function return so there is no outer component - it is exactly the same as using the standard `createComponent` except it passes in the needed imports.

benjamincharity:
Addressed comments

terminus-ui/input/src/input.component.md
```diff
@@ -544,15 +544,18 @@ The row count can be dynamically adjusted:
...
+| `getInputInstance`     |
+| `getInputElement`      |
+| `sendInput`            |
 
 
 
 <!-- LINKS -->
 [autocomplete]: https://github.com/getterminus/terminus-ui/blob/master/terminus-ui/src/input/input.component.ts#l64-73
 [input-types]: https://github.com/getterminus/terminus-ui/blob/master/terminus-ui/src/input/input.component.ts#l48-57
+[test-helpers-src]: https://github.com/GetTerminus/terminus-ui/blob/master/terminus-ui/input/testing/src/
```
shani-terminus: should this go into test-helpers.ts?

terminus-ui/radio-group/src/radio-group.component.md
```diff
@@ -212,3 +213,17 @@ items$: Observable<TsRadioOption[]> = of([
...
+Some helpers are exposed to assist with testing. These are imported from `@terminus/ui/radio-group/testing`;
+
+
+| Function                     |
+|------------------------------|
+| `getAllRadioGroupInstances`  |
+| `getRadioGroupInstance`      |
+| `getRadioGroupParentElement` |
+| `selectStandardRadio`        |
+| `selectVisualRadio`          |
```
shani-terminus: for consistency, there should be a link to the test-helpers.ts

terminus-ui/date-range/testing/src/test-helpers.ts
```diff
@@ -7,10 +7,77 @@ import {
...
+    throw new Error(`'getAllDateRangeInstances' found no date range instances`);
+  }
+  return debugElements;
+}
+
+/**
+ * Get all TsDateRangeComponent instances
+ *
+ * @param fixture - The component fixture
+ * @return The array of TsCheckboxComponent instances
```
atlwendy: Should be array of `TsDateRangeComponent` instances?

terminus-ui/date-range/testing/src/test-helpers.ts
```diff
@@ -7,10 +7,77 @@ import {
...
+/**
+ * Get an array of HTMLInputElements for a single TsDateRangeComponent
+ *
+ * @param fixture - The component fixture
+ * @param index - The index of the TsDateRangeComponent
+ * @return The array of TsInputComponent instances
+ */
+export function getRangeInputElements(fixture: ComponentFixture<any>, index = 0): HTMLInputElement[] {
+  return getDateRangeInputDebugElements(fixture, index).map((v) => v.componentInstance.inputElement.nativeElement);
+}
```
atlwendy: I wonder what's the use case here... for a single TsDateRangeComponent, there are two TsInputComponents. When an index passed in, there's just one input returned, why an array is needed here?

terminus-ui/date-range/testing/src/test-helpers.ts
```diff
@@ -27,18 +94,32 @@ export function createDateRangeGroup(startDate: null | Date = null, endDate: nul
...
+/**
+ * Set values for both date range inputs
+ *
+ * @param fixture - The component fixture
+ * @param startDate - The date to set the start input
+ * @param endDate - The date to set the end input
+ * @param index - The select instance
+ * @return The whenStable promise
+ */
+export function setDateRangeValues(fixture: ComponentFixture<any>, startDate: Date, endDate: Date, index = 0): Promise<any> {
```
atlwendy: Interesting it returns Promise here... since it's setting value, does it have to have returned type? Or it's because of async stuff?

benjamincharity: It doesn't _have_ to have a return, but since we can I figured we should. There could be cases where the function is triggering something that isn't instant (many items, animated stuff, timeouts etc) so this way they can chain to the function to make sure they are checking _after_ Angular settles back down:

```typescript
setDateRangeValues(fixture).then(() => {
  expect(...)...
});
```

terminus-ui/date-range/testing/src/test-helpers.ts
benjamincharity: In this case, the index is used for getting a specific `TsDateRangeComponent`. This is to support consumers that may have more than one date range in a view. Once we get the specific date range, then we return an array of both inputs within that component.

PR #1385, opened 2019-02-26
chore(Card): convert card component to integration test

### LINKED ISSUE #1282: Card: Convert to integration test
Currently this is still using the old method of 'newing' up the component class. We should convert this to use the TestBed and fully test it.

atlwendy:
closes #1282

### DIFF
--- a/terminus-ui/card/src/card.component.spec.ts
+++ b/terminus-ui/card/src/card.component.spec.ts
@@ -1,116 +1,180 @@
-import { TsCardComponent } from './card.component';
+import { Component, ViewChild } from '@angular/core';
+import { ComponentFixture } from '@angular/core/testing';
+import { By } from '@angular/platform-browser';
 
+import { createComponent } from '@terminus/ngx-tools/testing';
 
-describe('TsCardComponent', function() {
-  let component: TsCardComponent;
-  let component2: TsCardComponent;
+import { TsCardComponent, TsCardBorderOptions } from './card.component';
+import { TsCardModule } from './card.module';
+import { TsStyleThemeTypes } from '../../utilities/src/public-api';
+
+@Component({
+  template: `
+  <ts-card
+    [isDisabled]="disabled"
+    [flat]="flat"
+    [supportsInteraction]="supportsInteraction"
+    [theme]="theme"
+    [border]="border"
+  >
+  Here is my card!
+  </ts-card>
+  `,
+})
+class TestHostComponent {
+  border: TsCardBorderOptions | undefined;
+  disabled!: boolean;
+  flat!: boolean;
+  supportsInteraction!: boolean;
+  theme: TsStyleThemeTypes | undefined;
+
+  @ViewChild(TsCardComponent)
+  cardComponent!: TsCardComponent;
+}
+
+describe(`TsCardComponent`, function() {
+  let component: TestHostComponent;
+  let fixture: ComponentFixture<TestHostComponent>;
+  let card: HTMLElement;
+  let cardComponent: TsCardComponent;
+  let cardComponent2: TsCardComponent;
 
   beforeEach(() => {
-    component = new TsCardComponent();
-    component2 = new TsCardComponent();
+    fixture = createComponent(TestHostComponent, [], [TsCardModule]);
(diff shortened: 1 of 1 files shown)


terminus-ui/card/src/card.component.spec.ts
```diff
@@ -1,95 +1,139 @@
...
+    });
+    test(`card does not support interaction`, () => {
+      component.supportsInteraction = false;
+      fixture.detectChanges();
+      expect(cardComponent.supportsInteraction).toEqual(false);
+      expect(card.classList).not.toContain('c-card--interaction');
     });
   });
 
+  describe(`centeredContent`, () => {
```
benjamincharity: I wonder if there is a better way to test if this is actually happening.. What do you think of something like this:

Test component setup:
```html
<ts-card [centeredContent]="true" id="outer">
  <!-- inline element so that the container doesn't go full width -->
  <span id="inner">
    Centered content
  </span>
</ts-card>
```

Test:

```typescript
// Get the dimensions and position of both elements
const outerDimensions = document.getElementById('outer').getBoundingClientRect();
const innerDimensions = document.getElementById('inner').getBoundingClientRect();
// Find the space between the right edge of the inner to the right edge of the outer
const spaceOnRight = outerDimensions.right - innerDimensions.right;
const spaceOnLeft = innerDimensions.left;
// Find the difference between the space on either side and force to a positive number
const value = Math.abs(spaceOnLeft - spaceOnRight);

// We give it a buffer of 50 in case it's not exact
expect(value).toBeLessThan(50);
```

Thoughts?

terminus-ui/card/src/card.component.spec.ts
```diff
@@ -1,95 +1,139 @@
...
+    it(`should set and retrieve`, () => {
+      expect(cardComponent.centeredContent).toEqual(false);
+      cardComponent.centeredContent = true;
+      expect(cardComponent.centeredContent).toEqual(true);
+    });
+  });
 
   describe(`set aspectRatio()`, () => {
-
     it(`should convert the string aspect to a percentage`, () => {
```
benjamincharity: I'd like to see this test actually verifying the DOM changes..Not terrible dissimilar from the last comment:

```html
<ts-card [aspectRatio]="myAspect" id="outer">
  Content
</ts-card>
```

Test:

```typescript
// Set first ratio
component.myAspect = '1:1';
// Get the dimensions
let dimensions = document.getElementById('outer').getBoundingClientRect();
// Verify 1 to 1 ratio
expect(dimensions.width).toEqual(dimensions.height);

// Set new ratio
component.myAspect = '16:9';
// Get the new dimensions
dimensions = document.getElementById('outer').getBoundingClientRect();
// Verify widescreen aspect
expect(dimensions.width).toBeGreaterThan(dimensions.height);
```

terminus-ui/card/src/card.component.spec.ts
```diff
@@ -1,95 +1,139 @@
...
-
-    test(`should set isDisabled flag correctly`, () => {
-      component.isDisabled = true;
-      expect(component.isDisabled).toEqual(true);
-    });
-
-  });
-
-
   describe(`theme`, () => {
```
benjamincharity: We should test for the existence of theme classes in the DOM rather than the component property

terminus-ui/card/src/card.component.spec.ts
```diff
@@ -98,8 +142,8 @@ describe('TsCardComponent', function() {
   describe(`border`, () => {
 
     test(`should do nothing if no value is passed in`, () => {
-      component.border = '' as any;
-      expect(component.border).toEqual('none');
+      cardComponent.border = '' as any;
```
benjamincharity: Same here.. test the DOM. The inverse of the following test

terminus-ui/card/src/card.component.spec.ts
atlwendy: Interesting. I'll play with it for a bit. It's probably easier to test `centeredContent` being true than being false.

benjamincharity: Yeah I had to use it in a few places for the select panel math

Issue #1249, opened 2019-01-22
Pipe: Truncation


shani-terminus:
Core needs the ability to truncate long text in a usable way.

- [x] Pipe name: `tsTruncateAt`
- [x] The first param takes a positive number and defaults to 0
- [x] It should do nothing if no string is passed in
- [x] When truncation happens, the ellipses character should be inserted
- [x] If no truncation should happen, the ellipses character should not be added
- [x] When added, the ellipses should count as one of the allowed characters
- [x] Truncation should support `start`, `middle`, `end` options (defining where the ellipses appears in the string)
  - [x] This should default to `end`
  - [x] When set to `middle` and an even length is set, the starting string should receive the extra character (ie `'abcdefgh' | tsTruncateAt:6` -> `abc…gh`)

Example usage:

```
{{ 'Here is my string' | tsTruncateAt:7 }} // Outputs: `Here i…`
{{ 'Here is my string' | tsTruncateAt:8:'middle' }} // Outputs: `Here…ing`
```

lildziner:
@shani-terminus Found a use case that will affect us very soon with the Account List Dynamic Segmentation project. We will need to bring in the Global Filters from Hub into the wizard flow and we may need to restrict the display width. Here's an example of how BF handles it - It's annoying to wait for it to ticker tape be to get the whole name. A tooltip approach would be better but we're open to ideas. We would leverage this in menus with long names - all campaigns and tactics and image files have extremely long names.
![globalfilterstruncation](https://user-images.githubusercontent.com/4960080/51766541-d44e3300-20a8-11e9-8e0f-925232d3c39d.gif)

PR #1389, opened 2019-02-28
feat(Pipes): truncate pipe is available

shani-terminus:
ISSUES CLOSED: #1249

### DIFF
--- a/terminus-ui/navigation/src/navigation.component.ts
+++ b/terminus-ui/navigation/src/navigation.component.ts
@@ -237,25 +237,6 @@ export class TsNavigationComponent implements OnInit, AfterViewInit {
   @Input()
   public welcomeMsgLength: number = 25;
 
-  /**
-   * Getter to return truncated user name
-   */
-  public get truncatedUserName(): string | undefined {
-    // istanbul ignore else
-    if (this.usersFullName && this.usersFullName.length > this.userNameLength) {
-      return this.usersFullName.slice(0, this.userNameLength) + '...';
-    }
-  }
-
-  /**
-   * Getter to return truncated welcome message
-   */
-  public get truncatedWelcomeMsg(): string | undefined {
-    // istanbul ignore else
-    if (this.welcomeMessage.length > this.welcomeMsgLength) {
-      return this.welcomeMessage.slice(0, this.welcomeMsgLength) + '...';
-    }
-  }
 
   /**
    * Element reference for visible list items
--- a/terminus-ui/pipes/src/truncate/truncate.pipe.spec.ts
+++ b/terminus-ui/pipes/src/truncate/truncate.pipe.spec.ts
@@ -1,0 +1,71 @@
+import { TsTruncateAtPipe } from './truncate.pipe';
+
+describe(`TsTruncateAtPipe`, function() {
+  let pipeClass: TsTruncateAtPipe;
+  let pipe: Function;
+  const inputString = 'This is my test string';
+  const ellipses = '\u2026';
+
+  beforeEach(() => {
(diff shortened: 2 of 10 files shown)


terminus-ui/pipes/src/truncate/truncate.pipe.ts
```diff
@@ -0,0 +1,86 @@
...
+          charCountEnd = (charCount - 1) / 2;
+          charCountStart = charCountEnd + 1;
+        } else {
+          charCountEnd = (charCount - 1) / 2;
+          charCountStart = charCountEnd;
+        }
+        newString = value.slice(0, charCountStart) + '...' + value.slice(-(charCountEnd));
+        break;
+      case('end'):
+        newString = value.slice(0, charCount - 1) + '...';
```
benjamincharity: We should be using the ellipses character. It can be done like this:

```
const ellipses = '\u2026';
const start = 'foo';
const end = 'bar';
const myString = `${start}${ellipses}${end}`;
// myString = 'foo…bar'
```

This will also help enforce the character count since the ellipses will be a single character.

terminus-ui/pipes/src/truncate/truncate.pipe.spec.ts
```diff
@@ -0,0 +1,65 @@
+import { TsTruncateAtPipe } from './truncate.pipe';
+
+describe(`TsTruncateAtPipe`, function() {
```
benjamincharity: Once we're using the ellipsis character, we should add a couple expects that verify the length of the string too

terminus-ui/pipes/src/truncate/truncate.pipe.ts
```diff
@@ -0,0 +1,86 @@
+import {
+  Pipe,
+  PipeTransform,
+} from '@angular/core';
+import { isNumber } from '@terminus/ngx-tools';
+
+
+
+/**
+ * Define the accepted string values for the {@link TsTruncateAt} position
```
benjamincharity: These links need to be the full class name `TsTruncateAtPipe` (same for line 19)

terminus-ui/pipes/src/truncate/truncate.pipe.ts
```diff
@@ -0,0 +1,86 @@
...
+    if (!value) {
+      return;
+    }
+
+    // Insure the correct type
+    if (!isNumber(charCount)) {
+      console.warn(`${charCount} is not a number.`);
+    }
+
+    if ((value.length < charCount) || (charCount === 0)) {
```
benjamincharity: We should probably change this to `charCount < 1` instead of just checking for zero. This will protect us against someone passing in a negative number.

terminus-ui/navigation/src/navigation.component.ts
```diff
@@ -243,7 +244,8 @@ export class TsNavigationComponent implements OnInit, AfterViewInit {
   public get truncatedUserName(): string | undefined {
```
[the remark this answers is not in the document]
shani-terminus: Does this mean I don't have to test the truncated values? Is there a way to determine if the pipe was called?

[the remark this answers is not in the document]
benjamincharity: Yeah we don't want to retest the pipe here necessarily.. but if we wanted to verify it was used we could just test the length of the text..


```
myHtmlElement.textContent.trim().length === myExpectedLength
```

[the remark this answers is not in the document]
shani-terminus: This test doesn't use a fixture, only a component, so for now I have the lines of code commented out, with a note. Once we convert the component we should be able to test this better.

PR #1390, opened 2019-03-01
289 expansion panel

### LINKED ISSUE #289: Expansion Panel / Accordion
This component will have two primary use-cases:

a) A stand-alone expansion panel
b) Multiple panels used as an accordion

### Features

- [x] Should allow multiple sections to be opened at the same time
  - [x] Should have a setting to determine if only a single section may be opened
- [x] Should open/close smoothly (animated)
  - Static height should **not** be needed for animation
  - [x] Both the trigger and body height should animate
- [x] Should expose top level expanded/closed classes to allow trigger customization by consumer
- ~Should use transparency to fade the bottom (showing more items exist)~
  - This would require content to be rendered at all times which conflicts with the requirement for deferred loading of panel content.
- [x] Should be able to set a section as open by default with the `expanded` input
- [x] Should be able to disable sections
  - [x] Should still be able to programmatically control disabled triggers & panels
  - [x] Disabled triggers should have the appropriate cursor
- [x] Should defer rendering content until opened when using `ng-template`
  - `tsExpansionPanelContent`
- [x] Multiple expansion panel components can be used together as an accordion
  - `ts-accordion`
- [x] Each trigger should support simple text content
  - [x] Should also support a title and description component within the trigger
  - `<ts-expansion-panel-trigger>`
    - `<ts-expansion-panel-title>`
    - `<ts-expansion-panel-description>`
  - [x] Triggers should have customizable collapsed and expanded heights
    - `collapsedHeight: number`
    - `expandedHeight: number`
  - [x] Trigger should have visible focus state
- [x] Should support an 'action row' at the bottom (this allows for a 'stepper' functionality)
  - `ts-expansion-panel-action-row`
  - [x] Action items are right aligned using flexbox
- [x] Should be able to hide toggle icon (to support 'stepper' functionality)
- [x] Panel events:
  - [x] `opened()`
  - [x] `afterExpand()`
  - [x] `closed()`
  - [x] `afterCollapse()`
  - [x] `destroyed()`
  - [x] `expandedChange(boolean)`
- [x] Accordion events:
  - [x] `destroyed()`
- Accessibility
  - [x] Triggers should have the `role="button"`
  - [x] Triggers should represent expanded state via `aria-expanded="false"`
  - [x] Triggers should have `aria-controls="content-1"` set with the value matching the associated panel's ID
  - [x] Panel visible state should be represented by `aria-hidden="true"` and toggle the value appropriately
  - [x] Panel should have `attr.aria-labelledby` with the ID of the matching trigger
  - [x] Panels should have `role="region"`
- [x] Should support keyboard controls
  - Accordion:
    - [x] `HOME` should focus first trigger
    - [x] `END` should focus last trigger
    - [x] `ENTER|SPACE` should toggle the focused trigger
- [x] Accordion should expose group controls:
  - [x] `openAll()`
  - [x] `closeAll()`
- [x] Panel should expose individual controls:
  - [x] `open()`
  - [x] `close()`
  - [x] `toggle()`

---

### Usage examples

```html
<!-- Stand-alone panel -->
<ts-expansion-panel>
  <ts-expansion-panel-trigger>
    Here is my trigger!
  </ts-expansion-panel-trigger>

  And here is my standard panel content.
</ts-expansion-panel>
```

```html
<!-- Accordion Mode -->
<ts-accordion>
  <ts-expansion-panel>
    <ts-expansion-panel-trigger>
      Here is my trigger!
    </ts-expansion-panel-trigger>

    And here is my standard panel content.
  </ts-expansion-panel>

  <ts-expansion-panel>
    <ts-expansion-panel-trigger>
      Here is second trigger!
    </ts-expansion-panel-trigger>

    And here is another content section.
  </ts-expansion-panel>
</ts-accordion>
```

```html
<!-- Title & Description -->
<ts-expansion-panel>
  <ts-expansion-panel-trigger>
    <ts-expansion-panel-title>
      Panel title
    </ts-expansion-panel-title>

    <ts-expansion-panel-description>
      A brief description of the panel
    </ts-expansion-panel-description>
  </ts-expansion-panel-trigger>

  And here is my standard panel content.
</ts-expansion-panel>
```

```html
<!-- Action Panel -->
<ts-expansion-panel>
  <ts-expansion-panel-trigger>
    My title
  </ts-expansion-panel-trigger>

  And here is my standard panel content.

  <ts-expansion-panel-action-row>
    <button>Previous</button>
    <button>Next</button>
  </ts-expansion-panel-action-row>
</ts-expansion-panel>
```

```html
<!-- Lazy-loaded panel contents -->
<ts-expansion-panel>
  <ts-expansion-panel-trigger>
    My title
  </ts-expansion-panel-trigger>

  <ng-template tsExpansionPanelContent>
    My deferred content
  </ng-template>
</ts-expansion-panel>
```

---

Resources:
- https://material.io/guidelines/components/expansion-panels.html
- https://github.com/angular/material2/releases/tag/2.0.0-beta.7 (no docs yet)
- https://medium.com/@simonb90/hidden-features-of-angular-material-cdk-ee07066d73d1
- https://github.com/angular/material2/tree/master/src/lib/expansion

benjamincharity:
Closes #289 

- Exported select animations that were causing build warning
- Update mid-range utility gray to better reflect material color spacing
- Create expansion panel component
- Create accordion component

Full requirement list: https://github.com/GetTerminus/terminus-ui/issues/289

---

### Examples

paneldemo

customheight

acoordiondemo

disabled

stepperpanel

### DIFF
--- a/terminus-ui/expansion-panel/src/accordion/accordion-base.ts
+++ b/terminus-ui/expansion-panel/src/accordion/accordion-base.ts
@@ -1,0 +1,31 @@
+import { InjectionToken } from '@angular/core';
+import { CdkAccordion } from '@angular/cdk/accordion';
+
+
+/**
+ * Base interface for a {@link TsAccordionComponent}
+ */
+export interface TsAccordionBase extends CdkAccordion {
+  /**
+   * Handle keyboard events coming in from the panel triggers
+   */
+  handleTriggerKeydown: (event: KeyboardEvent) => void;
+
+  /**
+   * Handle focus events on the panel triggers
+   */
+  handleTriggerFocus: (trigger: any) => void;
+
+  /**
+   * Determine if the toggle icon should be hidden
+   */
+  hideToggle: boolean;
+}
+
+
+/**
+ * Token used to provide a {@link TsAccordionComponent} to {@link TsExpansionPanelComponent}.
+ *
+ * Used primarily to avoid circular imports between `TsAccordionComponent` and `TsExpansionPanelComponent`.
+ */
+export const TS_ACCORDION = new InjectionToken<TsAccordionBase>('TS_ACCORDION');
--- a/terminus-ui/expansion-panel/src/accordion/accordion.component.spec.ts
+++ b/terminus-ui/expansion-panel/src/accordion/accordion.component.spec.ts
@@ -1,0 +1,203 @@
+import { Type } from '@angular/core';
+import { ComponentFixture } from '@angular/core/testing';
+import { NoopAnimationsModule } from '@angular/platform-browser/animations';
+import * as testComponents from '@terminus/ui/expansion-panel/testing';
+import {
+  createComponent as createComponentInner,
+  createKeyboardEvent,
+} from '@terminus/ngx-tools/testing';
+import {
+  END,
+  ENTER,
+  HOME,
+  SPACE,
+  A,
+} from '@terminus/ngx-tools/keycodes';
+import {
+  getPanelInstance,
+  togglePanel,
+  getAccordionInstance,
+  getTriggerElement,
+  getAccordionElement,
+} from '@terminus/ui/expansion-panel/testing';
+
+import { TsExpansionPanelModule } from './../expansion-panel.module';
+
+
+function createComponent<T>(component: Type<T>): ComponentFixture<T> {
+  return createComponentInner<T>(component, undefined, [TsExpansionPanelModule, NoopAnimationsModule]);
+}
+
+
+
+describe(`TsAccordionComponent`, function() {
+
+  test(`should be able to enforce a single panel open at a time`, function() {
+    const fixture = createComponent<testComponents.Accordion>(testComponents.Accordion);
+    fixture.detectChanges();
+    const panel1 = getPanelInstance(fixture);
+    const panel2 = getPanelInstance(fixture, 1);
+
+    expect(panel1.expanded).toEqual(false);
+    expect(panel2.expanded).toEqual(false);
+
+    togglePanel(fixture);
+    expect(panel1.expanded).toEqual(true);
+    expect(panel2.expanded).toEqual(false);
+
+    togglePanel(fixture, 1);
+    expect(panel1.expanded).toEqual(false);
+    expect(panel2.expanded).toEqual(true);
+
+    expect.assertions(6);
+  });
+
+
+  test(`should allow multiple panels to be open`, function() {
+    const fixture = createComponent<testComponents.AccordionMulti>(testComponents.AccordionMulti);
+    fixture.detectChanges();
+    const panel1 = getPanelInstance(fixture);
+    const panel2 = getPanelInstance(fixture, 1);
+
+    expect(panel1.expanded).toEqual(false);
+    expect(panel2.expanded).toEqual(false);
+
+    togglePanel(fixture);
+    expect(panel1.expanded).toEqual(true);
+    expect(panel2.expanded).toEqual(false);
+
+    togglePanel(fixture, 1);
+    expect(panel1.expanded).toEqual(true);
+    expect(panel2.expanded).toEqual(true);
+
+    expect.assertions(6);
+  });
+
+
+  test(`should be able to open or close all panels when multi is true`, function() {
+    const fixture = createComponent<testComponents.AccordionMulti>(testComponents.AccordionMulti);
+    fixture.detectChanges();
+    const accordion = getAccordionInstance(fixture);
+    const panel1 = getPanelInstance(fixture);
+    const panel2 = getPanelInstance(fixture, 1);
+
+    expect(panel1.expanded).toEqual(false);
+    expect(panel2.expanded).toEqual(false);
+
+    accordion.openAll();
+    expect(panel1.expanded).toEqual(true);
+    expect(panel2.expanded).toEqual(true);
+
+    accordion.closeAll();
+    expect(panel1.expanded).toEqual(false);
+    expect(panel2.expanded).toEqual(false);
+
+    expect.assertions(6);
+  });
+
+
+  test(`should be able to hide all toggle icons`, function() {
+    const fixture = createComponent<testComponents.HideToggleAccordion>(testComponents.HideToggleAccordion);
+    fixture.detectChanges();
+    const triggerElement1: HTMLElement = getTriggerElement(fixture);
+    const toggleIcon1 = triggerElement1.querySelector('.ts-expansion-panel__indicator');
+
+    const triggerElement2: HTMLElement = getTriggerElement(fixture);
+    const toggleIcon2 = triggerElement2.querySelector('.ts-expansion-panel__indicator');
+
+    expect(toggleIcon1).toBeNull();
+    expect(toggleIcon2).toBeNull();
+  });
+
+
+  describe(`Keyboard controls`, function() {
+
+    test(`should focus first trigger when HOME is used and the last when END is used`, function() {
+      const fixture = createComponent<testComponents.Accordion>(testComponents.Accordion);
+      fixture.detectChanges();
+      const accordion = getAccordionElement(fixture);
+      const trigger1 = getTriggerElement(fixture);
+      const trigger2 = getTriggerElement(fixture, 1);
+
+      expect(document.activeElement === trigger1).toEqual(false);
+      expect(document.activeElement === trigger2).toEqual(false);
+
+      // Non-special key to complete coverage
+      const keyEvent = createKeyboardEvent('keydown', A, trigger1);
+      trigger1.dispatchEvent(keyEvent);
+      fixture.detectChanges();
+
+      const homeEvent = createKeyboardEvent('keydown', HOME, trigger1);
+      trigger1.dispatchEvent(homeEvent);
+      fixture.detectChanges();
+
+      expect(document.activeElement === trigger1).toEqual(true);
+      expect(document.activeElement === trigger2).toEqual(false);
+
+      const endEvent = createKeyboardEvent('keydown', END, trigger1);
+      trigger1.dispatchEvent(endEvent);
+      fixture.detectChanges();
+
+      expect(document.activeElement === trigger1).toEqual(false);
+      expect(document.activeElement === trigger2).toEqual(true);
+    });
+
+
+    // NOTE: ngx-tools `createKeyboardEvent` function seems to generate an event
+    // where `event.metaKey` is true which does not work for this test.
+    test(`should toggle a panel when SPACE or ENTER is used`, function() {
+      const fixture = createComponent<testComponents.Accordion>(testComponents.Accordion);
+      fixture.detectChanges();
+      const accordion = getAccordionElement(fixture);
+
+      const panel1 = getPanelInstance(fixture);
+      const trigger1 = getTriggerElement(fixture);
+
+      const panel2 = getPanelInstance(fixture, 1);
+      const trigger2 = getTriggerElement(fixture, 1);
+
+      const spaceEvent = document.createEvent('KeyboardEvent');
+      spaceEvent.initEvent('keydown', true, false);
+      Object.defineProperties(spaceEvent, {
+        keyCode: { get: () => SPACE },
+        key: { get: () => 'Space' },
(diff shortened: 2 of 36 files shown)


terminus-ui/expansion-panel/src/accordion/accordion-base.ts
```diff
@@ -0,0 +1,31 @@
+import { InjectionToken } from '@angular/core';
+import { CdkAccordion } from '@angular/cdk/accordion';
+
+
+/**
+ * Base interface for a {@link TsAccordionComponent}
+ */
+export interface TsAccordionBase extends CdkAccordion {
```
atlwendy: why didn't you use `MatAccordionBase`?

terminus-ui/expansion-panel/src/expansion-panel.component.ts
```diff
@@ -0,0 +1,322 @@
...
+    this.inputChanges.next(changes);
+  }
+
+
+  /**
+   * Destroy the parent and finalize any subscriptions
+   */
+  public ngOnDestroy(): void {
+    super.ngOnDestroy();
+    this.inputChanges.complete();
```
atlwendy: Would animation has weird behavior if we don't call `this.bodyAnimationDone.complete()`?

terminus-ui/tsconfig.spec.json
```diff
@@ -7,6 +7,7 @@
     "emitDecoratorMetadata": true,
     "experimentalDecorators": true,
     "types": [
+      "jsdom",
```
atlwendy: w/o specifying this, what do we lose?

terminus-ui/expansion-panel/src/accordion/accordion.component.ts
```diff
@@ -0,0 +1,141 @@
...
+   * The event emitted as the accordion is destroyed
+   */
+  @Output()
+  public destroyed: EventEmitter<void> = new EventEmitter();
+
+
+  /**
+   * Initialize the key manager
+   */
+  public ngAfterContentInit(): void {
```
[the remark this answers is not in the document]
benjamincharity: I only need what's passed into my view to be checked at this point (looking for the trigger component) where after view init will actually let all internal content be rendered and the internal lifecycle hooks called. 

So really I just don't need to wait for _all_ internals.. just the top level view that is passed in.

terminus-ui/expansion-panel/src/accordion/accordion-base.ts
```diff
@@ -0,0 +1,31 @@
...
+  hideToggle: boolean;
+}
+
+
+/**
+ * Token used to provide a {@link TsAccordionComponent} to {@link TsExpansionPanelComponent}.
+ *
+ * Used primarily to avoid circular imports between `TsAccordionComponent` and `TsExpansionPanelComponent`.
+ */
+export const TS_ACCORDION = new InjectionToken<TsAccordionBase>('TS_ACCORDION');
```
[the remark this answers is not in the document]
benjamincharity: Yeah I had to do the same thing with select I think. Basically just pulls a base class out that item A and item B can both import. Where otherwise A needs to import B and B needs to import A.

terminus-ui/expansion-panel/src/expansion-panel.component.ts
```diff
@@ -0,0 +1,322 @@
...
+  host: {
+    class: 'ts-expansion-panel',
+    '[class.ts-expansion-panel--expanded]': 'expanded',
+    '[class.ts-expansion-panel--animation-noopable]': 'animationMode === "NoopAnimations"',
+  },
+  providers: [
+    // Provide TsAccordionComponent as undefined to prevent nested expansion panels from registering to the same accordion.
+    {
+      provide: TS_ACCORDION,
+      useValue: undefined,
```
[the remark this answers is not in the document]
benjamincharity: That's actually a great question.. I honestly didn't try without as Material explicitly uses `undefined`.  My gut says that Angular probably has some internal check since the `useValue` key isn't required. So they likely ignore it if it's not set (meaning completely omitted). But in fairness I have not tested that

terminus-ui/expansion-panel/src/accordion/accordion-base.ts
benjamincharity: We aren't using any material for this component (besides some lifted code). Angular Material builds their accordion using the CDK and I just did the same thing.

This follows the same basic pattern we took with the select; extending the CDK directly. The long-term goal is to remove our dependency on Angular Material which will give us greater flexibility as we continue to define our own brand and needs through the library.

terminus-ui/tsconfig.spec.json
benjamincharity: I'm honestly not 100% sure. I'd assume at the very least we should get better typings when dealing with DOM in tests.

There was nothing explicit it was added for, but as I was looking through various github repos and tutorials trying to track down a fix for that odd DOM error I was fighting I saw this and figured it should be a good addition since we're definitely using JSDom and this `tsconfig` is only used for tests.

It didn't solve my issue, but I left it as it probably is correct to include since we are using JSDom.

terminus-ui/expansion-panel/src/expansion-panel.component.ts
benjamincharity: Not to my knowledge. We're using `untilComponentDestroyed` so it should automatically get completed.

I could probably have added the same pipe operator to `inputChanges` instead of calling complete here, but I just never took the time to go look through the base cdk components to see if anyone else was subscribing to this one.

benjamincharity:
Just realized I didn't update the primary README. PR updated.

terminus-ui/expansion-panel/src/trigger/expansion-panel-trigger.component.ts
```diff
@@ -0,0 +1,218 @@
...
+  Inject,
+  Input,
+  OnDestroy,
+  Optional,
+  ViewEncapsulation,
+} from '@angular/core';
+import {
+  EMPTY,
+  merge,
+  Subscription,
```
shani-terminus: Subscription is not being used

terminus-ui/expansion-panel/src/trigger/expansion-panel-trigger.component.ts
```diff
@@ -0,0 +1,218 @@
...
+    return this.panel.expanded;
+  }
+
+  /**
+   * Whether the associated panel is disabled.
+   *
+   * Implemented as a part of `FocusableOption`.
+   */
+  get disabled(): boolean {
+    return this.panel.disabled;
```
shani-terminus: is this the native disabled/different from our isDisabled?

terminus-ui/expansion-panel/testing/src/test-helpers.ts
```diff
@@ -0,0 +1,263 @@
+import { ComponentFixture } from '@angular/core/testing';
+import { By } from '@angular/platform-browser';
+import { DebugElement } from '@angular/core';
+import {
+  TsAccordionComponent,
+  TsExpansionPanelActionRowComponent,
```
shani-terminus: TsExpansionPanelActionRowComponent is not used.

terminus-ui/expansion-panel/testing/src/test-helpers.ts
```diff
@@ -0,0 +1,263 @@
...
+/**
+ * Get the element for a TsExpansionPanelActionRowComponent
+ *
+ * @param fixture - The component fixture
+ * @param index - The index of the desired TsExpansionPanelComponent
+ * @return The element
+ */
+export function getPanelActionRow(fixture: ComponentFixture<any>, index = 0): HTMLElement {
+  const debugElement = getPanelDebugElement(fixture, index);
+  const row = debugElement.query(By.css('.ts-expansion-panel__action-row'));
```
shani-terminus: here's the test helper for TsExpansionPanelActionRowComponent, using the query selector

terminus-ui/expansion-panel/src/accordion/accordion.component.spec.ts
```diff
@@ -0,0 +1,205 @@
+import {
+  createComponent as createComponentInner,
+  dispatchKeyboardEvent,
+  createKeyboardEvent,
+} from '@terminus/ngx-tools/testing';
+import { Type } from '@angular/core';
+import { ComponentFixture } from '@angular/core/testing';
+import { NoopAnimationsModule } from '@angular/platform-browser/animations';
+import * as testComponents from '@terminus/ui/expansion-panel/testing';
```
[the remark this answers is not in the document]
benjamincharity: This is definitely not super clear. `import *` imports _everything_ from the module. So I could technically use the test helper functions from `testComponents.myHelperFunction()`. I find that a bit confusing so I explicitly import my helper functions.

Ideally I need a way to export all test components under a single name without including _everything_ from the testing endpoint.

terminus-ui/expansion-panel/src/expansion-animations.ts
```diff
@@ -0,0 +1,84 @@
...
+  ]),
+
+  /**
+   * Animation that expands and collapses the panel trigger height
+   */
+  expansionTriggerHeight: trigger('expansionHeight', [
+    state('collapsed, void', style({
+      height: '{{collapsedHeight}}',
+    }), {
+      params: {collapsedHeight: '48px'},
```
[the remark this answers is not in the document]
benjamincharity: Yep, exactly!

terminus-ui/expansion-panel/src/expansion-panel.component.spec.ts
```diff
@@ -0,0 +1,321 @@
...
+
+  });
+
+
+
+
+  describe(`Accessibility`, function() {
+
+    test(`should have the correct role and aria labels on triggers`, function() {
+      const fixture = createComponent<testComponents.SinglePanel>(testComponents.SinglePanel);
```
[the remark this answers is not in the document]
benjamincharity: Only the toggle icon can be hidden, so all the trigger aria items would still exist in any situation.

terminus-ui/expansion-panel/src/expansion-panel.component.ts
```diff
@@ -0,0 +1,322 @@
...
+    },
+  ],
+  encapsulation: ViewEncapsulation.None,
```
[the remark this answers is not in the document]
benjamincharity: Ooh yes, very good catch

terminus-ui/expansion-panel/src/expansion-panel.component.ts
```diff
@@ -0,0 +1,322 @@
...
+   * This alias is to conform to our existing naming convention.
+   */
+  @Input()
+  public set isDisabled(value: boolean) {
+    this.disabled = value;
+  }
+  public get isDisabled(): boolean {
+    return this.disabled;
+  }
+
```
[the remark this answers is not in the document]
benjamincharity: The class that we are extending (`CdkAccordionItem`) sets up its own `@Input`s that we are extending. So they have defaults defined there.

This is also why we have `disabled` and `isDisabled`. The CKD implements the input as `disabled` but our library convention is a bit more explicit as `isDisabled`. So I've created inputs that just set the value on the existing input. That way consumers use our API but the code itself continues using what the CDK provides.

terminus-ui/expansion-panel/src/expansion-panel.component.ts
```diff
@@ -0,0 +1,322 @@
...
+   * This alias is to conform to our existing naming convention.
+   */
+  @Input()
+  public set isExpanded(value: boolean) {
+    this.expanded = value;
+  }
+  public get isExpanded(): boolean {
+    return this.expanded;
+  }
+
```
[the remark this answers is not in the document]
benjamincharity: same answer as above

terminus-ui/expansion-panel/src/trigger/expansion-panel-trigger.component.ts
benjamincharity: The CKD implements the input as `disabled` but our library convention is a bit more explicit as `isDisabled`. So I've created inputs that just set the value on the existing input. That way consumers use our API but the code itself continues using what the CDK provides.

Issue #830, opened 2018-04-10
Confirmation: Allow positioning


benjamincharity:
- [ ] Allow position to be set
  - before, after
  - above, below (will be centered)
- [ ] Create scss classes to position arrow of the modal based on allowed positions

<details>
  <summary>Original issue</summary>

  - if the confirmation appears close to the viewport edge, ~padding disappears~ (cannot repro) and the modal spills outside of the viewport.
  - ~the confirmation directive is setting `display: block` on the `ts-button`~
  - confirmation may not be centered if `ts-confirmation` is filling width of parent:
<img width="559" alt="screen shot 2018-05-14 at 7 14 22 am" src="https://user-images.githubusercontent.com/270193/39994204-8f24619a-5746-11e8-9e64-8d811b4be03d.png">
  ~ - Button should not be able to scroll away from popup~
moved to https://github.com/GetTerminus/terminus-ui/issues/1391

~Blocked until Material is upgraded to 6.x: https://material.angular.io/cdk/overlay/overview~

</details>

shani-terminus:
Modal is set to be centered, but it appears relative to the container, not the button. This is set in confirmation.directive.ts `createOverlay()` function

The little triangle is set in CSS (`:before`)

benjamincharity:
@shani-terminus now that #1391 has been split out, is this ticket finished? Or are there outstanding items to check?

shani-terminus:
I have confirmed the modal will be outside of the viewport if the button is too close to the bottom, and noted the modal is centered even if the button is justified. I think this ticket is still relevant but we need to figure out what we want as the solution:
- allow core to set a position? (we'd need both left/right and top/bottom)
- calculate the position of the button relative to the edge of the viewport and adjust automatically?
- something else?

benjamincharity:
> allow core to set a position? (we'd need both left/right and top/bottom)

benjamincharity:
> allow core to set a position? (we'd need both left/right and top/bottom)

This is my vote. Much simpler than dynamic positioning

shani-terminus:
Tooltip position is a good start, but this one is more elaborate as it has to accommodate two positions, and we'll need need new scss rules for the different variations.

PR #1397, opened 2019-03-05
feat(Confirmation): supports positioning

shani-terminus:
ISSUES CLOSED: #830


terminus-ui/confirmation/src/confirmation.directive.ts
```diff
@@ -24,6 +27,26 @@ import { TsButtonComponent } from '@terminus/ui/button';
...
 
 
+/**
+ * Define the accepted string values for the {@link TsConfirmationModalComponent} position
+ */
+export type TsConfirmationModalPositionTypes
+  = 'above'
+  | 'below'
+  | 'before'
+  | 'after'
```
atlwendy: Not sure `left` and `right` are clearer than `before` and `after`?

demo/app/components/confirmation/confirmation.component.html
```diff
@@ -7,6 +7,7 @@
     (cancelled)="cancel($event)"
     confirmationButtonText="Custom Confirmation Button Text"
     [explanationText]="explanation"
+    modalPosition="above"
```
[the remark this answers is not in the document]
benjamincharity: True.. we could add a dropdown or radio buttons to change the side. I think the tooltip demo does this

terminus-ui/confirmation/src/confirmation-modal.component.scss
```diff
@@ -32,9 +32,28 @@
...
+
+  &.ts-confirmation-overlay__panel-below::before {
     @include triangle(top, color(pure), spacing(small, 1));
     left: 50%;
     transform: translateX(-50%);
   }
+
+  &.ts-confirmation-overlay__panel-before::before {
+    @include triangle(right, color(pure), spacing(small, 1));
+    right: -12px;
```
benjamincharity: We can use `right: -#{spacing(small)}` here and on line 55.

The `#{}` forces interpolation of the function before adding the negative prefix. It doesn't come up much, but can be pretty powerful: https://sass-lang.com/documentation/file.SASS_REFERENCE.html#interpolation_

terminus-ui/confirmation/src/confirmation.directive.ts
```diff
@@ -8,12 +8,15 @@ import {
...
   Output,
+  isDevMode,
 } from '@angular/core';
 import {
   ConnectedPositionStrategy,
   Overlay,
   OverlayConfig,
   OverlayRef,
+  HorizontalConnectionPos,
+  VerticalConnectionPos,
```
benjamincharity: Imports sorting here and above

terminus-ui/confirmation/src/confirmation.directive.ts
benjamincharity: Yeah I think both would work. Material actually allows both types of phrasing. Here my vote is to stick with before/after simply because it matches our existing tooltip api.

terminus-ui/confirmation/src/confirmation.directive.ts
```diff
@@ -104,6 +128,23 @@ export class TsConfirmationDirective implements OnDestroy, OnInit {
   }
   private _explanationText: string | undefined;
 
+  /**
+   * Define position of the modal
+   */
+  @Input()
+  public set modalPosition(value: TsConfirmationModalPositionTypes) {
```
benjamincharity: I know you inherited the `modal` verbiage here (thanks, me!), but I'd like to throw out the idea of adjusting from `modal` to `overlay` for everything.

Possible reasons for change:

a) The library we are using under the covers (CDK) uses the term `overlay` so this would align us better with our tools.
b) Modal is a loaded term at Terminus. Prior to this feature, the term modal was only visible internally but now it will be part of the component API.
c) It should probably happen in this PR because once this feature is released, this change would be a breaking change.

---

We should address all instances of `modal` including filenames, types, methods etc. It seems or CSS is already using `overlay`.

```
confirmation-modal.component -> confirmation-overlay.component
TsConfirmationModalPositionTypes -> TsConfirmationOverlayPositionTypes
modalPosition -> overlayPosition
allowedModalPositionTypes -> allowedOverlayPositionTypes
modalPos___ -> overlayPos
```

terminus-ui/confirmation/src/confirmation.directive.ts
```diff
@@ -162,15 +203,51 @@ export class TsConfirmationDirective implements OnDestroy, OnInit {
...
+        modalPosOverlayX = 'start';
+        modalPosOverlayY = 'center';
+        positionClass = 'ts-confirmation-overlay__panel-after';
+        break;
+      default:
+        modalPosOriginY = 'bottom';
+        modalPosOverlayY = 'top';
+        positionClass = 'ts-confirmation-overlay__panel-below';
+        break;
+    }
```
benjamincharity: I'd love to see our config generation happen in another function just to keep concerns separated and methods smaller. Two possible options:

a) The `modalPosition` input setter calls a function each time it is set:

```typescript
set modalPosition(v) {
  this._modalPosition = v;
  this.overlayConfig = this.generateOverlayConfig(v);
}
```

b) The `createOverlay` method calls it directly:

```typescript
createOverlay() {
  const overlayConfig = this.generateOverlayConfig(this.modalPosition);
  ...
```

(in both examples I pass the position to the generate function which keeps our method 'pure')

demo/app/components/confirmation/confirmation.component.html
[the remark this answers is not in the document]
benjamincharity: Side note.. if we add the ability to change the side dynamically, we will probably want to use the same property to adjust the layout of the demo. That way when `after` is selected the button moves to the far left of the screen (using flexbox) so that it's a better representation of when `after` would be used

terminus-ui/confirmation/src/confirmation.directive.ts
shani-terminus: To echo Ben, I chose these options to match tooltip position. This nomenclature is better if we ever have to support RTL languages.

terminus-ui/confirmation/src/confirmation-modal.component.scss
shani-terminus: THANKS!

demo/app/components/confirmation/confirmation.component.html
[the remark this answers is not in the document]
shani-terminus: Great suggestion!

terminus-ui/confirmation/src/confirmation.directive.ts
shani-terminus: I went with option B

Issue #1404, opened 2019-03-08
Add tslint rule for alphabetically sorting imports


benjamincharity:
`"ordered-imports": true`

This will order collections alphabetically:

```
import { foo, bar } from 'baz';

// becomes:
import { bar, foo } from 'baz';

```

And group library imports together:

```
import { foo } from 'baz/foo';
import { bing } from 'bing';
import { bar } from 'baz/bar';

// becomes:
import { bar } from 'baz/bar';
import { foo } from 'baz/foo';
import { bing } from 'bing';
```

PR #1405, opened 2019-03-08
refactor(AllModules): ordered all imports via tslint

benjamincharity:
ISSUES CLOSED: #1404 

NOTE: This touches pretty much every lib file so it will likely cause conflicts if you were already branched off of master. Any conflicts should be fairly easy to fix as the changes are scoped to only file imports.

---

This adds a tslint rule to enforce a) alphabetically sorted imports and b) imports grouped by library. Both of these are 'fixable' rules, so running `yarn run lint` will automatically fix the order (that's what was run to create all the changes in this PR - `tslint.json` was the only manually edited file).

While this doesn't split 3rd party and local imports, it does respect that separation if it already exists.

I _highly_ recommend setting your editor up to run the lint command on save so that you don't need to think about this at all. It will automatically add missed semi-colons, trailing commas, fix spacing issues and more.

VSCode setting for `tslint`: 

```
"editor.codeActionsOnSave": {
    "source.fixAll.tslint": true
}
```

Reference: https://marketplace.visualstudio.com/items?itemName=ms-vscode.vscode-typescript-tslint-plugin

### DIFF
--- a/terminus-ui/button/src/button.component.spec.ts
+++ b/terminus-ui/button/src/button.component.spec.ts
@@ -1,13 +1,13 @@
 import { Component, ViewChild } from '@angular/core';
 import { By } from '@angular/platform-browser';
 import {
-  createMouseEvent,
   createComponent,
+  createMouseEvent,
 } from '@terminus/ngx-tools/testing';
 
+import { ComponentFixture, tick } from '@angular/core/testing';
 import { TsButtonComponent } from './button.component';
 import { TsButtonModule } from './button.module';
-import { ComponentFixture, tick } from '@angular/core/testing';
 
 @Component({
   template: `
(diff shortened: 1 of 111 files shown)


terminus-ui/button/src/button.component.spec.ts
```diff
@@ -1,13 +1,13 @@
 import { Component, ViewChild } from '@angular/core';
 import { By } from '@angular/platform-browser';
 import {
-  createMouseEvent,
   createComponent,
+  createMouseEvent,
 } from '@terminus/ngx-tools/testing';
 
+import { ComponentFixture, tick } from '@angular/core/testing';
```
shani-terminus: Looks like it's prioritizing external imports over relative imports!

terminus-ui/button/src/button.module.ts
```diff
@@ -1,5 +1,5 @@
-import { NgModule } from '@angular/core';
 import { CommonModule } from '@angular/common';
+import { NgModule } from '@angular/core';
```
shani-terminus: I bet it's alphabetizing the source.

Issue #1410, opened 2019-03-11
Add package references to package.json


benjamincharity:
We should be providing paths to various library sources:

```
  "module": "./fesm5/terminus-ui.js",
  "es2015": "./fesm2015/terminus-ui.js",
  "esm5": "./esm5/terminus-ui.js",
  "esm2015": "./esm2015/terminus-ui.js",
  "fesm5": "./fesm5/terminus-ui.js",
  "fesm2015": "./fesm2015/terminus-ui.js",
  "main": "bundles/terminus-ui.umd.js",
  "typings": "terminus-ui.d.ts",
  "sideEffects": false,
```

Reference: https://docs.google.com/document/d/1CZC2rcpxffTDfRDs6p1cfbmKNLA6x5O-NtkJglDaBVs/edit

PR #1411, opened 2019-03-11
fix(Packages): correctly point to source files

benjamincharity:
ISSUES CLOSED: #1410


benjamincharity:
ng-packagr is already handling this for us

PR #1420, opened 2019-03-19
chore(Link): tests converted to integration

shani-terminus:
ISSUES CLOSED: #1285

### DIFF
--- a/terminus-ui/link/src/link.component.spec.ts
+++ b/terminus-ui/link/src/link.component.spec.ts
@@ -1,23 +1,92 @@
+import { APP_BASE_HREF } from '@angular/common';
+import { Component, ViewChild } from '@angular/core';
+import { ComponentFixture, tick } from '@angular/core/testing';
+import { By } from '@angular/platform-browser';
+import { RouterModule } from '@angular/router';
+
+import { createComponent } from '@terminus/ngx-tools/testing';
+
 import { TsLinkComponent } from './link.component';
+import { TsLinkModule } from './link.module';
+
+@Component({
+  template: `
+  <ts-link
+    [destination]="destination"
+    [isExternal]="isExternal"
+    [tabIndex]="tabIndex"
+  >
+  My Link Text
+  </ts-link>
+  `,
+})
+class TestHostComponent {
+  destination!: any;
+  isExternal!: boolean;
+  tabIndex!: number | undefined;
+
+  @ViewChild(TsLinkComponent)
+  linkComponent!: TsLinkComponent;
+}
 
 
 describe(`TsLinkComponent`, function() {
-  let component: TsLinkComponent;
+  let component: TestHostComponent;
+  let fixture: ComponentFixture<TestHostComponent>;
+  let link: HTMLElement;
+  let linkComponent: TsLinkComponent;
 
   beforeEach(() => {
-    component = new TsLinkComponent();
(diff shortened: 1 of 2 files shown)


terminus-ui/link/src/link.component.spec.ts
```diff
@@ -1,23 +1,98 @@
...
 
-  it(`should exist`, () => {
-    expect(component).toBeTruthy();
+  test(`should exist`, () => {
+    expect(linkComponent).toBeTruthy();
+  });
+
+  describe(`isInternal`, () => {
+    test(`should default and retrieve`, () => {
+      link = fixture.debugElement.query(By.css('.c-link')).nativeElement as HTMLElement;
```
atlwendy: Since you're repeating this multiple times, probably good to put in `beforeEach` block.

terminus-ui/link/src/link.component.spec.ts
```diff
@@ -1,23 +1,98 @@
...
+      fixture.detectChanges();
+      component.isExternal = false;
+      component.destination = ['/#'];
+
+      expect(component.isExternal).toBeFalsy();
+      expect(link.classList).toContain('qa-link-internal');
+      expect(component.destination).toEqual(['/#']);
```
atlwendy: Except `expect(link.classList)`, the other two expect clause are not useful since you're setting them a couple of lines above?

terminus-ui/link/src/link.component.spec.ts
```diff
@@ -1,23 +1,98 @@
...
   describe(`isExternal`, () => {
-    it(`should set and retrieve`, () => {
+    test(`should set and retrieve`, () => {
+      component.destination = 'www.google.com';
       component.isExternal = true;
+      fixture.detectChanges();
+      link = fixture.debugElement.query(By.css('.c-link')).nativeElement as HTMLElement;
+
       expect(component.isExternal).toEqual(true);
+      expect(component.destination).toEqual('www.google.com');
```
atlwendy: Same to these two.

terminus-ui/link/src/link.component.spec.ts
```diff
@@ -1,23 +1,98 @@
...
 
+  describe(`tabIndex`, () => {
+    test(`should default to 0 and be set`, () => {
+      expect(link.tabIndex).toEqual(0);
+
+      component.tabIndex = 9;
+      fixture.detectChanges();
+      link = fixture.debugElement.query(By.css('.c-link')).nativeElement as HTMLElement;
+
+      expect(component.tabIndex).toEqual(9);
```
benjamincharity: Another unnecessary `expect`

terminus-ui/link/src/link.component.spec.ts
benjamincharity: Not a change request, but this does exemplify why I started creating test-helpers for my specs.. I was starting to see these super long queries _all_ over my tests so I wrapped each one up in a function for easy re-use.

shani-terminus: I had it in the beforeEach but it was not getting updated properly.

atlwendy: Might take a look at some existing specs. Some reference: https://github.com/GetTerminus/terminus-ui/blob/master/terminus-ui/button/src/button.component.spec.ts#L52-L58

PR #1436, opened 2019-04-01
chore(Accessibility): BREAKING CHANGE - base font weight increased to…

shani-terminus:
… 400


.github/feature.md
```diff
@@ -1,3 +1,13 @@
+<!-- START doctoc generated TOC please keep comment here to allow auto update -->
```
benjamincharity: We should undo this file's changes. I'll try to look into a way to exclude files from our ToC generation today

shani-terminus: In the meantime we'll have to make note

Issue #1233, opened 2019-01-17
All components: Refactor to remove coerceBoolean from all inputs


benjamincharity:
Current, the library coerces all boolean inputs to a true boolean. This allows users to by 'lazy' and simply use data attributes rather than binding:

```html
// Currently, `my-input` would be a boolean of true in the component
<my-component my-input="true"></my-component>
```

While this solves certain issues that arise from using a data-attribute to pass a value in that is not a string, it is working around the Angular frameworks default behavior. In the long run this is a disservice to the application engineers as our library inputs do not function exactly as default Angular inputs do.

We should revert this functionality so that our inputs function as default Angular inputs.

```html
<!-- Currently -->
<my-component my-input="true"></my-component>

<!-- After change -->
<my-component [my-input]="true"></my-component>
```

:boom: Breaking change :boom:

---

- [x] Remove all boolean coercions from inputs
  - If the input has a setter/getter but only uses them for coercing a boolean, we should remove the set/get in favor of the simple input format (example below)
- [x] Verify all tests still pass
- [x] Verify *all* documentation is updated to use the correct binding (`[foo]="bar"` vs `foo="bar"`)
  - [x] JSDoc comments for each component
  - [x] Usage docs for each component
  - [ ] Update any consumers (this should have been done during deprecation, but we should verify)
- [x] Write breaking change migration docs

```typescript
// Simple Angular input
@Input()
myInput: myType = 'foo';
```

PR #1441, opened 2019-04-05
1233 deprecate coerce boolean

shani-terminus:
also cleaning up related imports

💥 BREAKING CHANGE 💥 

Migration notes:
boolean values will have to be entered correctly according to angular in the
html([booleanvar]="true")

```html
<!-- before -->
<my-component my-input="true"></my-component>

<!-- after -->
<my-component [my-input]="true"></my-component>
```

ISSUES CLOSED: #1233

### DIFF
--- a/terminus-ui/icon/src/icon.component.md
+++ b/terminus-ui/icon/src/icon.component.md
@@ -7,6 +7,7 @@
 
 - [Basic usage](#basic-usage)
 - [Theming](#theming)
+- [Background](#background)
 - [Style with CSS](#style-with-css)
 - [Usage inline with text](#usage-inline-with-text)
 - [Custom Icons](#custom-icons)
@@ -42,6 +43,15 @@ Icons support the same themes as the rest of the components:
 ```
 
 Search for `TsStyleThemeTypes` to see all allowed types.
+
+
+## Background
+
+Icons can be shown white with a colored background (color is determined by theme) by setting `background` to true.
+
+```html
+<ts-icon [background]="true"></ts-icon>
+```
 
 
 ## Style with CSS
@@ -90,4 +100,4 @@ Any icon with a -color suffix will not accept themes. Currently they accept a ba
 | `engage`      | A right-pointing arrow stacked on a left-pointing arrow | Navigation for Engage product     |
 | `lightbulb`   | A lightbulb                                             | Pro-tip box                       |
 | `logo`        | Terminus logo, default is black, but accepts theme      | Logo, negative logo               |
-| `logo-color`  | Terminus logo in correct colors, does not accept theme  | Logo like it is supposed to look  |
+| `logo_color`  | Terminus logo in correct colors, does not accept theme  | Logo like it is supposed to look  |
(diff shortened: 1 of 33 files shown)


terminus-ui/icon/src/icon.component.md
```diff
@@ -44,6 +45,15 @@ Icons support the same themes as the rest of the components:
 Search for `TsStyleThemeTypes` to see all allowed types.
 
 
+## Background
+
+Icons can be shown white with a colored background (color is determined by theme) by adding a `background`.
```
benjamincharity: Possibly reword to say `... by setting background to true` for clarity.

terminus-ui/input/src/input.component.spec.ts
```diff
@@ -106,7 +106,7 @@ describe(`TsInputComponent`, function() {
         const component = getInputInstance(fixture);
         const el = component.inputElement.nativeElement;
 
-        expect(component.readOnly).toEqual(false);
+        expect(component.readOnly).toBeFalsy();
```
benjamincharity: Any specific reason these expects were changed to a less strict format?

terminus-ui/select/src/optgroup/optgroup.component.ts
```diff
@@ -58,7 +55,7 @@ let nextUniqueId = 0;
     role: 'group',
     '[class.ts-optgroup--disabled]': 'isDisabled',
     '[attr.id]': 'id',
-    '[attr.aria-disabled]': 'isDisabled.toString()',
+    '[attr.aria-disabled]': 'isDisabled ? "true" : "false"',
```
benjamincharity: What was the reason for this change?

terminus-ui/select/src/option/option.component.ts
```diff
@@ -119,7 +117,7 @@ let nextUniqueId = 0;
     '[class.ts-select-option--template]': 'optionTemplate',
     '[attr.tabindex]': 'tabIndex',
     '[attr.aria-selected]': 'selected.toString()',
-    '[attr.aria-disabled]': 'isDisabled.toString()',
+    '[attr.aria-disabled]': 'isDisabled ? "true" : "false"',
```
benjamincharity: Same question as above

terminus-ui/input/src/input.component.spec.ts
shani-terminus: The tests were failing. I _think_ it said expecting "false" and received "" (empty string)

benjamincharity: Hmm we should probably look deeper into this. We may need to update our tests to correctly use the input. `readOnly` should only be a boolean so it sounds like possibly coercion was handling an incorrect input from our test setup.

terminus-ui/select/src/optgroup/optgroup.component.ts
shani-terminus: Once coercion was removed, the tests failed due to "unable to read toString of undefined."

This format is similar to what's used for disabled in ts-input element:
https://github.com/GetTerminus/terminus-ui/blob/master/terminus-ui/input/src/input.component.html#L96

PR #1479, opened 2019-04-23
chore(testing): use ngx-tools createComponent utility

atlwendy:
Removed all the local defined `createComponent` methods and use ngx-tools utility instead.

### DIFF
--- a/demo/app/app.component.spec.ts
+++ b/demo/app/app.component.spec.ts
@@ -1,25 +1,31 @@
-import { async, TestBed } from '@angular/core/testing';
+import { Component } from '@angular/core';
+import { async, ComponentFixture } from '@angular/core/testing';
 import { RouterTestingModule } from '@angular/router/testing';
+import { createComponent } from '@terminus/ngx-tools/testing/public-api';
 
 import { AppComponent } from './app.component';
 
+@Component({
+  template: `
+  <demo-app></demo-app>
+  `,
+})
+class TestHostComponent {
+}
 
 describe('AppComponent', () => {
 
-  beforeEach(async(() => {
-    TestBed.configureTestingModule({
-      imports: [
-        RouterTestingModule,
-      ],
-      declarations: [
-        AppComponent,
-      ],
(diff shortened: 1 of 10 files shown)


demo/app/app.component.spec.ts
```diff
@@ -1,25 +1,31 @@
-import { async, TestBed } from '@angular/core/testing';
+import { async, TestBed, ComponentFixture } from '@angular/core/testing';
 import { RouterTestingModule } from '@angular/router/testing';
 
 import { AppComponent } from './app.component';
+import { Component } from '@angular/core';
```
shani-terminus: Imports should be alphabetized. You can run `yarn run lint:app:ts`.

terminus-ui/date-range/src/date-range.component.spec.ts
```diff
@@ -239,17 +239,11 @@ describe(`TsDateRangeComponent`, function() {
 
 // TODO: Move to ngx-tools (and all other instances of this utility)
```
shani-terminus: update or remove comment

shani-terminus (changes requested):
Looks good, just a couple things to clean up

demo/app/app.component.spec.ts
benjamincharity: This should also be up with 3rd party imports

terminus-ui/table/src/table.component.spec.ts
```diff
@@ -517,15 +525,5 @@ export function createComponent<T>(component: Type<T>, providers: Provider[] = [
...
-      {
-        provide: TsWindowService,
-        useClass: TsWindowServiceMock,
-      },
-      ...providers,
-    ],
-  }).compileComponents();
-
-  return TestBed.createComponent<T>(component);
+    );
```
benjamincharity: This should be de-indented

benjamincharity (changes requested):

benjamincharity (approved):

PR #1486, opened 2019-04-30
fix(FormField): label now matches input/select id instead of form-field

### LINKED ISSUE #1459: Input: Label currently associated with the form field rather than the HTML input
Our label has the correct aria for attribute but it is currently pointing to the form field rather than the underlying HTML input:

- [ ] The label should reference the input ID.
- [ ] Add test to ensure this behavior.

shani-terminus:
ISSUES CLOSED: #1459

### DIFF
--- a/terminus-ui/form-field/src/form-field.component.spec.ts
+++ b/terminus-ui/form-field/src/form-field.component.spec.ts
@@ -39,7 +39,8 @@ describe(`TsFormFieldComponent`, function() {
       fixture.detectChanges();
       const labelElement = fixture.debugElement.query(By.css('.ts-form-field__label'));
 
-      expect(labelElement.nativeElement.getAttribute('aria-owns')).toEqual('foo');
+      expect(labelElement.nativeElement.getAttribute('for')).toEqual('foo');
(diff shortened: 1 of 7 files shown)


terminus-ui/form-field/src/form-field.component.spec.ts
```diff
@@ -208,6 +209,7 @@ describe(`TsFormFieldComponent`, function() {
 export class Id {
   formControl = new FormControl();
   myId: string | undefined = 'foo';
+  parentId: string = 'bar';
```
benjamincharity: Unnecessary type

terminus-ui/form-field/src/form-field.component.ts
```diff
@@ -233,6 +234,12 @@ export class TsFormFieldComponent implements AfterContentInit, AfterContentCheck
   }
   protected _id: string = this.uid;
 
+  /**
+   * Define the parent's id
+   */
+  @Input()
+  public parentId = this.parentId;
```
benjamincharity: What is this self reference accomplishing?

benjamincharity (changes requested):

Issue #1296, opened 2019-01-30
Tooltip: Convert to integration test


benjamincharity:
Currently this is still using the old method of 'newing' up the component class. We should convert this to use the TestBed and fully test it.

Might have to add a toggle (material has one, we don't) to show/hide the tooltip, even if it's just for testing purposes.

shani-terminus:
Once this has been completed, we should revisit paginator tests to verify the UI for the tooltips.

PR #1514, opened 2019-05-20
chore(Paginator): converted to integration tests

shani-terminus:
closes https://github.com/GetTerminus/terminus-ui/issues/1296

### DIFF
--- a/terminus-ui/paginator/src/paginator.component.spec.ts
+++ b/terminus-ui/paginator/src/paginator.component.spec.ts
@@ -1,470 +1,372 @@
+import { Type } from '@angular/core';
 import {
-  SimpleChange,
-  SimpleChanges,
-} from '@angular/core';
-import { ChangeDetectorRefMock } from '@terminus/ngx-tools/testing';
+  ComponentFixture,
+  fakeAsync,
+  tick,
+} from '@angular/core/testing';
+import { By } from '@angular/platform-browser';
+import { createComponent as createComponentInner } from '@terminus/ngx-tools/testing';
+import * as TestComponents from '../testing/src/test-components';
 import {
-  TsSelectChange, TsSelectComponent,
-} from '@terminus/ui/select';
+  clickToChangePage,
+  updateRecordsPerPage,
+} from '../testing/src/test-helpers';
+import { TsPaginatorMenuItem } from './paginator.component';
+import { TsPaginatorModule } from './paginator.module';
 
-import { TsPaginatorComponent } from './paginator.component';
 
 
 describe(`TsPaginatorComponent`, function() {
-  let component: TsPaginatorComponent;
-  const TOTAL_RECORDS = 100;
-
-  beforeEach(() => {
-    component = new TsPaginatorComponent(
-      new ChangeDetectorRefMock(),
-    );
-  });
-
-
   test(`should exist`, () => {
-    component.totalRecords = 125;
-    component.ngAfterViewInit();
+    const fixture = createComponent(TestComponents.Basic);
+    fixture.detectChanges();
 
-    expect(component).toBeTruthy();
+    expect(fixture.debugElement.query(By.css('.ts-paginator'))).toBeTruthy();
   });
 
   describe(`showRecordsPerPageSelector`, () => {
-    test(`should set and retrieve`, () => {
-      component.showRecordsPerPageSelector = true;
-      expect(component.showRecordsPerPageSelector).toEqual(true);
-    });
-  });
-
-
-  describe(`ngAfterViewInit()`, () => {
-
-    test(`should call the initialization method`, () => {
-      component['initialize'] = jest.fn();
-      component.ngAfterViewInit();
-
-      expect(component['initialize']).toHaveBeenCalled();
-    });
-
-  });
-
-
-  describe(`ngOnChanges()`, () => {
+    let fixture: ComponentFixture<TestComponents.RecordsPerPage>;
+    let hostComponent: TestComponents.RecordsPerPage;
 
     beforeEach(() => {
-      component['initialize'] = jest.fn();
+      fixture = createComponent(TestComponents.RecordsPerPage);
+      hostComponent = fixture.componentInstance;
     });
 
+    test(`should set and retrieve dropdown to choose page`, () => {
+      fixture.detectChanges();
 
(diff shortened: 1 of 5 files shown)


terminus-ui/paginator/src/paginator.component.spec.ts
```diff
@@ -1,470 +1,309 @@
+import { Type } from '@angular/core';
 import {
-  SimpleChange,
-  SimpleChanges,
-} from '@angular/core';
-import { ChangeDetectorRefMock } from '@terminus/ngx-tools/testing';
+  ComponentFixture, fakeAsync, tick,
```
benjamincharity: Let's drop these items onto their own lines. Our linter doesn't quite get it for us yet :/

terminus-ui/paginator/testing/src/test-helpers.ts
```diff
@@ -0,0 +1,37 @@
...
+;
+
+/**
+ * Helper function to click a button to update results set
+ *
+ * @param fixture - fixture
+ * @param dir - direction: first, last, next or previous, to determine which button to click
+ */
+// tslint:disable-next-line no-any
+export function clickForResults(fixture: ComponentFixture<any>, dir: TsPaginatorPage) {
```
benjamincharity: Perhaps there's a more clear name for this helper since it's only focused on clicking a page button. (thinking about consumers quickly understanding why/how they should use this)

terminus-ui/paginator/src/paginator.component.spec.ts
```diff
@@ -1,470 +1,309 @@
...
+      expect(npIndex).toEqual(1);
+    }));
+
+    test(`should go to the previous page`, fakeAsync(() => {
+      const fixture = createComponent(TestComponents.RecordsPerPage);
+      fixture.detectChanges();
+      const dir = 'previous';
+      tick(2000);
+      clickForResults(fixture, 'next');
+      clickForResults(fixture, dir);
```
benjamincharity: Can we assert something in between these two calls? We are kind of changing something, then changing it back, then verifying it never changed. So it seems we could remove both `clickForResults` calls and the test would still pass.

Or possibly just use a TestHost that has the pagination defaulted to something other than then first page.

terminus-ui/paginator/src/paginator.component.spec.ts
```diff
@@ -1,470 +1,309 @@
...
+      const fixture = createComponent(TestComponents.RecordsPerPage);
+      fixture.detectChanges();
+      const dir = 'first';
+      tick(2000);
+      clickForResults(fixture, 'next');
+      clickForResults(fixture, 'next');
+      clickForResults(fixture, dir);
+
+      const titleEl = fixture.debugElement.query(By.css('.qa-paginator-current-page-menu .c-button__content')).nativeElement as HTMLElement;
+      // eslint-disable-next-line max-len
```
benjamincharity: We should only be disabling this rule if there is truly no way around it (ie a really long string).

For items like this we can just drop it down after the `=`:

```typescript
const firstPageBut =
  fixture.debugElement.query(By.css(`.qa-paginator-${dir}-page-button .c-button`)).nativeElement as HTMLButtonElement;
```

Same note for several lines below

terminus-ui/paginator/src/paginator.component.spec.ts
```diff
@@ -1,470 +1,309 @@
...
-      component.totalRecords = 99;
-      component.ngAfterViewInit();
-
-      const actual = component.shouldShowRecordsMessage('foo', 100, 99);
-      const expected = false;
-      expect(actual).toEqual(expected);
-    });
-
-  });
+    test(`should show all results if they fit on a page`, () => {
```
benjamincharity: We should check that buttons get disabled also

benjamincharity (changes requested):
Couple small tweaks.

One note to all of us, we should be striving to not use abbreviated or truncated naming conventions. Not requiring future devs to mentally map meanings is well worth the added characters.

---

It does look like a couple paths were missed simply because they require a specific setup to be triggered:

Screen Shot 2019-05-28 at 8 56 47 AM

For the first missed path, we need a setup with: 
a) more than 1 page of options
b) the length of options should not fit into pages perfectly
c) should not be zero-based

For the second missed path we need a setup with:
a) less records than a single page
b) should not be zero-based

terminus-ui/paginator/src/paginator.component.spec.ts
```diff
@@ -1,470 +1,309 @@
...
+      expect(fixture.debugElement.query(By.css('.qa-paginator-per-page-select'))).toBeTruthy();
     });
 
-  });
-
-
-  describe(`isFirstPage()`, () => {
+    test(`should change page on select`, () => {
+      const fixture = createComponent(TestComponents.RecordsPerPage);
+      const hostComponent = fixture.componentInstance;
```
atlwendy: I wonder whether inside this describe scope, `fixture` and `hostComponent` could be extracted into `beforeEach` block as the exact same assignment happened so many times.

terminus-ui/paginator/src/paginator.component.spec.ts
```diff
@@ -1,470 +1,309 @@
...
-    test(`should return TRUE when the passed in page is 0`, () => {
-      component.totalRecords = 125;
-      component.ngAfterViewInit();
+      const titleEl = fixture.debugElement.query(By.css('.qa-paginator-current-page-menu .c-button__content')).nativeElement as HTMLElement;
+      expect(titleEl.textContent).toContain('1 - 20 of 100');
 
-      expect(component.isFirstPage(0)).toEqual(true);
     });
 
+    test(`should be hidden when set`, () => {
```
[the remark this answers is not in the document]
shani-terminus: I chose to separate them because there are two things required for the drop down to show. Maybe once we remove the logic from the template, we can combine them.
```html
    <ts-select
      class="qa-paginator-per-page-select"
      *ngIf="recordsPerPageChoices && showRecordsPerPageSelector"
```

terminus-ui/paginator/src/paginator.component.spec.ts
shani-terminus: great suggestion!

benjamincharity (approved):

PR #1522, opened 2019-05-28
chore: update contributors

benjamincharity:


atlwendy (approved):
We should write to a script to automate this 😄

PR #1530, opened 2019-06-04
1519 investigate sass true

### LINKED ISSUE #1520: SCSS: Add tests for all helpers
### Is your feature request related to a problem? Please describe.

No, we have a gap in our testing. SCSS testing framework has been defined, use the docs and examples in _color.spec.scss, _a11y.spec.scss, and _spacing.spec.scss to create tests for other functions and mixins.

Tests for all SCSS 

### Describe alternatives you've considered

An understanding and acceptance of the risks involved with not testing code

### LINKED ISSUE #1529: SCSS: Fix incorrect shadow mixin usage
https://github.com/GetTerminus/terminus-ui/blob/master/terminus-ui/scss/helpers/_shadows.scss#L302

Currently the code is:

```scss
@include elevation-transition($elevation2 -$elevation2);
```

The `elevation-transition` mixin only takes a single argument so this is being evaluated as `@include elevation-transition(1 - 1);` which means we are only every passing `0` to this mixin.

Originally library also exhibits the same behavior: https://github.com/prestopasto/SCSS-Material-Shadows

---

- [ ] Figure out what should be passed to the mixin
- [ ] Once fixed, visually smoke-test the UI

shani-terminus:
Also closes #1520 and closes #1529

### DIFF
--- a/terminus-ui/scss/helpers/_color.spec.scss
+++ b/terminus-ui/scss/helpers/_color.spec.scss
@@ -9,63 +9,141 @@
     @include assert-equal($test, $expect);
   }
 
-  @include test('should return the correct theme and color for background-color') {
-    @include assert {
-      @include output {
-        @include theme-color(background-color);
-      }
-      @include expect {
-        @each $key, $value in $theme-and-color {
-          ::ng-deep .u-theme--#{$key} & {
-            background-color: $value;
-
-            &[disabled] {
-              background-color: color(utility, light);
-              color: color(utility);
-            }
-
-            @if (not(($key == 'pure')) and not(($key == 'highlight'))) {
-              color: $color__pure--light;
-            }
-
-            @if ($key == 'highlight') {
-              color: $color__pure--dark;
-            }
-          }
-        }
-      }
-    }
-  }
-
   @include test('should return the correct theme and color for color') {
     @include assert {
       @include output {
         @include theme-color(color);
       }
       @include expect {
-        @each $key, $value in $theme-and-color {
-          ::ng-deep .u-theme--#{$key} & {
-            color: $value;
+        ::ng-deep .u-theme--primary & {
+          color: #00538a;
+        }
+        ::ng-deep .u-theme--primary &[disabled] {
+          background-color: #cecdd1;
+          color: #999;
+        }
+        ::ng-deep .u-theme--accent & {
+          color: #2f854d;
+        }
+        ::ng-deep .u-theme--accent &[disabled] {
+          background-color: #cecdd1;
+          color: #999;
+        }
+        ::ng-deep .u-theme--error & {
+          color: #c8604d;
+        }
+        ::ng-deep .u-theme--error &[disabled] {
+          background-color: #cecdd1;
+          color: #999;
+        }
+        ::ng-deep .u-theme--warn & {
+          color: #dea83b;
+        }
+        ::ng-deep .u-theme--warn &[disabled] {
+          background-color: #cecdd1;
+          color: #999;
+        }
+        ::ng-deep .u-theme--highlight & {
+          color: #efe557;
+        }
+        ::ng-deep .u-theme--highlight &[disabled] {
+          background-color: #cecdd1;
+          color: #999;
+        }
+        ::ng-deep .u-theme--pure & {
+          color: #fafafa;
+        }
+        ::ng-deep .u-theme--pure &[disabled] {
+          background-color: #cecdd1;
+          color: #999;
+        }
+        ::ng-deep .u-theme--utility & {
+          color: #999;
+        }
+        ::ng-deep .u-theme--utility &[disabled] {
+          background-color: #cecdd1;
+          color: #999;
+        }
+      }
+    }
 
-            &[disabled] {
-              background-color: color(utility, light);
-              color: color(utility);
-            }
+    @include test('should return the correct theme and color for background-color') {
+      @include assert {
+        @include output {
+          @include theme-color(background-color);
+        }
+        @include expect {
+          ::ng-deep .u-theme--primary & {
+            background-color: #00538a;
+            color: #fafafa;
+          }
+          ::ng-deep .u-theme--primary &[disabled] {
+            background-color: #cecdd1;
+            color: #999;
+          }
+          ::ng-deep .u-theme--accent & {
+            background-color: #2f854d;
+            color: #fafafa;
+          }
+          ::ng-deep .u-theme--accent &[disabled] {
+            background-color: #cecdd1;
+            color: #999;
+          }
+          ::ng-deep .u-theme--error & {
+            background-color: #c8604d;
+            color: #fafafa;
+          }
+          ::ng-deep .u-theme--error &[disabled] {
+            background-color: #cecdd1;
+            color: #999;
+          }
+          ::ng-deep .u-theme--warn & {
+            background-color: #dea83b;
+            color: #fafafa;
+          }
+          ::ng-deep .u-theme--warn &[disabled] {
+            background-color: #cecdd1;
+            color: #999;
+          }
+          ::ng-deep .u-theme--highlight & {
+            background-color: #efe557;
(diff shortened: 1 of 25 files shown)


terminus-ui/scss/helpers/_color.spec.scss
```diff
@@ -9,80 +9,65 @@
...
+        @include theme-color(color);
+      }
+      @include expect {
+        @each $key, $value in $theme-and-color {
+          ::ng-deep .u-theme--#{$key} & {
+            color: $value;
+
+            &[disabled] {
+              background-color: color(utility, light);
+              color: color(utility);
```
atlwendy: Does this only test for `disabled` color, or this is like a skeleton for others to come along?

terminus-ui/scss/helpers/_layout.scss
```diff
@@ -35,7 +35,7 @@
   $position: absolute,
   $content: ''
 ) {
-  content: $content;
+  content: unquote("\'#{$content}\'");
```
atlwendy: This is an interesting way to escape...

atlwendy (approved):
Heard about it for a while but never tried it out. Looks pretty cool.

terminus-ui/scss/helpers/_layout.scss
shani-terminus: We had to do this because our linter was changing double quote to single quote. There were other linting issues, but I assume we'll iron those out.

The function is built in to SASS: https://sass-lang.com/documentation/functions/string

terminus-ui/scss/helpers/_color.spec.scss
shani-terminus: It's a test of what we have now. The `theme-color` mixin sets the disabled pseudo class. 

Here's the code: https://github.com/GetTerminus/terminus-ui/blob/82e377d588a4e28aebe5cfe3f4bc2dddde1f5db4/terminus-ui/scss/helpers/_color.scss#L354

terminus-ui/scss/helpers/_color.spec.scss
```diff
@@ -9,80 +9,65 @@
...
   }
 
-  /*   @include test('should return the correct theme (this seems to fail due to class being added by true)') {
-    @include assert{
+  @include test('should return the correct theme and color for background-color') {
+    @include assert {
       @include output {
         @include theme-color(background-color);
       }
       @include expect {
```
benjamincharity: Any reason we seem to be testing Sass code here but testing pure CSS output in our other tests?

terminus-ui/scss/helpers/_cursors.spec.scss
```diff
@@ -1,2 +1,79 @@
 @import 'true';
 @import './cursors';
+
+@include describe ('cursors') {
```
benjamincharity: Seems like we test all cursor types here together, then again individually below. Do we need both sets of tests?

terminus-ui/scss/helpers/_input-placeholder.spec.scss
```diff
@@ -1,2 +1,40 @@
...
+* This test fails with the message "Compared values have no visual difference."
+https://github.com/GetTerminus/terminus-ui/issues/1531
+
+
+@include describe ('input-placeholder') {
+  @include test('should set input-placeholder') {
+    @include assert {
+      @include output {
+        @include input-placeholder(){
+          .test {
```
benjamincharity: Not sure how Sass treats it, but as far as the output CSS goes, only properties should go inside this mixin rather than selectors. Also, the mixin doesn't need to be 'invoked' if no params are sent:

```scss
.test {
  @include input-placeholder {
    color: red;
  }
}

// Becomes
.test::-webkit-input-placeholder {
  color: red;
}
.test::-moz-placeholder {
  color: red;
}
.test:-ms-input-placeholder {
  color: red;
}
.test::placeholder {
  color: red;
}
.test::-moz-placeholder {
  color: red;
}
.test::-webkit-input-placeholder {
  color: red;
}
.test:-ms-input-placeholder {
  color: red;
}
```

---

Also, side note / pro-tip: When you need the generated CSS output for test expects, I've found it useful to just paste my code into https://www.sassmeister.com/ and copy the output.

terminus-ui/scss/helpers/_shadows.spec.scss
```diff
@@ -1,2 +1,729 @@
...
+          (0, 16, 24, 2, rgba(0, 0, 0, .14)),
+          (0, 6, 30, 5, rgba(0, 0, 0, .12))
+        )
+      );
+
+      @include assert-equal($output, $expect);
+    }
+  }
+
+/*
```
benjamincharity: We should add a comment here to explain why it's commented out.

terminus-ui/scss/helpers/_typography.scss
```diff
@@ -219,6 +219,7 @@ $typography-body-levels: (
 
   // FIXEDWIDTH
   @if $format == 'fixedwidth' {
+    $typeface: monospace;
```
benjamincharity: Unused variable

terminus-ui/scss/helpers/_z-index.spec.scss
```diff
@@ -1,2 +1,46 @@
 @import 'true';
 @import './z-index';
+
+@include describe ('z-index') {
+  @include test('should set z index global-overlay') {
+    $direction: z(global-overlay);
```
benjamincharity: Possibly update variable name

benjamincharity (changes requested):
Couple questions, couple small notes

terminus-ui/scss/helpers/_color.spec.scss
shani-terminus: The classes added by true are not something we should be explicitly looking for, and when taking straight output, I was getting errors with true-added-class `.test-output`

In general, I'm not concerned about testing mixins or functions within mixins, but this one is native sass, so I will go back, since I know more now, to see if there's a better way of testing this.

terminus-ui/scss/helpers/_cursors.spec.scss
shani-terminus: Right. This is testing the mixin, and the one below tests the function.

terminus-ui/scss/helpers/_shadows.spec.scss
shani-terminus: will do

terminus-ui/scss/helpers/_typography.scss
shani-terminus: ah! I forgot to take this out. I was trying to work around this bad practice:
`@included typography(fixedwidth)` returns
```scss
          .format-test {
            font-family: Roboto, Helvetica Neue, Helvetica, Arial, sans-serif;
            font-family: Lucida Console, Monaco, DejaVu Sans Mono, monospace;
          }
```

benjamincharity: Ah great point. We can definitely improve our output there. I'll make a ticket 👍

terminus-ui/scss/helpers/_z-index.spec.scss
shani-terminus: done

terminus-ui/scss/helpers/_input-placeholder.spec.scss
shani-terminus: Thanks for the sassmeister tip. 

I made the change, and still got the same failure error `Compared values have no visual difference.` which is addressed in another ticket. So for this test, it doesn't seem to make a difference whether the selector is inside or outside. I might have even had it outside initially, then moved it in after talking to Miriam Suzanne at OddBird (who wrote sass-true).

More specifically, I had issues with media queries when it was outside (as you suggested), and it worked when it was inside (as I had it originally).

Here's that conversation: https://github.com/oddbird/true/issues/147#issuecomment-499140337

benjamincharity (approved):

PR #1553, opened 2019-06-10
Add event emitter autocomplete

### LINKED ISSUE #1550: Autocomplete: adding back event emitter and docs
`opened` and `closed` event emitter were removed in new component. Need to add them back, also a few docs that need to be updated for usage purpose.

atlwendy:
closes #1550


benjamincharity (approved):
👍 	

For future reference, there is also a commit prefix for docs: `docs(Navigation): etc..`

PR #1557, opened 2019-06-12
chore(Tooltip): converted to int test

### LINKED ISSUE #1296: Tooltip: Convert to integration test

shani-terminus:
ISSUES CLOSED: #1296

### DIFF
--- a/terminus-ui/button/src/button.component.spec.ts
+++ b/terminus-ui/button/src/button.component.spec.ts
@@ -7,7 +7,6 @@ import {
   tick,
 } from '@angular/core/testing';
 import { By } from '@angular/platform-browser';
-import { getDomAttribute } from '@terminus/ngx-tools';
 import {
   createComponent,
   createMouseEvent,
--- a/terminus-ui/button/src/button.component.spec.ts
+++ b/terminus-ui/button/src/button.component.spec.ts
@@ -7,6 +7,7 @@ import {
   tick,
 } from '@angular/core/testing';
 import { By } from '@angular/platform-browser';
+import { getDomAttribute } from '@terminus/ngx-tools';
 import {
   createComponent,
   createMouseEvent,
--- a/terminus-ui/tooltip/src/tooltip.component.html
+++ b/terminus-ui/tooltip/src/tooltip.component.html
@@ -1,4 +1,4 @@
-<span
+<span #tooltip="matTooltip"
   class="c-tooltip qa-tooltip"
   [matTooltip]="tooltipValue"
   [matTooltipPosition]="tooltipPosition"
--- a/terminus-ui/tooltip/src/tooltip.component.spec.ts
+++ b/terminus-ui/tooltip/src/tooltip.component.spec.ts
@@ -1,46 +1,131 @@
+import { ComponentFixture } from '@angular/core/testing';
+import { By } from '@angular/platform-browser';
+import { NoopAnimationsModule } from '@angular/platform-browser/animations';
+import { createComponent } from '@terminus/ngx-tools/testing';
+
 import { TsTooltipComponent } from './tooltip.component';
+import { TsTooltipModule } from './tooltip.module';
+
+import * as testComponents from '../testing/src/test-components';
+import { getTooltipText } from '../testing/src/test-helpers';
 
 
 describe(`TsTooltipComponent`, function() {
-  let component: TsTooltipComponent;
 
-  beforeEach(() => {
-    component = new TsTooltipComponent();
+  describe(`Defaults`, () => {
+    let component: testComponents.Basic;
+    let fixture: ComponentFixture<testComponents.Basic>;
+    let tooltipEl: HTMLElement;
+    let tooltipComponent: TsTooltipComponent;
+
+    beforeEach(() => {
+      fixture = createComponent(testComponents.Basic, [], [TsTooltipModule]);
+      component = fixture.componentInstance;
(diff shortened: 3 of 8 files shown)


terminus-ui/tooltip/src/tooltip.component.spec.ts
```diff
@@ -1,46 +1,139 @@
...
-      component.tooltipPosition = 'above';
-      expect(component.tooltipPosition).toEqual('above');
+    describe(`tooltipPosition`, () => {
+      test(`should set and retrieve above`, () => {
+        component.tooltipPosition = 'above';
+        fixture.detectChanges();
 
-    });
+        expect(tooltipComponent.tooltipPosition).toEqual('above');
+        expect(tooltipEl.attributes.getNamedItem('ng-reflect-position').value).toEqual('above');
```
benjamincharity: I don't think we should be testing `ng-*` data attributes directly. We have no control over those values. We should probably be testing for positions or classes. (same comment below)

terminus-ui/tooltip/src/tooltip.component.html
```diff
@@ -1,8 +1,9 @@
-<span
+<span #tooltip="matTooltip"
   class="c-tooltip qa-tooltip"
   [matTooltip]="tooltipValue"
   [matTooltipPosition]="tooltipPosition"
   [ngClass]="{'c-tooltip--underline': hasUnderline}"
+  (click)="tooltip.show()"
```
benjamincharity: We shouldn't be changing how the tooltip works for consumers. I'd rather add a `show` method to our tooltip component that calls Material's `show`.

terminus-ui/tooltip/testing/src/test-helpers.ts
```diff
@@ -0,0 +1,17 @@
+import { ComponentFixture } from '@angular/core/testing';
+import { By } from '@angular/platform-browser';
+
+/**
+ * Helper function to display the tooltip
+ *
+ * @param fixture - the fixture with a tooltip
+ */
+// tslint:disable-next-line:no-any
+export function showTooltip(fixture: ComponentFixture<any>) {
```
benjamincharity: Need `void` return type here

benjamincharity (changes requested):

terminus-ui/tooltip/testing/src/test-helpers.ts
```diff
@@ -0,0 +1,20 @@
+import { ComponentFixture } from '@angular/core/testing';
+import { By } from '@angular/platform-browser';
+import { MatTooltip } from '@angular/material';
```
benjamincharity: I don't think this one is being used anymore

terminus-ui/tooltip/testing/src/test-helpers.ts
```diff
@@ -0,0 +1,20 @@
...
+
+
+// @ViewChild(MatTooltip) tooltip:MatTooltip;
```
benjamincharity: Legacy comment

terminus-ui/tooltip/testing/src/test-components.ts
```diff
@@ -0,0 +1,69 @@
+import {
+  Component,
+  NgModule,
+  ViewChild,
+} from '@angular/core';
+
+import {
+  TsTooltipComponent,
+  TsTooltipModule,
+} from '../../src/tooltip.module';
```
benjamincharity: This import should be like an external consumer: `@terminus/ui/tooltip`

terminus-ui/tooltip/src/tooltip.component.ts
```diff
@@ -79,4 +83,17 @@ export class TsTooltipComponent {
...
+  /**
+   * Access Material Tooltip Directive
+   */
+  @ViewChild('tooltip', {static: true})
+  public matTooltip;
+
+  /**
+   * Call Material Tooltip's show() method
+   */
+  public showTooltip() {
```
benjamincharity: Missing void return type

terminus-ui/tooltip/src/tooltip.component.ts
```diff
@@ -3,8 +3,12 @@ import {
   Component,
   Input,
   isDevMode,
+  ViewChild,
   ViewEncapsulation,
 } from '@angular/core';
+import { MatTooltip } from '@angular/material';
```
benjamincharity: I don't think this one is being used anymore

benjamincharity (changes requested):

terminus-ui/button/src/button.component.spec.ts
```diff
@@ -7,6 +7,7 @@ import {
   tick,
 } from '@angular/core/testing';
 import { By } from '@angular/platform-browser';
+import { getDomAttribute } from '@terminus/ngx-tools';
```
benjamincharity: Oh, also, this is a rebase issue I think.. you can go ahead and remove this line.

terminus-ui/tooltip/testing/src/test-components.ts
shani-terminus: Yes, this was supposed to be temporary

benjamincharity (approved):

PR #1562, opened 2019-06-19
Update dependencies

benjamincharity:
⚠️ BREAKING CHANGE ⚠️ 

🚨 NOTE: Create v13 legacy branch before merging this to `release` 🚨 

---

### Primary Changes

- Enable incremental TS builds
- Update Angular to v8 and all other dependencies to latest
- Fix trailing errors from ExpansionPanel tests
- Add static definition to all View/ContentChild's
- Alphabetized jest config to match other files
- Update Jest config for latest versions of Jest and jest-preset-angular
- Updated `build:app` command. NOTE: You must build the library before running this command.
- Fix inconsistent yarn commands
- Add resolution to force [JSDom to v14 to get past JSDom bug](https://github.com/jsdom/jsdom/issues/2499#issuecomment-472168858)
- Reorder select view-access to match other components order
- Fix tests and types that mysteriously broke during upgrade.
- Cast chart types to any in Demo. This is a temporary solution as amcharts is still working through strict mode (it didn't exist when they first wrote v4).
- Don't lint `test-sass.js`
- Update `Dockerfile` to use a newer version of node required by angular buildkit
- Use un-deprecated version of `startWith`

---

@GetTerminus/uxe NOTE: please thoroughly read https://angular.io/guide/static-query-migration to understand what is happening with the static view queries and why it's needed.


package.json
```diff
@@ -167,109 +166,101 @@
...
+    "@ngrx/effects": "^8.0.1",
+    "@ngrx/store": "^8.0.1",
+    "@semantic-release/changelog": "^3.0.4",
+    "@semantic-release/exec": "^3.3.3",
+    "@semantic-release/git": "^7.0.12",
+    "@semantic-release/github": "^5.4.0",
+    "@semantic-release/npm": "^5.1.9",
+    "@semantic-release/release-notes-generator": "^7.2.0",
+    "@terminus/eslint-config-frontend": "^1.0.5",
+    "@terminus/tslint-config-frontend": "^1.0.6",
```
shani-terminus, line 204: Do we need stylelint?

terminus-ui/autocomplete/src/autocomplete.component.spec.ts
```diff
@@ -168,7 +168,14 @@ describe(`TsAutocompleteComponent`, function() {
 
 
     test(`should allow removal with the backspace key`, () => {
-      jest.useFakeTimers();
+      // NOTE: `dispatchKeyboardEvent` suddenly stopped working for this test during the Angular v8 upgrade
```
shani-terminus, line 171: this work-around and NOTE should also be applied to "test(`should not open when disabled`". With dispatchKeyboardEvent not working, that test is not reliable.

terminus-ui/autocomplete/src/autocomplete.component.ts
```diff
@@ -270,23 +252,6 @@ export class TsAutocompleteComponent implements OnInit,
   @ContentChildren(TsOptgroupComponent)
   public optionGroups!: QueryList<TsOptgroupComponent>;
 
-  /**
```
shani-terminus, line 278: are these vestiges?

.gitignore
```diff
@@ -14,3 +14,4 @@ e2e/**/*.js.map
 .vscode
 docs/*
 coverage/
+.tsbuildinfo
```
atlwendy, line 17: I don't have this locally. does build generate this?

terminus-ui/button/src/button.component.spec.ts
```diff
@@ -7,6 +7,7 @@ import {
   tick,
 } from '@angular/core/testing';
 import { By } from '@angular/platform-browser';
+import { getDomAttribute } from '@terminus/ngx-tools';
```
shani-terminus: This import isn't getting used.

demo/app/components/chart/chart.component.ts
```diff
@@ -68,25 +72,25 @@ export class ChartComponent {
     if (tsChartXYTypeCheck(chart)) {
       chart.data = XY_DATA;
 
-      const dateAxis = chart.xAxes.push(new am4charts.DateAxis());
+      const dateAxis = chart.xAxes.push(new am4charts.DateAxis() as any);
```
atlwendy, line 75: Would ts be mad if no `as any` passed in?

package.json
```diff
@@ -17,7 +17,7 @@
     "ng": "ng",
     "start:app": "yarn run build:styles && yarn run link && ng serve",
     "start:app:aot": "yarn run link && ng serve --aot",
-    "build:app": "cpr dist/terminus-ui demo/@terminus/ui --overwrite && ng build --app packages --prod && rimraf demo/@terminus",
+    "build:app": "ng build --project packages --prod",
```
atlwendy, line 20: so much better 😄

package.json
```diff
@@ -35,10 +35,9 @@
     "build:explore": "source-map-explorer dist/terminus-ui/bundles/terminus-ui.umd.min.js dist/terminus-ui/bundles/terminus-ui.umd.min.js.map",
     "////////// Section: Library:Testing": "==============================",
     "test:NOTE": "jest --watch: currently hangs when determining which tests to run. For now we simply run all",
-    "test": "jest --watch",
+    "test:lib": "jest --config ./jest.config.js",
     "test:ci": "jest --runInBand --coverage",
     "test:ci:local": "jest --coverage",
```
atlwendy, line 40: So we can't do `yarn test` any more? why did you take `--watch` away?

terminus-ui/autocomplete/src/autocomplete-panel/autocomplete-trigger.directive.ts
```diff
@@ -162,7 +162,7 @@ export class TsAutocompleteTriggerDirective<ValueType = string> implements Contr
   /**
    * Stream of autocomplete option selections
    */
-  public readonly optionSelections: Observable<TsOptionSelectionChange> = defer(() => {
+  public readonly optionSelections: Observable<TsOptionSelectionChange> | Observable<{}> = defer(() => {
     if (this.autocompletePanel && this.autocompletePanel.options) {
       return merge(...this.autocompletePanel.options.map(option => option.selectionChange));
```
atlwendy: is this possible return of `Observable<{}>` due to rxjs upgrade? it looked like it always should be there, not sure why it was fine.

terminus-ui/autocomplete/src/autocomplete.component.spec.ts
```diff
@@ -168,7 +168,14 @@ describe(`TsAutocompleteComponent`, function() {
...
+      const event = document.createEvent('KeyboardEvent');
+      event.initEvent('keydown', true, false);
+      Object.defineProperties(event, {
+        keyCode: { get: () => KEYS.BACKSPACE.keyCode },
+        key: { get: () => KEYS.BACKSPACE.code },
+      });
+
```
atlwendy, line 178: basically reimplemented `dispatchKeyboardEvent` here... wonder whether v8 makes that helper working intermittently.

terminus-ui/expansion-panel/src/expansion-panel.component.spec.ts
```diff
@@ -203,44 +209,51 @@ describe(`TsExpansionPanelComponent`, function() {
...
+      fixture.detectChanges();
+      expect(triggerElement.getAttribute('style')).toContain('height:55px');
 
       const toggleIcon = triggerElement.querySelector('.ts-expansion-panel__indicator');
       expect(toggleIcon).toBeNull();
+
+      discardPeriodicTasks();
       expect.assertions(3);
-    });
+    }));
```
atlwendy, line 256: So great to see all the red console errors gone!!!

terminus-ui/expansion-panel/src/expansion-panel.component.ts
```diff
@@ -297,10 +303,8 @@ export class TsExpansionPanelComponent extends CdkAccordionItem implements After
     // istanbul ignore else
     if (this.lazyContent) {
       // Render the content as soon as the panel becomes open.
-      this.opened.pipe(
-        // tslint:disable: no-non-null-assertion
-        startWith<void>(null!),
-        // tslint:enable: no-non-null-assertion
+      scheduled([[of(null)], this.opened], asap).pipe(
+        concatAll(),
```
atlwendy, line 307: oh I like this... is this new in rxjs update?

terminus-ui/input/testing/src/test-components.ts
```diff
@@ -31,24 +31,18 @@ import {
...
 
 
-@Component({
-  template: `
-    <ts-input
-      [datepicker]="true"
-    ></ts-input>
-  `,
-})
+@Component({template: `<ts-input [datepicker]="true"></ts-input>`})
```
atlwendy, line 39: honestly i'd prefer the old layout, but might not go with your rule well...

terminus-ui/tabs/src/collection/tab-collection.component.spec.ts
```diff
@@ -587,7 +587,6 @@ describe(`TsTabCollectionComponent`, function() {
       secondLabel.click();
       fixture.detectChanges();
       tick();
-      fixture.detectChanges();
```
atlwendy, line 590: why did you take detectChanges away after tick? couple of times i thought detectChanges was necessary after tick.

terminus-ui/tabs/testing/src/test-components.ts
```diff
@@ -179,7 +179,7 @@ export class AsyncTabs implements OnInit {
 export class SimpleLabels {
   public otherLabel = 'Bar';
   public otherContent = 'bar content';
-  @ViewChild('test') public testSelector: any;
+  @ViewChild('test', {static: true}) public testSelector: any;
```
atlwendy, line 182: is this public declaration supposed to move down? is that the rule or just preference?

tooling/Dockerfile
```diff
@@ -1,9 +1,13 @@
...
-RUN apt-get update && \
-apt-get install -y python-pip python-dev && \
-pip install awscli
+RUN apt-get update
+
+RUN apt-get install -y python-pip
+
+RUN apt-get install -y python2.7-dev
+
+RUN pip install awscli
```
atlwendy, line 11: really? you can't tag them all together any more?

atlwendy (approved):
Nothing jumps out... mostly just questions...
Hopefully when you merge, no random test failure 🤞

.gitignore
benjamincharity: I have not actually seen that item yet, but the the TypeScript docs about turning on incremental builds it calls out ignoring that.

demo/app/components/chart/chart.component.ts
benjamincharity: Yeah, that's why I had to make these changes. Strict mode didn't exist when amCharts v4 was written so they are working on making it fully compliant - but some holes still exist.

I opted to let consumers kill typings rather than disabling 'strict' for everything.

shani-terminus (changes requested):
Excited about the v8 updates!

package.json
benjamincharity: This must be a bad rebase.. my change was actually the opposite - moving the command back to just `test`

benjamincharity: Definitely also missed the watch flag.. adding back

terminus-ui/autocomplete/src/autocomplete-panel/autocomplete-trigger.directive.ts
benjamincharity: Yeah it's either a TS or RxJS upgrade issue. Should likely have always been like this

terminus-ui/expansion-panel/src/expansion-panel.component.ts
benjamincharity: Yep, there's actually two ways to solve this. Basically they deprecated the version of `startWith` that has only a single type. So we can use `startWith<null, void>` (which I did somewhere). But the other refactor was using `scheduled` which I had never used so I thought I'd try it. Both seem pretty logical to me

terminus-ui/input/testing/src/test-components.ts
benjamincharity: I'd never do this in a component, but when dealing with test components, if it's a single line of template I don't mind the compact nature. Open to enforcing it across test files too if our team wants 👍

terminus-ui/tabs/src/collection/tab-collection.component.spec.ts
benjamincharity: It didn't seem to be needed post-tick. There are some cases it could be needed for though so this isn't a broad rule. I believe it's only needed if the next tick may change something internal during the tick

terminus-ui/tabs/testing/src/test-components.ts
benjamincharity: There's no lint rule for this, (we should write one) but it has been enforced in component files by review - just not in test files. I am open though if we'd prefer the same format rules in test components too

tooling/Dockerfile
benjamincharity: You can, but you don't get individual failures when you do so it's much harder to debug. I split them apart while trying to fix some install issues and really didn't see a benefit to merging them back together ¯\_(ツ)_/¯

terminus-ui/expansion-panel/src/expansion-panel.component.ts
benjamincharity: Example of the other version:

Screen Shot 2019-06-21 at 9 29 30 AM

benjamincharity: New definition:

```typescript
export function startWith<T, D>(v1: D): OperatorFunction<T, T | D>;
```

package.json
benjamincharity: Not sure how I lost that during the rebase.. added

terminus-ui/autocomplete/src/autocomplete.component.spec.ts
benjamincharity: I don't want to add refactoring 19 tests to this ticket. My vote is to handle this in the ticket for `dispatchKeyboardEvent`

terminus-ui/autocomplete/src/autocomplete.component.ts
benjamincharity: I believe so

terminus-ui/button/src/button.component.spec.ts
benjamincharity: Removed. Also added a ticket in our lint repo for this https://github.com/GetTerminus/eslint-config-frontend/issues/24

PR #1565, opened 2019-06-21
fix(autocomplete): checkbox no longer appears in front of autocomplete options

### LINKED ISSUE #1561: Autocomplete: Should not show checkboxes in dropdown
In production mode, the autocomplete dropdown is showing checkboxes.

Currently checkboxes are not shown in local development environment, but shown in our demo host at GitHub as well as in engage. So it's possible to have behavior difference somewhere in the code in dev or production environment.

Example:
Screen Shot 2019-06-17 at 2 09 38 PM

Possible culprit might live in the `option.component.ts`:

```
   if (this.viewContainerRef['_data'].componentView) {
      // eslint-disable-next-line dot-notation
      const parentComponent = this.viewContainerRef['_data'].componentView.parent.component.constructor.name.toLowerCase();
      if (parentComponent.includes('autocomplete')) {
        this.autocompleteComponent = true;
        console.log('this is autocompletecomponent');
        console.log('allowMultiple: ', this.allowMultiple);
      } else if (parentComponent.includes('select')) {
        this.selectComponent = true;
      }
    }
```

atlwendy:
closes #1561


terminus-ui/select/src/select.component.ts
```diff
@@ -256,6 +256,11 @@ export class TsSelectComponent implements
   OnDestroy,
   TsFormFieldControl<unknown> {
 
+  /**
+   *
```
benjamincharity: Missing comment

benjamincharity (changes requested):

terminus-ui/option/src/option.component.ts
```diff
@@ -53,6 +53,7 @@ export class TsOptionSelectionChange {
  * Contains properties that the options can inherit. Used by {@link TS_OPTION_PARENT_COMPONENT}
  */
 export interface TsOptionParentComponent {
+  componentName: string;
```
benjamincharity, line 56: @shani-terminus for the sake of knowledge sharing:

The option component needs to know whether it's parent is the autocomplete or select to know if it should show the checkbox when multiple mode is enabled. The code we originally had for this was actually giving us a false positive.. it was looking *all* the way up to the very top parent component - which because our demos are named for the item within, our parent just _happened_ to have the same name so it seemed to be working.

But when used in another app that doesn't use the same component name it doesn't work correctly. This is why we couldn't reproduce locally.

So we opted to just add a `componentName` property so the child can check directly.

atlwendy: A minor correction on that explanation. The original code wasn't getting all the way to the top parent component. It was current component's parent, which in demo case happens to be`AutocompleteComponent`, while in others could be different.

benjamincharity (approved):

PR #1589, opened 2019-07-02
fix(RadioGroup): visual radios now dispatch change event

### LINKED ISSUE #1549: RadioGroup: Change event not getting fired in visual mode
- [ ] Make sure all events are being fired in all modes

Reported by Dean via Engage

shani-terminus:
ISSUES CLOSED: #1549

### DIFF
--- a/demo/app/components/radio/radio.component.html
+++ b/demo/app/components/radio/radio.component.html
@@ -40,6 +40,7 @@
       [formatModelValueFn]="modelFormatter"
       [formControl]="myForm.get('myRadioGroup2')"
       [options]="items2$ | async"
+      (change)="selected($event)"
     ></ts-radio-group>
 
     Selected value: {{ myForm.get('myRadioGroup2')?.value }}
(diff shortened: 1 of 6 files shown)


benjamincharity (changes requested):
Hmm now we've got an issue where we have two different change events depending on which version of the radio you are using:

default: `TsRadioChange`
visual: `MouseEvent`

From the functional side I think all we need to do is fire `change.emit(myEvent)` inside the label event. But we also can't just use that same event either. `TsRadioChange` is just an extension of `MatRadioChange` which expects a `MatRadioButton` as it's source value: 

https://github.com/angular/components/blob/5c51301b4a3c266cc700717b814f1db25528f649/src/material/radio/radio.ts#L81

So I think what we need to do is create our own event and fire it:

A) stop extending the Material event and define our own:

```typescript
export class TsRadioChange {
  constructor(
    // The group that emit the change event
    public source: TsRadioGroupComponent,
    // The value of the TsRadioButton
    public value: any,
  ) {}
}
```

You'll probably want a helper function to create/fire the event `new TsRadioChange(this, this.value)`

---

I'm putting the do not merge label on this until I hear back from the core teams about whether they are using `event.source`.

benjamincharity (approved):

atlwendy (approved):

PR #1592, opened 2019-07-10
chore(LoginForm): login form will have a deprecation warning when usi…

### LINKED ISSUE #1468: LogInForm: Deprecate existing submit output and add replacement
Custom components should not have event emitters whose names can collide with native methods. Listeners subscribed to an output with such a name will also be invoked when the native event is raised.

- [x] Add a new emitter named `submission` that functions exactly as the existing `submit` emitter does
- [x] Deprecate the `submit` emitter with an `@deprecated` JSDoc comment. Let's not callout specifically which version will remove it as lately we've had to move our changes around.
- [ ] Test the new emitter
- [x] Update JSDoc and usage docs

shani-terminus:
…ng 'submit'

and a note to use 'submission' instead

ISSUES CLOSED: #1468

### DIFF
--- a/terminus-ui/login-form/src/login-form.component.html
+++ b/terminus-ui/login-form/src/login-form.component.html
@@ -59,6 +59,7 @@
       [showProgress]="inProgress || isRedirecting"
       [isDisabled]="!loginForm?.valid"
       (clicked)="submit.emit(loginForm?.value)"
+      (clicked)="submission.emit(loginForm?.value)"
       tabindex="-1"
       tabIndex="4"
     >
--- a/terminus-ui/login-form/src/login-form.component.md
+++ b/terminus-ui/login-form/src/login-form.component.md
@@ -1,0 +1,19 @@
+<h1>Login Form</h1>
+
+<!-- START doctoc generated TOC please keep comment here to allow auto update -->
+<!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
+**Table of Contents**
+
+- [Basic Usage](#basic-usage)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+## Basic Usage
+
+Create a basic login form with submit function:
+
+```html
+<ts-login-form
+  (submission)="mySumbitFunction($event)"
+></ts-login-form>
+```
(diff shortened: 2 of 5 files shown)


shani-terminus:
Considering the queue, I'm thinking it's best to leave it as is for now and add the test later with the ticket.

terminus-ui/login-form/src/login-form.component.html
```diff
@@ -58,7 +58,7 @@
       buttonType="submit"
       [showProgress]="inProgress || isRedirecting"
       [isDisabled]="!loginForm?.valid"
-      (clicked)="submit.emit(loginForm?.value)"
+      (clicked)="submission.emit(loginForm?.value)"
```
benjamincharity, line 62: Changing this will result in a breaking change rather than a deprecation. During the deprecation period, we need the emitter for both `submit` and `submission` to be fired for the same click.

terminus-ui/login-form/src/login-form.component.md
```diff
@@ -0,0 +1,34 @@
...
+Create a basic login form with submit function:
+
+```html
+<ts-login-form
+  (submission)="mySumbitFunction($event)"
+></ts-login-form>
+```
+
+
+## Customizations
```
benjamincharity: I don't think we need to call out all the inputs in this doc. The usage docs are more 'use-case docs'.

terminus-ui/login-form/src/login-form.component.ts
```diff
@@ -187,10 +187,20 @@ export class TsLoginFormComponent implements OnChanges {
...
   // TODO: Rename to avoid conflict with native events: https://github.com/GetTerminus/terminus-ui/issues/1469
+  /**
+   * @deprecated submit should be changed to 'submission'
+   */
   // tslint:disable-next-line: no-output-native
   @Output()
   public readonly submit: EventEmitter<TsLoginFormResponse> = new EventEmitter();
 
+  /**
+ * Emit an event on form submission
```
benjamincharity: Comment alignment is a little off here

benjamincharity (changes requested):

terminus-ui/login-form/src/login-form.component.html
shani-terminus: 🤦‍♀

benjamincharity:
@shani-terminus I forgot to add, we should update the demo to use the new emitter also

benjamincharity (approved):

PR #1593, opened 2019-07-10
chore(Toggle): has a deprecation notice for 'change'

### LINKED ISSUE #1471: Toggle: Deprecate existing change output and add replacement
Custom components should not have event emitters whose names can collide with native methods. Listeners subscribed to an output with such a name will also be invoked when the native event is raised.

- [x] Add a new emitter named `selectionChange` that functions exactly as the existing `change` emitter does
- [x] Deprecate the `change` emitter with an `@deprecated` JSDoc comment. (Let's not callout specifically which version will remove it as lately we've had to move our changes around.)
- [ ] Test the new emitter
- [x] Update JSDoc and usage docs

shani-terminus:
use 'selectionChange' moving forward

ISSUES CLOSED: #1471

### DIFF
--- a/terminus-ui/toggle/src/toggle.component.html
+++ b/terminus-ui/toggle/src/toggle.component.html
@@ -8,6 +8,7 @@
   [disabled]="isDisabled"
   [name]="name"
   (change)="change.emit($event)"
+  (change)="selectionChange.emit($event)"
 >
   <ng-content></ng-content>
 </mat-slide-toggle>
--- a/terminus-ui/toggle/src/toggle.component.ts
+++ b/terminus-ui/toggle/src/toggle.component.ts
@@ -35,7 +35,7 @@ export class TsSlideToggleChange extends MatSlideToggleChange {}
  *              labelPosition="before"
  *              name="My toggle"
  *              theme="accent"
- *              (change)="myMethod($event)"
+ *              (selectionChange)="myMethod($event)"
  * >My Toggle!</ts-toggle>
  *
  * <example-url>https://getterminus.github.io/ui-demos-release/components/toggle</example-url>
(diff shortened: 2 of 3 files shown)


terminus-ui/toggle/src/toggle.component.ts
```diff
@@ -104,7 +104,18 @@ export class TsToggleComponent extends TsReactiveFormBaseComponent {
    * Emit an event each time the toggle value changes
    */
   // TODO: Rename to avoid conflict with native events: https://github.com/GetTerminus/terminus-ui/issues/1472
+  /**
+   * @deprecated change will be deprecated, use 'selectionChange' instead
```
benjamincharity, line 108: We don't need to actually say the item will be deprecated. The `@deprecated` annotation will get printed out by the linter so we just need our custom message after. eg

`@deprecation Use 'selectionChange' instead`

Will print out as

`'change' is deprecated. Use 'selectionChange' instead'`

terminus-ui/toggle/src/toggle.component.ts
```diff
@@ -104,7 +104,18 @@ export class TsToggleComponent extends TsReactiveFormBaseComponent {
    * Emit an event each time the toggle value changes
    */
   // TODO: Rename to avoid conflict with native events: https://github.com/GetTerminus/terminus-ui/issues/1472
```
benjamincharity, line 106: I think we can remove this. We have tickets to track it and it's deprecated in the codebase. 👍

terminus-ui/toggle/src/toggle.component.ts
```diff
@@ -104,7 +104,18 @@ export class TsToggleComponent extends TsReactiveFormBaseComponent {
...
+   * @deprecated change will be deprecated, use 'selectionChange' instead
+   */
   // tslint:disable-next-line: no-output-native
   @Output()
   public readonly change: EventEmitter<TsSlideToggleChange> = new EventEmitter();
+
+  /**
+   * Emit an event each time the toggle value changes
+   */
+  // tslint:disable-next-line: no-output-native
```
benjamincharity, line 117: We don't need this line. It was to disable linting for `change` as it was rightly calling us out for using a native event name.

terminus-ui/toggle/src/toggle.component.html
```diff
@@ -8,6 +8,7 @@
   [disabled]="isDisabled"
   [name]="name"
   (change)="change.emit($event)"
+  (change)="selectionChange.emit($event)"
```
benjamincharity, line 11: I honestly had no clue we could use the same emitter twice on an element 🙃 learn something new every day!

benjamincharity (changes requested):
We should also update the demo to use the new emitter

benjamincharity (approved):

PR #1609, opened 2019-07-26
656 autofill bug

### LINKED ISSUE #656: LogInForm: Labels overlay input content when injected by the browser
This was originally reported by someone using LastPass but it has now been reported by someone only using Chrome's injection.

After letting Chrome inject saved information:
image (1)

Clicking anywhere on the page seems to trigger a change detection cycle to update it:
image (2)

- [ ] Determine if there is an event we can tie into to trigger a UI update after injection
- [ ] Add tests (if possible)

benjamincharity:
Fixes #656 

- Correctly set the autofill input for the email field
- Pass the correct element to the autofill monitor

### DIFF
--- a/terminus-ui/input/src/input.component.ts
+++ b/terminus-ui/input/src/input.component.ts
@@ -15,7 +15,6 @@ import {
   NgZone,
   OnChanges,
   OnDestroy,
-  OnInit,
   Optional,
   Output,
   Renderer2,
@@ -253,7 +252,6 @@ const DEFAULT_TEXTAREA_ROWS = 4;
 export class TsInputComponent implements
   // tslint:disable-next-line no-any
   TsFormFieldControl<any>,
-  OnInit,
   AfterViewInit,
   AfterContentInit,
   DoCheck,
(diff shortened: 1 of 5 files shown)


terminus-ui/login-form/src/login-form.component.html
```diff
@@ -15,6 +15,7 @@
     [isRequired]="true"
     [spellcheck]="false"
     [autocapitalize]="false"
+    autocomplete="email"
```
atlwendy, line 18: what difference does it make with or without this `autocomplete` input?

terminus-ui/input/src/input.component.ts
```diff
@@ -882,13 +880,30 @@ export class TsInputComponent implements
...
-   * Begin monitoring for the input autofill
+   * After the view is initialized, trigger any needed animations
    */
-  public ngOnInit(): void {
-    this.autofillMonitor.monitor(this.elementRef.nativeElement).subscribe(event => {
+  public ngAfterViewInit(): void {
+    // Begin monitoring for the input autofill
+    this.autofillMonitor.monitor(this.inputElement.nativeElement).subscribe(event => {
       this.autofilled = event.isAutofilled;
       this.stateChanges.next();
```
atlwendy, line 889: This is probably the meat of this fix...

atlwendy (approved):

terminus-ui/login-form/src/login-form.component.html
benjamincharity: The `autocomplete` attribute tells the browser what data should be used when providing autocompletion.

In my quick tests it seems that both Chrome and FF can infer this if the input has a name or type that defines `email`. But since this is the 'correct' way to define it, I thought it should be added.

terminus-ui/input/src/input.component.ts
benjamincharity: Yep. Mainly we were passing a reference to our component rather than the actual `<input>` element. Moving the setup into `ngAfterViewInit` just makes sure all the `ngIf`s are settled before using that reference.

shani-terminus (approved):

PR #1611, opened 2019-07-31
1572 select trigger refresh

### LINKED ISSUE #1572: Select: Allow the user to trigger a refresh of the available options
This wanted is needed for LinkedIn Forms & Ironman.

There is a use case where the user begins the wizard creation without first creating the needed items in an outside system. This feature would allow them to go create the needed item and return mid-flow to request an updated list to find the newly created item.

- [x] Add new input `showRefresh: boolean` that defaults to `false` (for backward compatibility)
- [x] If `true`, it should
  - [x] show as an option
  - [x] be last in the list of options
  - [x] be using the typography styles 'body 2' and 'caption' together (caption must be applied last)
  - [x] color: utility dark
  - [x] background: utility, xlight
  - [x] should have the text `Refresh & reload to bring in new choices.`
- [x] Clicking this option should emit an event `optionsRefreshRequested: void`
- [x] Add tests
- [x] Add to JSDoc and usage docs

Design: 

Screen Shot 2019-07-16 at 10 17 50 AM

benjamincharity:
Closes #1572

- Allow user to request a refresh of options + added to last select demo (gif below)
- upgrade lint-config
- move config to separate file
- use correct tsconfig for linting
- update excludes for lint tsconfig

refreshDemo

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,8 +47,8 @@
     "//=> Section: Yarn Management": "==============================",
     "fresh-yarn-install": "rm -rf node_modules && yarn install",
     "////////// Section: Linting": "==============================",
-    "lint:tslint": "npx tslint --project ./terminus-ui/tsconfig.json --config ./tslint.json --format stylish",
-    "lint:tslint:fix": "npx tslint --project ./terminus-ui/tsconfig.json --config ./tslint.json --format stylish --fix",
+    "lint:tslint": "npx tslint --project ./terminus-ui/tsconfig.lint.json --config ./tslint.json --format stylish",
+    "lint:tslint:fix": "npx tslint --project ./terminus-ui/tsconfig.lint.json --config ./tslint.json --format stylish --fix",
     "lint:tslint:spec": "npx tslint --project ./terminus-ui/tsconfig.spec.json --config ./tslint.spec.json --format stylish",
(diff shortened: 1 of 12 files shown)


atlwendy (approved):

package.json
```diff
@@ -87,21 +87,6 @@
       "config": "./tooling/cz-config.js"
     }
   },
-  "lint-staged": {
```
shani-terminus, line 90: I assume these settings aren't needed with the new version.

terminus-ui/tsconfig.lint.json
```diff
@@ -1,10 +1,8 @@
 {
-  "extends": "./../tsconfig.json",
+  "extends": "./tsconfig.json",
```
shani-terminus, line 2: it looks like this is a new file, but the whole thing isn't highlighted. 

Are these other excluded items excluded in tsconfig.json? I don't see that file in this PR.

package.json
benjamincharity: These now live in `lint-staged.config`

terminus-ui/tsconfig.lint.json
benjamincharity: That file has always existed and has no changes. Before we were extending like `this file < file B < file A` and now we are just extending straight from `file A`.

shani-terminus (approved):

PR #1618, opened 2019-08-01
feat(Link): support for local anchor links

### LINKED ISSUE #1607: Link: Extend functionality to support anchor (local) links
### Is your feature request related to a problem? Please describe.

As a consumer, I need a way to use `ts-link` with local hash links.

TsLink needs to expose another input: `fragment: string`. This is added to the underlying `<a>` tag like so:

```htm l
<!-- local link on the same page -->
<a [routerLink]="['.']" fragment="myElementId">My link</a>

<!-- local link on a different page -->
<a [routerLink]="['foo', 'bar]" fragment="myElementId">My link</a>
```

- [x] Add new input
- [x] Add to JSDoc
- [x] Add tests
- [x] Add to usage docs
	- Add notes about needed changes to router settings:
		- Enable `scrollPositionRestoration` in the router module using the `ExtraOptions` interface
		- Enable `anchorScrolling` in the router module using the `ExtraOptions` interface

benjamincharity:
ISSUES CLOSED: #1607

### DIFF
--- a/terminus-ui/link/src/link.component.spec.ts
+++ b/terminus-ui/link/src/link.component.spec.ts
@@ -1,14 +1,19 @@
 import { APP_BASE_HREF } from '@angular/common';
 import {
   Component,
+  Provider,
+  Type,
   ViewChild,
 } from '@angular/core';
 import {
+  async,
   ComponentFixture,
+  TestBed,
   tick,
 } from '@angular/core/testing';
 import { By } from '@angular/platform-browser';
-import { RouterModule } from '@angular/router';
+import { RouterTestingModule } from '@angular/router/testing';
+import { getDomAttribute } from '@terminus/ngx-tools';
 import { createComponent } from '@terminus/ngx-tools/testing';
 import { TsStyleThemeTypes } from '@terminus/ui/utilities';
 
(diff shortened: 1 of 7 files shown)


terminus-ui/link/src/link.component.spec.ts
```diff
@@ -39,25 +45,28 @@ class TestHostComponent {
 }
 
 
+export function createComponent<T>(component: Type<T>): ComponentFixture<T> {
+  TestBed.configureTestingModule({
```
shani-terminus: Why are we using TestBed instead of our createComponent from ngx-tools?

terminus-ui/link/src/link.component.spec.ts
```diff
@@ -1,15 +1,19 @@
...
 } from '@angular/core';
 import {
+  async,
   ComponentFixture,
+  TestBed,
   tick,
 } from '@angular/core/testing';
 import { By } from '@angular/platform-browser';
-import { RouterModule } from '@angular/router';
-import { createComponent } from '@terminus/ngx-tools/testing';
```
shani-terminus: ?

terminus-ui/link/src/link.component.spec.ts
```diff
@@ -126,4 +135,29 @@ describe(`TsLinkComponent`, function() {
 
   });
 
+
+  describe(`fragment`, function() {
```
shani-terminus: should this live within 'internal'?

terminus-ui/link/src/link.component.spec.ts
benjamincharity: Ahh originally the test was set up in a way that required 2 declared components which our function didn't support. That changed so I can go back to our function now 👍

terminus-ui/link/src/link.component.spec.ts
benjamincharity: ??

shani-terminus: why are we removing createComponent from ngx-tools?

benjamincharity: Same response as above: I was unable to use it originally. It has since been added back in response to the other callout

atlwendy (approved):

shani-terminus (approved):

PR #1629, opened 2019-08-12
BREAKING 1614 table pinned header

shani-terminus:
Once this is reviewed, I will rebase the commits into one.

Kapture 2019-08-12 at 11 55 53


terminus-ui/table/src/table.component.html
```diff
@@ -1,7 +1,5 @@
-<div class="ts-table__container">
-  <div class="ts-table__table">
-    <ng-container headerRowOutlet></ng-container>
-    <ng-container rowOutlet></ng-container>
-    <ng-container footerRowOutlet></ng-container>
-  </div>
-</div>
+<div class="ts-table__table">
```
benjamincharity: I think we should be able to remove this container also

terminus-ui/table/src/table.component.md
```diff
@@ -314,6 +314,31 @@ directive `alignment` and pass it any valid TsTableColumnAlignment value (`left`
...
+  <ng-container tsColumnDef="updated" noWrap="true" minWidth="100px" sticky>
+    <ts-header-cell *tsHeaderCellDef ts-sort-header>
+      Updated
+    </ts-header-cell>
+    <ts-cell *tsCellDef="let item">
+      {{ item.updated_at | date:"shortDate" }}
+    </ts-cell>
+  </ng-container>
+```
+
```
benjamincharity, line 344: We should add another sub section here about `stickyEnd`

terminus-ui/table/src/table.component.md
```diff
@@ -314,6 +314,31 @@ directive `alignment` and pass it any valid TsTableColumnAlignment value (`left`
...
+Defining the header as sticky will ensure it is always visible as the table scrolls. Set `sticky: true` in the `ts-header-row`'s `tsHeaderRowDef` directive.
+
+```html
+  <ts-header-row *tsHeaderRowDef="displayedColumns; sticky: true"></ts-header-row>
+```
+
+
+## Sticky column
+
+Defining a sticky column will pin it to the side as the table scrolls horizontally. Add the sticky parameter to the `<ng-container>` of the column definition.
```
benjamincharity: We should call out that more than one column can be sticky.

demo/app/components/table/table.component.ts
```diff
@@ -105,17 +105,14 @@ export class ExampleHttpDao {
...
   selector: 'demo-table',
   templateUrl: './table.component.html',
+  styleUrls: ['./table.component.scss'],
 })
 export class TableComponent implements AfterViewInit {
   allColumns = COLUMNS_SOURCE_GITHUB.slice(0);
   displayedColumns: string[] = [
     'updated',
     'number',
-    'state',
```
benjamincharity, line 114: Any specific reason to remove these columns?

terminus-ui/table/src/table.component.spec.ts
```diff
@@ -1,359 +1,24 @@
...
+import { createComponent } from '@terminus/ngx-tools/testing';
+
 import { TsTableDataSource } from './table-data-source';
 import { TsTableComponent } from './table.component';
 import { TsTableModule } from './table.module';
 
+import * as testComponents from '../testing/src/test-components';
+import {
+  expectTableToMatchContent, getHeaderCells, getHeaderRow,
+} from '../testing/src/test-helpers';
```
benjamincharity: These items should be imported from `@terminus/ui/table/testing`

terminus-ui/table/testing/src/public-api.ts
```diff
@@ -0,0 +1,2 @@
+export * from './test-components';
+export * from './test-helpers';
```
benjamincharity, line 2: We'll need an associated `package.json` file (See other test endpoints for exact contents and location)

terminus-ui/table/src/table.component.scss
```diff
@@ -136,7 +110,13 @@ $ts-row-hover: rgba(color(utility, xlight), .5);
...
   .ts-cell,
   .ts-header-cell {
-    display: table-cell;
+    align-items: center;
+    display: flex;
+    flex: 1;
+    flex-basis: 6em;
+    min-height: inherit;
+    overflow: hidden;
+    word-wrap: break-word;
```
benjamincharity, line 124: Let's add a style to `.ts-cell.ts-table--sticky` that sets border-right to `1px solid color(utility, xlight)` (matching other borders)

This will create a nice border for content that's scrolling beneath other cells:
Screen Shot 2019-08-13 at 3 31 47 PM

demo/app/components/table/table.component.html
```diff
@@ -14,123 +14,120 @@
...
-        [totalRecords]="resultsLength"
-        recordCountTooHighMessage="Please refine your filters."
-        (pageSelect)="onPageSelect($event)"
-        (recordsPerPageChange)="perPageChange($event)"
-      ></ts-paginator>
-    </div>
-  </div>
-</ts-card>
+<div class="example-container">
+  <p>Add columns to force a horizontal scroll.</p>
```
benjamincharity: I think we should default to a scrollable state so that product/ux can view it in our released demos.

terminus-ui/table/src/table.component.scss
```diff
@@ -91,9 +63,10 @@ $ts-row-hover: rgba(color(utility, xlight), .5);
   // Any row
   .ts-row,
   .ts-header-row {
+    align-items: center;
```
benjamincharity: I was wrong on this one.. looks like we need `stretch` as the value here so that when a large cell scrolls under a smaller cell, no content is visible underneath.

demo/app/components/table/table.component.html
```diff
@@ -14,123 +14,120 @@
...
+      <ts-header-cell *tsHeaderCellDef>
+        ID
+      </ts-header-cell>
+      <ts-cell *tsCellDef="let item">
+        {{ item.id }},
+      </ts-cell>
+    </ng-container>
+
+    <ts-header-row *tsHeaderRowDef="displayedColumns; sticky: true"></ts-header-row>
+    <!-- <ts-header-row *tsHeaderRowDef="displayedColumns">
```
benjamincharity: Comment can be removed

terminus-ui/table/src/table.component.scss
```diff
@@ -21,40 +21,12 @@ $ts-row-hover: rgba(color(utility, xlight), .5);
 .ts-table {
   @include reset;
   @include typography;
```
benjamincharity, line 23: OK, spent a while trying to figure out an odd scrolling issue where the pinning just 'stops':

Kapture 2019-08-13 at 16 28 13

Notice the stickiness ends right when the row ends (see the background color ending). The browser handles this incorrect layout fine on its own when there is no sticky trickery.

It looks like adding `display: inline-block` to `.ts-table` fixes this

benjamincharity (changes requested):
Found a couple cleanup items and a couple CSS items that weren't called out on the ticket but that would enhance this feature 👍

terminus-ui/table/testing/src/test-helpers.ts
```diff
@@ -0,0 +1,111 @@
+import { DataSource } from '@angular/cdk/collections';
```
benjamincharity, line 1: Really glad to have all this stuff pulled out of the spec file finally 🎉

demo/app/components/table/table.component.ts
shani-terminus: The columns were removed because the table automatically renders to the size of its container, and I needed to add columns to force a scroll. I'm looking into adding a minWidth to avoid that, and remove the note about adding columns, but it's not as straightforward as it sounds.

shani-terminus: Simply adding minWidth seems to only apply to the header:
Screen Shot 2019-08-14 at 7 37 06 AM

terminus-ui/table/src/table.component.scss
shani-terminus: This breaks the stickiness.

terminus-ui/table/testing/src/public-api.ts
shani-terminus: 👍 
I was modeling off of tooltip. That component doesn't have a package.json in its testing directory like other components, so I will put in a new issue.

terminus-ui/table/src/table.component.scss
benjamincharity: Can you clarify this? It was definitely working for me

demo/app/components/table/table.component.scss
```diff
@@ -0,0 +1,5 @@
+.example-container {
+  height: 400px;
+  overflow: auto;
+  width: 1500px;
```
benjamincharity: We should probably set this to a smaller width so that we can see the table scroll better. For me, this width was larger than my browser window.

demo/app/components/table/table.component.ts
benjamincharity: I'm not sure I'm following.. removing columns helped show scrolling? Seems like that would be opposite?

benjamincharity: On minWidth: Perhaps you're only targeting the header cells? I was able to get flex-basis working locally

terminus-ui/table/src/cell.ts
```diff
@@ -70,9 +70,21 @@ export class TsHeaderCellDefDirective extends CdkHeaderCellDef {}
...
+  ) {
     super(columnDef, elementRef);
     elementRef.nativeElement.classList.add(`ts-column-${columnDef.cssClassFriendlyName}`);
+
+    // tslint:disable-next-line no-any
+    const column: any = columnDef;
+
+    // Set inline style for min-width if passed in
+    if (column.minWidth) {
+      renderer.setStyle(elementRef.nativeElement, 'minWidth', column.minWidth);
```
benjamincharity: Is this in the correct PR? It seems it should be part of the resizable columns PR?

terminus-ui/table/src/table.component.md
```diff
@@ -314,6 +317,47 @@ directive `alignment` and pass it any valid TsTableColumnAlignment value (`left`
...
+  <ts-header-row *tsHeaderRowDef="displayedColumns; sticky: true"></ts-header-row>
+```
+
+
+## Sticky column
+
+Defining a sticky column will pin it to the left side as the table scrolls horizontally. Add the sticky parameter to the `<ng-container>` of the column definition. This can be applied to more than one column.
+
+```html
+  <ng-container tsColumnDef="updated" noWrap="true" minWidth="100px" sticky>
```
benjamincharity: Let's remove `noWrap` and `minWidth` to make the example only deal with the docs at hand.

terminus-ui/table/src/table.component.scss
```diff
@@ -136,7 +110,18 @@ $ts-row-hover: rgba(color(utility, xlight), .5);
...
+    background-color: color(pure);
+    display: flex;
+    flex: 1;
+    flex-basis: 6em;
+    min-height: inherit;
+    overflow: hidden;
+    word-wrap: break-word;
+
+    &.ts-table--sticky {
+      border-right: 1px solid color(utility, xlight);
```
benjamincharity: This is targeting both cells and header cells:

Screen Shot 2019-08-15 at 9 58 36 AM

We only want the border for body cells

terminus-ui/tsconfig.spec.json
```diff
@@ -11,43 +11,45 @@
...
-      "@terminus/ui/table"               : ["terminus-ui/table/src/public-api.ts"],
-      "@terminus/ui/toggle"              : ["terminus-ui/toggle/src/public-api.ts"],
-      "@terminus/ui/tooltip"             : ["terminus-ui/tooltip/src/public-api.ts"],
-      "@terminus/ui/utilities"           : ["terminus-ui/utilities/src/public-api.ts"],
-      "@terminus/ui/validation-messages" : ["terminus-ui/validation-messages/src/public-api.ts"],
-      "@terminus/ui/validators"          : ["terminus-ui/validators/src/public-api.ts"]
+      "@terminus/ui/autocomplete"         : ["terminus-ui/autocomplete/src/public-api.ts"],
+      "@terminus/ui/autocomplete/testing" : ["terminus-ui/autocomplete/testing/src/public-api.ts"],
+      "@terminus/ui/autofocus"            : ["terminus-ui/autofocus/src/public-api.ts"],
+      "@terminus/ui/button"                : ["terminus-ui/button/src/public-api.ts"],
```
benjamincharity: Indentation of the colon is off 1 character

benjamincharity (changes requested):
Looks like our layout is a little broken:

tableSmall

terminus-ui/table/src/cell.ts
shani-terminus: Yes. Without this, the column header widths are being set by flex, equally distributed within the viewport, and thus, not lining up with the column data.
Screen Shot 2019-08-15 at 11 07 16 AM

benjamincharity (approved):

benjamincharity:
Closing PR as this merge will be managed here: https://github.com/GetTerminus/terminus-ui/pull/1646

PR #1644, opened 2019-08-21
Feature(Autocomplete): show template value as chip

deanterm:
Shows the options template value as the chip instead of the `[value]` of the option. Also, allows the use of any type of object to be used as the `[value]` instead of just strings

### DIFF
--- a/terminus-ui/autocomplete/src/autocomplete.component.ts
+++ b/terminus-ui/autocomplete/src/autocomplete.component.ts
@@ -1,7 +1,6 @@
 // NOTE: A method must be used to dynamically format values for the UI
 // tslint:disable: template-no-call-expression
 import { ActiveDescendantKeyManager } from '@angular/cdk/a11y';
-import { CdkConnectedOverlay } from '@angular/cdk/overlay';
 import {
   AfterContentInit,
   AfterViewInit,
@@ -29,7 +28,6 @@ import { MatChipList } from '@angular/material';
 import { MatAutocompleteSelectedEvent } from '@angular/material/autocomplete';
 import { TsDocumentService } from '@terminus/ngx-tools/browser';
 import { coerceNumberProperty } from '@terminus/ngx-tools/coercion';
-import { isString } from '@terminus/ngx-tools/type-guards';
 import {
   hasRequiredControl,
   untilComponentDestroyed,
@@ -54,6 +52,7 @@ import {
   switchMap,
 } from 'rxjs/operators';
 
+import { isArray } from '@terminus/ngx-tools';
 import {
   TsAutocompletePanelComponent,
   TsAutocompletePanelSelectedEvent,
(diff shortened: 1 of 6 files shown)


terminus-ui/autocomplete/src/autocomplete.component.ts
```diff
@@ -45,7 +43,7 @@ import { TsStyleThemeTypes } from '@terminus/ui/utilities';
 import {
   BehaviorSubject,
   of,
-  Subject,
+  Subject, throwError,
```
atlwendy: Might want two lines...

terminus-ui/autocomplete/src/autocomplete.component.ts
```diff
@@ -806,22 +815,19 @@ export class TsAutocompleteComponent implements OnInit,
       }
 
       // Add to the collection
-      const newSelection = this.autocompleteSelections.slice();
-      newSelection.push(selection.option.value);
-      this.autocompleteSelections = newSelection;
+      this.autocompleteSelections = this.autocompleteSelections.concat(selection.option);
```
atlwendy, line 818: `.concat` returns a new array, right? we had to do the way we're doing because engage reducer wouldn't allow mutation...

terminus-ui/autocomplete/src/autocomplete.component.ts
```diff
@@ -782,10 +793,8 @@ export class TsAutocompleteComponent implements OnInit,
    * @param selection - The item to select
    */
   public autocompleteSelectItem(selection: TsAutocompletePanelSelectedEvent): void {
-    if (!isString(selection.option.value)) {
-      throw Error('The value passing into autocomplete has to be string type');
-    }
-    const isDuplicate = this.autocompleteSelections.indexOf(selection.option.value) >= 0;
+    const isDuplicate = this.autocompleteSelections
+      .findIndex(s => selection.option.value === s.value) >= 0;
```
atlwendy, line 797: do you always enforce `.value` attribution in option?

atlwendy (approved):
👍 nothing big jumps out...

terminus-ui/autocomplete/src/autocomplete.component.ts
benjamincharity: We really need to write a lint rule for this.. 😞

deanterm: let me fix. I think this import isn't needed either..

terminus-ui/autocomplete/src/autocomplete.component.ts
deanterm: as in requiring the user to specify a value on the `ts-option`?  good question, I can look to see if that's required

terminus-ui/autocomplete/src/autocomplete.component.ts
deanterm: yes'mm it does indeed

benjamincharity: Yep it does

benjamincharity (approved):
Everything looks good. ~Only question is what @atlwendy called out about value possibly not being set~

terminus-ui/autocomplete/src/autocomplete.component.ts
deanterm: fixed!

PR #1652, opened 2019-08-22
feat(autocomplete): add comparator/formatter. fix things

deanterm:


terminus-ui/autocomplete/src/autocomplete.component.spec.ts
```diff
@@ -350,14 +350,13 @@ describe(`TsAutocompleteComponent`, function() {
 
         // Verify the selection DID work
         chips = getAllChipInstances(fixture);
-        expect(chips.length).toEqual(1);
+        expect(chips.length).toEqual(2);
```
atlwendy, line 353: has the seeded value got updated? (used to be 1 element in an array)

terminus-ui/autocomplete/src/autocomplete.component.ts
```diff
@@ -848,25 +851,17 @@ export class TsAutocompleteComponent implements OnInit,
    *
    * @param value - The value of the item to remove
    */
-  public autocompleteDeselectItem(option: TsOptionComponent): void {
+  public autocompleteDeselectItem(option: any): void {
```
atlwendy: `any`! 😱

terminus-ui/autocomplete/src/autocomplete.component.ts
```diff
@@ -534,7 +547,9 @@ export class TsAutocompleteComponent implements OnInit,
             throw Error('ngModel must be an array of values');
           }
           this.autocompleteFormControl.setValue(this.ngControl.value);
-          this.setSelections();
+          if (!this.allowMultiple) {
+            this.searchQuery = this.displayFormatter(this.ngControl.value[0]);
```
atlwendy, line 553: why `searchQuery` needs to be updated only if it's single selection?

deanterm: because `searchQuery` is only really used to display a selected item in single selection only. We show chips in multiple selection instead

terminus-ui/autocomplete/src/autocomplete.component.spec.ts
deanterm: it shouldn't have been 1. the method is testing to allow duplicates. so 1 already exists (florida) and then we pick the same one again. so, 2 chips should be present. i guess before it wasn't updating the chips properly.

benjamincharity (approved):

benjamincharity:
@all-contributors please add @deanproxy for design and bug reports

benjamincharity:
@all-contributors please add @deanterm for design and bug reports

PR #1658, opened 2019-08-22
feat(Copy): Support for minimal and icon modes

benjamincharity:
1st commit:
- quick copy icon now appears on hover
- support minimal mode
- support icon mode
- remove scrollbar
- quick copy enabled by default
- updated demo for more control
- using TsTooltip in place of a browser tooltip for content

2nd commit:
- update ToC that was missed in previous PR

example

### DIFF
--- a/terminus-ui/copy/src/copy.component.ts
+++ b/terminus-ui/copy/src/copy.component.ts
@@ -3,14 +3,28 @@ import {
   Component,
   ElementRef,
   Input,
-  ViewChild,
+  isDevMode, QueryList,
+  ViewChild, ViewChildren,
   ViewEncapsulation,
 } from '@angular/core';
 import {
   TsDocumentService,
   TsWindowService,
 } from '@terminus/ngx-tools/browser';
-import { TsStyleThemeTypes } from '@terminus/ui/utilities';
+import { TsTooltipComponent } from '@terminus/ui/tooltip';
+import {
+  TsStyleThemeTypes,
+  TsUILibraryError,
+} from '@terminus/ui/utilities';
+
+
+
+
+export type TsCopyDisplayFormat
+  = 'standard'
+  | 'minimal'
+  | 'icon'
+;
 
 
 /**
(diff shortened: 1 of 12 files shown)


terminus-ui/copy/src/copy.component.ts
```diff
@@ -83,7 +111,29 @@ export class TsCopyComponent {
...
+  @Input()
+  public set format(value: TsCopyDisplayFormat) {
+    this._format = value ? value : 'standard';
+
+    if (this.format === 'icon' && !this.enableQuickCopy) {
+      this.enableQuickCopy = true;
+
+      // istanbul ignore else
+      if (isDevMode()) {
+        throw new TsUILibraryError(`'enableQuickCopy' must be set to 'true' when using the icon only display mode`);
```
atlwendy, line 128: `TsUILibraryError`! Nice

terminus-ui/utilities/src/error/ui-error.ts
```diff
@@ -0,0 +1,11 @@
+/**
+ * A Terminus UI specific Error
+ */
+export class TsUILibraryError extends Error {
+  constructor(message: string) {
+    super(message);
+    this.message = message;
+    this.name = 'TsUILibraryError';
+    this.stack = (new Error(message)).stack;
```
atlwendy, line 9: 👍

atlwendy (approved):

terminus-ui/copy/src/copy.component.ts
benjamincharity: Yeah! Originally created it for two reasons,
a) It's nice for the stack trace to call out our library specifically when debugging
b) I had read about a way to create a more useful stack trace - which I couldn't get working initially but hope to revisit.

Ended up still using for `a)` at the very least

benjamincharity: Only real difference in use is that we can do `throw Error(...)` but with ours we _must_ call new `throw new TsUILibraryError(...)`

PR #1660, opened 2019-08-23
fix(date-range): fix invalid date range when changing text

deanterm:
When a user changes the number in the middle of the date, it will cause the mask to get wonky until they type the date back out.  Getting the value of after the mask has had a chance to fix the wonky value fixes this issue.  Also, turning off `keepCharPositions` helps to prevent the cursor from walking into weird sections during the textual editing.

### DIFF
--- a/terminus-ui/input/src/input.component.spec.ts
+++ b/terminus-ui/input/src/input.component.spec.ts
@@ -2,13 +2,11 @@
 import { AutofillMonitor } from '@angular/cdk/text-field';
 import {
   ElementRef,
-  Provider,
   Type,
 } from '@angular/core';
 import {
   ComponentFixture,
   fakeAsync,
-  TestBed,
   tick,
 } from '@angular/core/testing';
 import {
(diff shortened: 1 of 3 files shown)


terminus-ui/input/src/input.component.ts
```diff
@@ -1168,13 +1173,16 @@ export class TsInputComponent implements
       // Update the mask.
       this.textMaskInputElement.update(trimmedValue);
 
+      // Reset the value after the mask has had a chance to update it.
+      value = target.value;
```
atlwendy, line 1177: the meat is here...

atlwendy (approved):
any specs need to be updated?

terminus-ui/input/src/input.component.spec.ts
```diff
@@ -901,6 +899,17 @@ describe(`TsInputComponent`, function() {
...
 
+    test('should return if target is not set', () => {
+      const fixture = createComponent(TestComponents.DateFilter);
+      const component = fixture.componentInstance.inputComponent;
+      component._valueChange.emit = jest.fn();
+      component.selected.emit = jest.fn();
+      component.onInput(null as any);
+
+      expect(component._valueChange.emit).not.toHaveBeenCalled();
+      expect(component.selected.emit).not.toHaveBeenCalledWith();
```
benjamincharity: Should this line also be `toHaveBeenCalled()` vs `...With()`?

terminus-ui/input/src/input.component.ts
deanterm: the rest is just potatoes

terminus-ui/input/src/input.component.spec.ts
deanterm: yes. copy/paste error. fixed

benjamincharity (approved):

Issue #285, opened 2017-11-02
Atom: Chip


benjamincharity:
### Purpose

We've recently been solving for some complex user interaction cases around allowing a user to create large lists (ie 50+) from even larger lists (ie 50,000+). When approaching that problem with a toolset based around form controls we have struggled to create a good experience. I don't find that surprising since there are no standard web inputs that exist to solve this need.

In the mentioned user need, the user is not really selecting (select) or free-writing options (autocomplete) - they are really _searching_ and adding to a collection.

The hope is that if we can allow a collection to be shown separately from the actual form of user input, this would open up a wider array of creative solutions around how the user makes selections when creating these large lists.

In turn, this will also allow us to fix a few issues in our existing autocomplete where it is not functioning as a true autocomplete as users expect.

---

### Chip

- [ ] Should support removal icon/action
  - [ ] `[isRemovable]="true"`
  - [ ] When removable a clickable `x` icon should be present at the right of the chip
  - [ ] Should expose method to allow programmatic removal `remove()`
  - [ ] Should not allow removal when the parent collection is in readonly mode.
- [ ] Should support selection
  - [ ] This can be disabled `[isSelectable]="false"`
  - [ ] Selection ability should take into account the internal property `chipCollectionIsSelectable` which is set by the chip collection when the collection is marked as selectable.
  - [ ] Should reflect selected state via styles
- [ ] Should support focus
  - [ ] Should support programmatic focus via input `[isFocused]="true"` and public method `focus()`
  - [ ] Should reflect focused state via styles
- [ ] Should support selection
  - [ ] Should support programmatic selection via input `[isSelected]="true"` and public method `select()`
  - [ ] Should support multiple selection `[allowMultiple]="true"`
- [ ] Can be disabled `[isDisabled]="true"`
  - [ ] All interaction is disabled
  - [ ] If parent collection is disabled, the chip should be disabled too
  - [ ] Should be styled to appear disabled
- [ ] Events
  - [ ] `selectionChange: TsChipSelectionChange`
  - [ ] `destroyed: TsChipEvent`
  - [ ] `removed: TsChipEvent`
- [ ] Should have accessibility support:
  - [ ] `aria-role="option"`
  - [ ] `aria-disabled="true"` true when `isDisabled` is true
  - [ ] `aria-selected="true"` true when `isSelected` is true
- [ ] When chip is removed, the new chip that takes it's place should gain selection
[...]

PR #1666, opened 2019-08-27
Chip Component

atlwendy:
Here is the requirement:

#285

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+.vscode/
 dist/
 node_modules/
 out-tsc/
--- a/demo/app/components/chip/chip.component.ts
+++ b/demo/app/components/chip/chip.component.ts
@@ -1,0 +1,36 @@
+import { Component, OnInit, ViewChild, AfterViewInit, QueryList, ElementRef, TemplateRef } from '@angular/core';
+import {
+  TsChipCollectionComponent,
+  TsChipComponent,
+ } from '@terminus/ui/chip';
+
+@Component({
+  selector: 'demo-chip',
+  templateUrl: './chip.component.html',
+})
+export class ChipComponent implements OnInit {
+
+  options = ['banana', 'apple', 'orange', 'pear'];
+
+  constructor() { }
+
+  ngOnInit() {
+  }
+
+  removingOption(value: string) {
+    const ctrlValue = this.options;
+    const index = ctrlValue.indexOf(value);
+    const selections = ctrlValue.slice();
+    // If not found
+    if (index < 0) {
+      return;
+    }
+    selections.splice(index, 1);
+    this.options = selections;
+  }
+
+  change(value: string | string[]) {
+    console.log('DEMO: collection change triggered value: ', value);
+  }
+
+}
--- a/terminus-ui/autocomplete/src/autocomplete.component.html
+++ b/terminus-ui/autocomplete/src/autocomplete.component.html
@@ -15,25 +15,17 @@
   <div class="ts-autocomplete__input-wrap">
 
     <ng-container *ngIf="allowMultiple">
-      <mat-chip-list #chipList="matChipList">
-        <mat-chip
-          *ngFor="let option of autocompleteFormControl.value; trackBy: trackByFn"
-          class="qa-autocomplete-chip"
-          [removable]="true"
-          (removed)="autocompleteDeselectItem(option)"
-        >
-          <span class="ts-autocomplete-chip-value">
-            {{ displayFormatter(option) }}
-          </span>
-
-          <ts-icon matChipRemove>
-            cancel
-          </ts-icon>
-        </mat-chip>
-
+      <ts-chip-collection
+        [chipsInput]="autocompleteFormControl.value"
+        [allowMultipleSelections]="true"
+        [displayFormatter]="displayFormatter"
+        [isDisabled]="false"
+        [isReadonly]="false"
+        (removed)="autocompleteDeselectItem($event)"
+        (tabUpdateFocus)="focusInput()"
+        #chipCollection="tsChipCollection">
         <input
           class="ts-autocomplete__input qa-select-autocomplete-input"
-          [matChipInputFor]="chipList"
           [tsAutocompleteTrigger]="auto"
           [reopenAfterSelection]="reopenAfterSelection"
           [attr.id]="id"
@@ -41,12 +33,10 @@
           [readonly]="isDisabled ? 'true' : null"
           (ngModelChange)="querySubject.next($event)"
           (blur)="handleInputBlur($event)"
-          (keydown)="$event.stopPropagation()"
           #input
         />
-
-        <ng-template *ngTemplateOutlet="spinnerTemplate"></ng-template>
-      </mat-chip-list>
+      </ts-chip-collection>
+      <ng-template *ngTemplateOutlet="spinnerTemplate"></ng-template>
     </ng-container>
 
     <ng-container *ngIf="!allowMultiple">
--- a/terminus-ui/autocomplete/src/autocomplete.component.ts
+++ b/terminus-ui/autocomplete/src/autocomplete.component.ts
@@ -23,7 +23,6 @@ import {
   FormControl,
   NgControl,
 } from '@angular/forms';
-import { MatChipList } from '@angular/material';
 import { MatAutocompleteSelectedEvent } from '@angular/material/autocomplete';
 import { TsDocumentService } from '@terminus/ngx-tools/browser';
 import { coerceNumberProperty } from '@terminus/ngx-tools/coercion';
@@ -52,6 +51,7 @@ import {
 } from 'rxjs/operators';
 
 import { isArray } from '@terminus/ngx-tools';
+import { TsChipCollectionComponent } from '@terminus/ui/chip';
 import {
   TsAutocompletePanelComponent,
   TsAutocompletePanelSelectedEvent,
@@ -234,8 +234,8 @@ export class TsAutocompleteComponent implements OnInit,
   /**
    * Access the chip list
    */
-  @ViewChild('chipList', { static: false })
-  public chipList: MatChipList | undefined;
+  @ViewChild('chipCollection', { static: false })
+  public chipCollection: TsChipCollectionComponent | undefined;
 
   /**
    * Access the actual HTML element
@@ -847,6 +847,12 @@ export class TsAutocompleteComponent implements OnInit,
     this.selectionChange.emit(new TsAutocompleteChange(this, this.autocompleteFormControl.value));
   }
 
+  /**
+   * Chip component emit a focusInput event, autocomplete puts focus on input field.
+   */
+  public focusInput(): void {
+    this.focus();
+  }
 
   /**
    * Deselect an item
@@ -880,7 +886,6 @@ export class TsAutocompleteComponent implements OnInit,
     this.optionDeselected.emit(new TsAutocompleteChange(this, option));
     this.selectionChange.emit(new TsAutocompleteChange(this, options));
   }
-
 
   /**
    * Function for tracking for-loops changes
--- a/terminus-ui/autocomplete/src/autocomplete.module.ts
+++ b/terminus-ui/autocomplete/src/autocomplete.module.ts
@@ -6,11 +6,11 @@ import {
   FormsModule,
   ReactiveFormsModule,
 } from '@angular/forms';
-import { MatChipsModule } from '@angular/material/chips';
 import { MatRippleModule } from '@angular/material/core';
 import { MatInputModule } from '@angular/material/input';
 import { MatProgressSpinnerModule } from '@angular/material/progress-spinner';
 import { TsCheckboxModule } from '@terminus/ui/checkbox';
+import { TsChipModule } from '@terminus/ui/chip';
 import { TsFormFieldModule } from '@terminus/ui/form-field';
 import { TsIconModule } from '@terminus/ui/icon';
 import { TsInputModule } from '@terminus/ui/input';
@@ -33,12 +33,12 @@ export * from './autocomplete-panel/autocomplete-panel.component';
     FormsModule,
     ReactiveFormsModule,
     FlexLayoutModule,
-    MatChipsModule,
     MatInputModule,
     MatProgressSpinnerModule,
     MatRippleModule,
     OverlayModule,
     TsCheckboxModule,
+    TsChipModule,
     TsIconModule,
     TsFormFieldModule,
     TsOptionModule,
--- a/terminus-ui/chip/src/chip-collection.component.html
+++ b/terminus-ui/chip/src/chip-collection.component.html
@@ -1,0 +1,17 @@
+<div 
+  class="ts-chip-collection-wrapper"
+  fxLyaout="row"
+  [ngClass]="(orientation==='vertical') ? 'ts-chip-collection--stacked':''"
+>
+  <ts-chip
+    *ngFor="let chip of chipsInput; trackBy: trackByFn"
+    [isRemovable]="!isReadonly"
+    [isDisabled]="isDisabled"
+    [isSelectable]="isSelectable"
+    [value]="displayFormatter(chip)"
+    (removed)="remove(chip)"
+    (clicked)="clickSelect($event)"
+  >
+  </ts-chip>
+  <ng-content></ng-content>
+</div>
--- a/terminus-ui/chip/src/chip-collection.component.spec.ts
+++ b/terminus-ui/chip/src/chip-collection.component.spec.ts
@@ -1,0 +1,476 @@
+import { Type } from '@angular/core';
+import {
+  ComponentFixture,
+  fakeAsync,
+  tick,
+} from '@angular/core/testing';
+import {
+  FormsModule,
+  ReactiveFormsModule,
+} from '@angular/forms';
+import { By } from '@angular/platform-browser';
+import { NoopAnimationsModule } from '@angular/platform-browser/animations';
+import { KEYS } from '@terminus/ngx-tools/keycodes';
+import {
+  createComponent as createComponentInner,
+  createKeyboardEvent,
+  createMouseEvent,
+  dispatchEvent,
+  dispatchMouseEvent,
+  typeInElement,
+} from '@terminus/ngx-tools/testing';
+import { TsAutocompleteModule } from '@terminus/ui/autocomplete';
+import {
+  TsChipCollectionComponent,
+  TsChipModule,
+} from '@terminus/ui/chip';
+import * as testComponents from '@terminus/ui/chip/testing';
+// eslint-disable-next-line no-duplicate-imports
+import {
+  getAllChipDebugElements,
+  getChipCollectionInstance,
+  getChipElement,
+  getChipInstance,
+} from '@terminus/ui/chip/testing';
+import { TsOptionModule } from '@terminus/ui/option';
+
+function createComponent<T>(component: Type<T>): ComponentFixture<T> {
+  const moduleImports = [
+    FormsModule,
+    ReactiveFormsModule,
+    TsChipModule,
+    TsAutocompleteModule,
+    TsOptionModule,
+    NoopAnimationsModule,
+  ];
+
+  return createComponentInner(component, undefined, moduleImports);
+}
+
+describe(`TsChipCollection`, function() {
+  let fixture;
+
+  let chipElements;
+  let chipDebugElement;
+  let chipNativeElement;
+  let chipCollectionDebugElement;
+  let chipCollectionElement;
+  let chipCollectionInstance;
+  let chipCollectionNativeElement;
+  let chipInsideElement;
+  let chipInstance;
+  let chips;
+  let manager;
+  let nativeChips;
+  let nativeInput;
+  let testComponent;
+  let BACKSPACE_EVENT: KeyboardEvent;
+  let BACKSPACE_EVENT_CHIP: KeyboardEvent;
+
+  function setupStandardCollection(exampleComponent) {
+    fixture = createComponent(testComponents[exampleComponent]);
+    fixture.detectChanges();
+    chipCollectionElement = fixture.debugElement.query(By.directive(TsChipCollectionComponent));
+    chipCollectionNativeElement = chipCollectionElement.nativeElement;
+
+    chipCollectionInstance = getChipCollectionInstance(fixture);
+    chipInstance = getChipInstance(fixture);
+    chipElements = getAllChipDebugElements(fixture);
+    chipDebugElement = fixture.debugElement.query(By.css('.ts-chip'));
+    chipNativeElement = chipDebugElement.nativeElement;
+    chipInsideElement = fixture.debugElement.query(By.css('.c-chip')).nativeElement;
+    nativeChips = chipCollectionNativeElement.querySelectorAll('ts-chip');
+    chips = chipCollectionInstance.chips;
+    manager = chipCollectionInstance.keyManager;
+    testComponent = fixture.debugElement.componentInstance;
+  }
+
+  test(`should exist`, function() {
+    setupStandardCollection('RegularChip');
+
+    expect(chipNativeElement).toBeTruthy();
+  });
+
+  test(`should disable chip if isDisabled set`, function() {
+    setupStandardCollection('DisabledChip');
+
+    expect(chipInstance.isDisabled).toEqual(true);
+    expect(chipNativeElement.classList).toContain('ts-chip-disabled');
+  });
+
+  test(`should disable chip if isReadonly set`, function() {
+    setupStandardCollection('ReadonlyChip');
+    chipCollectionInstance.isReadonly = true;
+    fixture.detectChanges();
+
+    expect(chipInstance.isRemovable).toBeFalsy();
+    expect(chipNativeElement.classList).not.toContain('c-chip-removable');
+  });
+
+  test(`should be able to select`, fakeAsync(function() {
+    setupStandardCollection('RegularChip');
+
+    chipInstance.select();
+    const chip1 = chipElements[0].nativeElement;
+    const chip2 = chipElements[1].nativeElement;
+    chip2._selected = false;
+    tick(1000);
+    fixture.detectChanges();
+
+    expect(chip1.classList).toContain('ts-chip-selected');
+    expect(chip2.classList).not.toContain('ts-chip-selected');
+  }));
+
+  test(`should set value properly`, function() {
+    fixture = createComponent(testComponents.NoValueChip);
+    fixture.detectChanges();
+    const instance = getChipCollectionInstance(fixture);
+
+    expect(instance.value).toEqual([]);
+    instance.value = ['banana'];
+    fixture.detectChanges();
+    expect(instance.value).toEqual(['banana']);
+  });
+
+  test(`should be able to set isSelectable from chip collection and populate to chips`, fakeAsync(function() {
+    setupStandardCollection('RegularChip');
+
+    chipCollectionInstance.isSelectable = true;
+    const chip = chips.toArray()[0];
+    fixture.detectChanges();
+    chipInstance.select();
+    tick(1000);
+    fixture.detectChanges();
+
+    expect(chip.chipCollectionSelectable).toBeTruthy();
+  }));
+
+  test(`should emit remove event when chip is removed`, fakeAsync(function() {
+    setupStandardCollection('RegularChip');
+    const chip = fixture.debugElement.query(By.css('.ts-chip')).nativeElement;
+    const chipRemovalButton = chip.querySelector('.ts-chip-remove');
+    chipCollectionInstance.remove = jest.fn();
+    chipCollectionInstance.removed.subscribe(event => {
+      expect(event).toEqual('apple');
+    });
+
+    dispatchMouseEvent(chipRemovalButton, 'click');
+    tick(1000);
+    fixture.detectChanges();
+
+    expect(chipCollectionInstance.remove).toHaveBeenCalled();
+  }));
+
+  test(`should focus on nothing if blur and no focused chip`, () => {
+    setupStandardCollection('RegularChip');
+    chipCollectionInstance.blur();
+    fixture.detectChanges();
+    expect(manager.activeItemIndex).toBe(-1);
+  });
+
+  test(`should deselect current selected chip if select another chip under allowMultipleSelections false`, () => {
+    setupStandardCollection('NotAllowMultipleSelections');
+    const allChips = chips.toArray();
+    allChips[0].selected = true;
+    fixture.detectChanges();
+
+    allChips[1].selected = true;
+    fixture.detectChanges();
+    expect(allChips[0].selected).toBeFalsy();
+
+  });
+
+  describe(`focus`, function() {
+    test(`should return undefined if isDisabled`, function() {
+      setupStandardCollection('DisabledChip');
+      expect(chipCollectionInstance.focus()).toEqual(undefined);
+    });
+
+    test(`should focus the first non-disabled chip`, function() {
+      setupStandardCollection('RegularChip');
+      chipCollectionInstance.keyManager.setFirstItemActive = jest.fn();
+      chipCollectionInstance.focus();
+      fixture.detectChanges();
+
+      expect(chipCollectionInstance.keyManager.setFirstItemActive).toHaveBeenCalled();
+    });
+
+    test(`should set chip focus state if focus method is triggered`, function() {
+      setupStandardCollection('RegularChip');
+      chipInstance.focus();
+      expect(chipInstance.hasFocus).toBeTruthy();
+      expect(chipCollectionInstance.focused).toBeTruthy();
+    });
+
+  });
+
+  describe(`tabIndex`, () => {
+    beforeEach(() => {
+      setupStandardCollection('Tabindex');
+    });
+
+    test(`should set tabindex if not disabled`, () => {
+      expect(chipCollectionNativeElement.getAttribute('tabindex')).toEqual('4');
+      fixture.componentInstance.index = -1;
+      fixture.detectChanges();
+      expect(chipCollectionNativeElement.getAttribute('tabindex')).toEqual('-1');
+      expect(chipCollectionInstance.tabIndex).toEqual(-1);
+    });
+
+    test(`should have tabindex null if disabled`, () => {
+      fixture.componentInstance.isDisabled = true;
+      fixture.detectChanges();
+      expect(chipCollectionNativeElement.getAttribute('tabindex')).toBeFalsy();
+    });
+  });
+
+  test(`should not allow keyboard focus when there's 0 chip`, fakeAsync(() => {
+    fixture = createComponent(testComponents.NoChip);
+    chipCollectionInstance = getChipCollectionInstance(fixture);
+    fixture.detectChanges();
+    tick(1000);
+    expect(chipCollectionInstance._tabIndex).toBe(-1);
+  }));
+
+
+  test('should not have the aria-selected attribute when is not selectable', () => {
+    setupStandardCollection('isSelectable');
+    testComponent = fixture.debugElement.componentInstance;
+    testComponent.isSelectable = false;
+    fixture.detectChanges();
+
+    const chipsValid = chips.toArray().every(chip => !chip.isSelectable && !chip._elementRef.nativeElement.hasAttribute('aria-selectable'));
+
+    expect(chipsValid).toBe(true);
+  });
+
+  describe(`ID`, function() {
+    beforeEach(() => {
+      setupStandardCollection('Id');
+    });
+
+    test(`should get UID as its id`, () => {
+      fixture.componentInstance.myId = undefined as any;
+      fixture.detectChanges();
+      expect(chipCollectionNativeElement.getAttribute('id')).toEqual(expect.stringContaining('ts-chip-collection-'));
+      expect(chipCollectionInstance.id).toEqual(expect.stringContaining('ts-chip-collection-'));
+    });
+
+    test(`should set custom id`, () => {
+      fixture.componentInstance.myId = 1;
+      fixture.detectChanges();
+      expect(chipCollectionInstance.id).toEqual(1);
+    });
+
+  });
+
+  describe(`used in autocomplete component`, function() {
+    beforeEach(() => {
+      setupStandardCollection('Autocomplete');
+      nativeInput = fixture.nativeElement.querySelector('input');
+      BACKSPACE_EVENT = createKeyboardEvent('keydown', KEYS.BACKSPACE, nativeInput);
+      BACKSPACE_EVENT_CHIP = createKeyboardEvent('keydown', KEYS.BACKSPACE, chipCollectionNativeElement);
+    });
+
+    test(`should focus on last item if input is empty and BACKSPACE is used`, () => {
+
+      // Focus the input
+      nativeInput.focus();
+      expect(manager.activeItemIndex).toBe(-1);
+
+      // Press the BACKSPACE key
+      chipCollectionInstance.keydown(BACKSPACE_EVENT);
+      fixture.detectChanges();
+
+      // It focuses the last chip
+      expect(manager.activeItemIndex).toEqual(chips.length - 1);
+      chipCollectionInstance.keydown(BACKSPACE_EVENT_CHIP);
+      fixture.detectChanges();
+      const nodeName = document.activeElement ? document.activeElement.nodeName : '';
+      // Focus still on the chip
+      expect(nodeName).toEqual('TS-CHIP');
+    });
+
+    test(`should focus on input if input field is not empty and BACKSPACE is used`, () => {
+      // Focus the input
+      nativeInput.focus();
+      // Type letter into the input field
+      typeInElement('a', nativeInput);
+
+      // Press the BACKSPACE key
+      chipCollectionInstance.keydown(BACKSPACE_EVENT);
+
+      // It focuses on the input field
+      fixture.detectChanges();
+      const nodeName = document.activeElement ? document.activeElement.nodeName : '';
+      expect(nodeName).toBe('INPUT');
+    });
+
+    test(`should emit removed event when a chip removed`, () => {
+      const allchips = chips.toArray();
+      chipCollectionInstance.removed.emit = jest.fn();
+      fixture.detectChanges();
+      chipCollectionInstance.remove(allchips[0]);
+      fixture.detectChanges();
+
+      expect(chipCollectionInstance.removed.emit).toHaveBeenCalled();
+    });
+
+  });
+
+  describe(`keydown`, function() {
+    beforeEach(() => {
+      setupStandardCollection('StandardChipCollection');
+    });
+
+    test(`should set first item active when use HOME key`, fakeAsync(function() {
+      const element = getChipInstance(fixture)['elementRef'].nativeElement as HTMLElement;
+      chipCollectionInstance.keyManager.setFirstItemActive = jest.fn();
+
+      const event = document.createEvent('KeyboardEvent');
+      event.initEvent('keydown', true, false);
+      Object.defineProperties(event, { code: { get: () => KEYS.HOME.code } });
+
+      fixture.detectChanges();
+      element.dispatchEvent(event);
+      fixture.detectChanges();
+
+      expect(chipCollectionInstance.keyManager.setFirstItemActive).toHaveBeenCalled();
+    }));
+
+    test(`should focus on last item when press END`, () => {
+      const firstNativeChip = nativeChips[0] as HTMLElement;
+
+      const array = chips.toArray();
+      const lastIndex = array.length - 1;
+      const firstItem = array[0];
+      const END_EVENT = createKeyboardEvent('keydown', KEYS.END, firstNativeChip);
+
+      firstItem.focus();
+      chipCollectionInstance.keydown(END_EVENT);
+      fixture.detectChanges();
+      expect(manager.activeItemIndex).toEqual(lastIndex);
+    });
+
+    test(`should select OR deselect all items with CTRL + A`, fakeAsync(function() {
+      chipCollectionInstance.allowMultipleSelections = true;
+      fixture.detectChanges();
+      tick(1000);
+      const element = getChipElement(fixture);
+
+      const event = createKeyboardEvent('keydown', KEYS.A, element);
+      Object.defineProperty(event, 'ctrlKey', { get: () => true });
+      dispatchEvent(element, event);
+      fixture.detectChanges();
+      tick(3000);
+
+      expect(chipCollectionInstance.selectionModel.selected.length).toEqual(1);
+      expect(event.defaultPrevented).toEqual(true);
+
+      dispatchEvent(element, event);
+      fixture.detectChanges();
+
+      expect(chipCollectionInstance.selectionModel.selected.length).toEqual(0);
+    }));
+  });
+
+  describe(`updateFocus`, () => {
+
+    beforeEach(() => {
+      setupStandardCollection('StandardChipCollection');
+    });
+
+    test(`should focus on last chip when END key pressed`, () => {
+      const firstNativeChip = nativeChips[0] as HTMLElement;
+      const array = chips.toArray();
+      const lastIndex = array.length - 1;
+      const firstItem = array[0];
+      const END_EVENT = createKeyboardEvent('keydown', KEYS.END, firstNativeChip);
+
+      firstItem.focus();
+      chipCollectionInstance.keydown(END_EVENT);
+      fixture.detectChanges();
+      expect(manager.activeItemIndex).toEqual(lastIndex);
+    });
+
+    test('should focus the next chip when one is removed ', () => {
+      jest.useFakeTimers();
+      testComponent = fixture.debugElement.componentInstance;
+
+      const array = chips.toArray();
+      const lastIndex = array.length - 1;
+      const lastItem = array[lastIndex];
+
+      // Focus the last item
+      lastItem.focus();
+
+      // Destroy the last item
+      testComponent.options.pop();
+      fixture.detectChanges();
+      jest.advanceTimersByTime(2000);
+
+      // It focuses the next-to-last item
+      expect(manager.activeItemIndex).toEqual(lastIndex - 1);
+    });
+  });
+
+  test('should focus the chip collection if the last focused item is removed', fakeAsync(function() {
+    setupStandardCollection('OneChip');
+    const chip = chipCollectionInstance.chips.toArray()[0];
+    testComponent = fixture.debugElement.componentInstance;
+    chip.hasFocus = true;
+    fixture.detectChanges();
+    chipCollectionInstance.focus = jest.fn();
+
+    // Remove the chip
+    testComponent.options = [];
+    fixture.detectChanges();
+    tick(1000);
+    fixture.detectChanges();
+
+    // Focus is now on chip collection
+    expect(chipCollectionInstance.focus).toHaveBeenCalled();
+  }));
+
+  describe(`shift+click`, () => {
+    let mouseEvent: MouseEvent;
+    let chip;
+
+    beforeEach(() => {
+      setupStandardCollection('StandardChipCollection');
+      chip = chipCollectionInstance.chips.toArray()[0];
+      mouseEvent = createMouseEvent('click');
+      Object.defineProperty(mouseEvent, 'shiftKey', { get: () => true });
+    });
+    test(`should select a chip if it's not selected and allowMultiple being true`, fakeAsync(() => {
+      chipCollectionInstance.allowMultipleSelections = true;
+      // If chip is not selected, shift click will select the chip
+      chip.selected = false;
+      fixture.detectChanges();
+      dispatchEvent(chipInsideElement, mouseEvent);
+      fixture.detectChanges();
+      tick(3000);
+      expect(chip.selected).toBeTruthy();
+
+      // If chip is selected, shift click will deselect the chip
+      chip.selected = true;
+      fixture.detectChanges();
+      dispatchEvent(chipInsideElement, mouseEvent);
+      fixture.detectChanges();
+      tick(3000);
+      expect(chip.selected).toBeFalsy();
+    }));
+
+    test(`should not select a chip if allowMultiple being false`, fakeAsync(() => {
+      chipCollectionInstance.allowMultipleSelections = false;
+      chip.selected = false;
+      fixture.detectChanges();
+
+      dispatchEvent(chipInsideElement, mouseEvent);
+      fixture.detectChanges();
+      tick(3000);
+      expect(chip.selected).toBeFalsy();
+    }));
+  });
+
+});
--- a/terminus-ui/chip/src/chip-collection.component.ts
+++ b/terminus-ui/chip/src/chip-collection.component.ts
@@ -1,0 +1,640 @@
+// NOTE: A method must be used to dynamically format values for the UI
+// tslint:disable: template-no-call-expression
+import { FocusKeyManager } from '@angular/cdk/a11y';
+import { Directionality } from '@angular/cdk/bidi';
+import { coerceBooleanProperty } from '@angular/cdk/coercion';
+import { SelectionModel } from '@angular/cdk/collections';
+import {
+  AfterViewInit,
+  ChangeDetectionStrategy,
+  ChangeDetectorRef,
+  Component,
+  ElementRef,
+  EventEmitter,
+  Input,
+  NgZone,
+  OnDestroy,
+  OnInit,
+  Optional,
+  Output,
+  QueryList,
+  ViewChildren,
+  ViewEncapsulation,
+} from '@angular/core';
+import { untilComponentDestroyed } from '@terminus/ngx-tools';
+import { KEYS } from '@terminus/ngx-tools/keycodes';
+import {
+  merge,
+  Observable,
+  Subject,
+  Subscription,
+} from 'rxjs';
+import {
+  startWith,
+  takeUntil,
+} from 'rxjs/operators';
+import {
+  TsChipComponent,
+  TsChipEvent,
+  TsChipSelectionChange,
+} from './chip.component';
+
+
+// Increasing integer for generating unique ids for chip-collection components.
+let nextUniqueId = 0;
+
+/**
+ * Change event object that is emitted when the chip collection value has changed.
+ */
+export class TsChipCollectionChange {
+  constructor(
+    // Chip collection that emitted the event.
+    public source: TsChipCollectionComponent,
+    // Value of the chip collection when the event was emitted.
+    // tslint:disable-next-line no-any
+    public value: string[],
+  ) { }
+}
+
+
+/**
+ * Component that is used to group instances of {@link TsChipComponent}s
+ *
+ * @example
+ * <ts-chip-collection
+ *              [allowMultipleSelections]="true"
+ *              [isDisabled]="false"
+ *              [isReadonly]="false"
+ *              [isRemovable]="true"
+ *              [isSelectable]="false"
+ *              [orientation]="horizontal"
+ *              (collectionChange)="collectionChange($event)"
+ * ></ts-chip-collection>
+ *
+ * <example-url>https://getterminus.github.io/ui-demos-release/components/chip</example-url>
+ */
+@Component({
+  selector: 'ts-chip-collection',
+  templateUrl: `./chip-collection.component.html`,
+  exportAs: 'tsChipCollection',
+  host: {
+    '[attr.tabindex]': 'isDisabled ? null : _tabIndex',
+    '[attr.aria-describedby]': 'ariaDescribedby || null',
+    '[attr.aria-disabled]': 'isDisabled',
+    '[attr.aria-multiselectable]': 'allowMultipleSelections',
+    '[attr.aria-orientation]': 'ariaOrientation',
+    '[attr.aria-readonly]': 'isReadonly',
+    '[attr.aria-required]': 'false',
+    '[attr.aria-selectable]': 'isSelectable',
+    '[attr.role]': 'role',
+    '[class.ts-chip-collection-disabled]': 'isDisabled',
+    '[class.ts-chip-collection-selectable]': 'isSelectable',
+    'class': 'ts-chip-collection',
+    '(focus)': 'focus()',
+    '(blur)': 'blur()',
+    '(keydown)': 'keydown($event)',
+    '[id]': 'id',
+  },
+  styleUrls: ['./chip.component.scss'],
+  encapsulation: ViewEncapsulation.None,
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class TsChipCollectionComponent implements AfterViewInit, OnInit, OnDestroy {
+
+  /**
+   * Subject that emits when the component has been destroyed.
+   */
+  private destroyed = new Subject<void>();
+
+  /**
+   * Subscription to focus changes in the chips.
+   */
+  private chipFocusSubscription: Subscription | null = null;
+
+  /**
+   * Subscription to blur changes in the chips.
+   */
+  private chipBlurSubscription: Subscription | null = null;
+
+  /**
+   * Subscription to selection changes in chips.
+   */
+  private chipSelectionSubscription: Subscription | null = null;
+
+  /**
+   * Subscription to remove changes in chips.
+   */
+  private chipRemoveSubscription: Subscription | null = null;
+
+  /**
+   * Uid of the chip collection
+   */
+  protected uid = `ts-chip-collection-${nextUniqueId++}`;
+
+  /**
+   * The aria-describedby attribute on the chip collection for improved a11y.
+   */
+  public ariaDescribedby!: string;
+
+  /**
+   * User defined tab index.
+   * When it is not null, use user defined tab index. Otherwise use _tabIndex
+   */
+  public _userTabIndex: number | null = null;
+
+  /**
+   * When a chip is destroyed, we store the index of the destroyed chip until the chips
+   * query list notifies about the update. This is necessary because we cannot determine an
+   * appropriate chip that should receive focus until the array of chips updated completely.
+   */
+  public lastDestroyedChipIndex: number | null = null;
+
+  /**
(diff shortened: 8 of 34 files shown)


.gitignore
```diff
@@ -1,3 +1,5 @@
+.vscode/
+coverage/
```
benjamincharity: `coverage/` is already listed

tooling/cz-config.js
```diff
@@ -21,6 +21,7 @@ module.exports = {
     {name: 'Card'},
     {name: 'Chart'},
     {name: 'Checkbox'},
+    {name: 'Chip'},
```
benjamincharity, line 24: ❤️

terminus-ui/chip/src/chip.component.ts
```diff
@@ -0,0 +1,419 @@
...
+   * is set to the global ripple options since we don't have any configurable options for
+   * the chip ripples.
+   */
+  public rippleConfig: RippleConfig & RippleGlobalOptions;
+
+  /**
+   * Whether ripples are disabled on interaction
+   */
+  public get rippleDisabled(): boolean {
+    return this.isDisabled || this.disableRipple || !!this.rippleConfig.disabled;
```
benjamincharity: I think we can remove `this.disableRipple` and the `rippleConfig` as a whole. There should be no reason disable the ripple unless the entire chip is disabled

terminus-ui/chip/src/chip.component.ts
```diff
@@ -0,0 +1,419 @@
...
+
+  constructor(public elementRef: ElementRef,
+              private ngZone: NgZone,
+              platform: Platform,
+              private changeDetectorRef: ChangeDetectorRef,
+              @Optional() @Inject(MAT_RIPPLE_GLOBAL_OPTIONS)
+              globalRippleOptions: RippleGlobalOptions | null) {
+    super(elementRef);
+
+    this.chipRipple = new RippleRenderer(this, ngZone, elementRef, platform);
```
benjamincharity: Don't need to change this since it's all in place, but for future reference you can use the ripple directive directly in the dom to avoid setting it up manually (unless there's a need for this that I haven't gotten to yet). Let's you not worry about subscriptions etc

terminus-ui/chip/src/chip.component.ts
```diff
@@ -0,0 +1,419 @@
...
+  }
+  public get id(): string {
+    return this._id;
+  }
+  protected _id: string = this.uid;
+
+
+  /** Whether the chip is selected. */
+  @Input()
+  public get selected(): boolean {
```
benjamincharity: Let's keep the format of setters first, then getters

terminus-ui/chip/src/chip.component.ts
```diff
@@ -0,0 +1,419 @@
...
+
+  public ngOnDestroy() {
+    this.destroyed.emit({chip: this});
+    // eslint-disable-next-line no-underscore-dangle
+    this.chipRipple._removeTriggerEvents();
+  }
+
+  /** Selects the chip. */
+  public select(): void {
+    if (!this._selected) {
```
benjamincharity: I know this is a material convention, but I prefer to not directly access the internal property when there is a getter. This helps to verify that everyone is getting the same thing (as some getters may perform some actions on the data prior to returning)

I'll call this out once here rather than commenting on all the places

terminus-ui/chip/src/chip.component.ts
```diff
@@ -0,0 +1,419 @@
...
+  public focus(): void {
+    if (!this.hasFocus) {
+      this.elementRef.nativeElement.focus();
+      this.onFocus.next({chip: this});
+    }
+    this.hasFocus = true;
+  }
+
+  /**
+   * Allows for programmatic removal of the chip. Called by the TsChipCollectionComponent when the DELETE or
```
benjamincharity, line 356: If we format component references a certain way, our docs generation will create a link:

`{@link TsChipCollectionComponent}`

terminus-ui/chip/src/chip.component.ts
```diff
@@ -0,0 +1,419 @@
...
+  public get isRemovable(): boolean {
+    return this._removable;
+  }
+  public set isRemovable(value: boolean) {
+    this._removable = coerceBooleanProperty(value);
+    this.changeDetectorRef.detectChanges();
+  }
+  protected _removable = true;
+
+  /** Emits when the chip is focused. */
```
benjamincharity: For consistencies sake, we should convert these inline comments to JSDoc format.

Commenting only in one place, but applies in multiple

terminus-ui/chip/src/chip.component.ts
```diff
@@ -0,0 +1,419 @@
...
+
+  /** Emitted when the chip is selected or deselected. */
+  @Output() public readonly selectionChange: EventEmitter<TsChipSelectionChange> =
+      new EventEmitter<TsChipSelectionChange>();
+
+  /** Emitted when the chip is destroyed. */
+  @Output() public readonly destroyed: EventEmitter<TsChipEvent> = new EventEmitter<TsChipEvent>();
+
+  /** Emitted when a chip is to be removed. */
+  @Output() public readonly removed: EventEmitter<TsChipEvent> = new EventEmitter<TsChipEvent>();
```
benjamincharity: This is definitely the pattern I started, but I've realized we are double typing our outputs.

I've started using this format now:

```
@Output() public readonly removed = new EventEmitter<TsChipEvent>();
```

As it's shorter and still retains the typings.

(again, multiple places, single comment)

terminus-ui/chip/src/chip.component.ts
```diff
@@ -0,0 +1,419 @@
...
+
+  /** The ARIA selected applied to the chip. */
+  public get ariaSelected(): string | null {
+    // Remove the `aria-selected` when the chip is deselected in single-selection mode, because
```
benjamincharity: We should prefix notes like these: `NOTE: My info...` (existing pattern)

terminus-ui/chip/src/chip.component.ts
```diff
@@ -0,0 +1,419 @@
...
+   *
+   * Informs any listeners of the removal request. Does not remove the chip from the DOM.
+   */
+  public remove(): void {
+    if (this.isRemovable) {
+      this.removed.emit({chip: this});
+    }
+  }
+
+  /** Handles click events on the chip. */
```
benjamincharity: Looks like many of these methods are missing JSDoc comments

terminus-ui/chip/src/chip.component.ts
```diff
@@ -0,0 +1,419 @@
...
+}
+
+const TsChipMixinBase: CanColorCtor & CanDisableRippleCtor & CanDisableCtor & typeof TsChipBase =
+    mixinColor(mixinDisableRipple(mixinDisabled(TsChipBase)), 'primary');
+
+
+/**
+ * Directive to add CSS class to chip trailing icon.
+ */
+@Directive({
```
benjamincharity, line 83: I know this is a Material pattern, but we should put all directives/components into their own files for clarity

terminus-ui/chip/src/chip.component.scss
```diff
@@ -0,0 +1,77 @@
...
+@import './../../scss/helpers/spacing';
+@import './../../scss/helpers/typography';
+
+$ts-chips-chip-margin: 4px;
+$ts-chip-height: 35px;
+$ts-chip-vertical-padding: 3px;
+$ts-chip-horizontal-padding: 8px;
+$ts-chip-remove-width: 18px;
+$ts-chip-remove-height: 18px;
+$ts-chip-background-color: #e0e0e0;
```
benjamincharity: Any chance our existing grays could work for this design?

terminus-ui/chip/src/chip.component.scss
```diff
@@ -0,0 +1,77 @@
...
+  height: $ts-chip-height;
+  margin: $ts-chips-chip-margin;
+  overflow: hidden;
+  padding: $ts-chip-vertical-padding $ts-chip-horizontal-padding;
+  position: relative;
+  -webkit-tap-highlight-color: transparent;
+  transform: translateZ(0);
+  transition: box-shadow 280ms cubic-bezier(.4, 0, .2, 1);
+
+  &:focus {
```
benjamincharity: We should add `:hover` to this style too

terminus-ui/chip/src/chip.component.scss
```diff
@@ -0,0 +1,77 @@
...
+
+.ts-icon {
+  position: absolute;
+  right: 5%;
+  top: 42%;
+  transform: translate(-50%, -50%);
+}
+
+.ts-chip-collection-wrapper {
+  align-items: center;
```
benjamincharity, line 70: Curious, any reason we doing flex here rather than using the directives?

terminus-ui/chip/src/chip.component.html
```diff
@@ -0,0 +1,17 @@
+<div
+  class="c-chip qa-chip"
```
benjamincharity: Not a required change and totally on me for not conveying this, but I've stopped adding QA classes to components since they are no longer used

terminus-ui/chip/src/chip.component.html
```diff
@@ -0,0 +1,17 @@
...
+    'c-chip--disabled': isDisabled,
+    'c-chip--removable': isRemovable
+  }"
+  [attr.id]="id"
+>
+  <span class="ts-chip-value">
+    {{ value }}
+  </span>
+  <ts-icon tsChipRemove *ngIf="isRemovable"
+  >
```
benjamincharity: Probably bump this back up to the previous line

terminus-ui/chip/src/chip-collection.component.html
```diff
@@ -0,0 +1,15 @@
+<div 
+  class="ts-chip-collection-wrapper"
+  [ngClass]="(orientation=='vertical') ? 'ts-chip-collection-stacked':''"
```
benjamincharity: We should be using strict equality here

terminus-ui/chip/src/chip.component.scss
```diff
@@ -0,0 +1,77 @@
...
+}
+
+.ts-icon {
+  position: absolute;
+  right: 5%;
+  top: 42%;
+  transform: translate(-50%, -50%);
+}
+
+.ts-chip-collection-wrapper {
```
benjamincharity, line 69: Seems like this should be in a different stylesheet?

terminus-ui/chip/src/chip.component.scss
```diff
@@ -0,0 +1,77 @@
...
+  align-items: center;
+  display: flex;
+  flex-flow: row wrap;
+  margin: -$ts-chips-chip-margin;
+  &.ts-chip-collection-stacked {
```
benjamincharity: We should probably use the bem style selector here: `ts-chip-collection--stacked`

terminus-ui/chip/src/chip-collection.component.ts
```diff
@@ -0,0 +1,569 @@
...
+ *              [isRemovable]="true"
+ *              [isSelectable]="false"
+ *              [orientation]="horizontal"
+ *              (collectionChange)="collectionChange($event)"
+ * ></ts-chip-collection>
+ *
+ * <example-url>https://getterminus.github.io/ui-demos-release/components/chip</example-url>
+ */
+@Component({
+  moduleId: module.id,
```
benjamincharity: This is for systemjs and not needed for us

terminus-ui/chip/src/chip-collection.component.ts
```diff
@@ -0,0 +1,569 @@
...
+    return this.chips && this.chips.length === 0;
+  }
+
+  /** The array of selected chips inside chip collection. */
+  public get selected(): TsChipComponent[] | TsChipComponent {
+    return this.allowMultipleSelections ? this.selectionModel.selected : this.selectionModel.selected[0];
+  }
+
+  // /** The ARIA role applied to the chip collection. */
+  // get role(): string | null { return this.empty ? null : 'collectionbox'; }
```
benjamincharity: ?

terminus-ui/chip/src/chip-collection.component.ts
```diff
@@ -0,0 +1,569 @@
...
+    return this._value;
+  }
+  public set value(value: string[]) {
+    this._value = value;
+  }
+
+  /**
+   * Get chip collection id
+   */
+  public get id(): string {
```
benjamincharity, line 268: We should allow consumers to set an ID to match our existing convention

terminus-ui/chip/src/chip-collection.component.ts
```diff
@@ -0,0 +1,569 @@
...
+  /** Whether the user should be allowed to select multiple chips. */
+  @Input()
+  public get allowMultipleSelections(): boolean {
+    return this._multiple;
+  }
+  public set allowMultipleSelections(value: boolean) {
+    this._multiple = coerceBooleanProperty(value);
+    this.syncChipsState();
+  }
+
```
benjamincharity: Extra line break

terminus-ui/chip/src/chip-collection.component.ts
```diff
@@ -0,0 +1,569 @@
...
+    '(blur)': 'blur()',
+    '(keydown)': 'keydown($event)',
+    '[id]': 'uid',
+  },
+  styleUrls: ['./chip.component.scss'],
+  encapsulation: ViewEncapsulation.None,
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class TsChipCollectionComponent implements AfterViewInit, OnInit, OnDestroy {
+
```
benjamincharity, line 103: Going to make a single comment here to cut down on the noise:

a) We should align the order to our other components:

1. private properties
1. public properties
1. getters
1. view references
1. inputs
1. outputs
1. constructor
1. public methods
1. protected/static methods
1. private methods

b) We should use JSDoc comments in all the places

c) Where applicable, private properties should be below the getter/setters

```
get foo
set foo
_foo
```

d) setters should be before getters (see last example)

e) decorators should be on their own line to match our existing conventions

```
@Input/@Output
public ...
```

f) we should only prefix fields with an underscore if they would conflict with a getter or setter name

terminus-ui/chip/src/chip-collection.component.ts
```diff
@@ -0,0 +1,569 @@
...
+  /** Function when changed */
+  public onChange: (value: string[]) => void = () => {};
+  private compareWith = (o1: string, o2: string) => o1 === o2;
+
+  constructor(protected elementRef: ElementRef<HTMLElement>,
+              private changeDetectorRef: ChangeDetectorRef,
+              public zone: NgZone,
+              @Optional() private dir: Directionality,) {
+    this._role = this.empty ? null : 'listbox';
+    zone.runOutsideAngular(() => {
```
benjamincharity, line 364: Curious if you know exactly why this needs to be set twice, with one time outside of zone? If you do happen to, it would be nice to add a comment here for future reference

terminus-ui/chip/src/chip-collection.component.ts
```diff
@@ -0,0 +1,569 @@
...
+        chip.chipCollectionSelectable = this._selectable;
+      });
+    }
+  }
+  protected _selectable = true;
+
+  @Input() public orientation: 'horizontal' | 'vertical' = 'horizontal';
+
+  @Input()
+  public set tabIndex(value: number) {
```
benjamincharity, line 327: We should add a getter here and update references to the private prop

terminus-ui/chip/src/chip-collection.component.ts
```diff
@@ -0,0 +1,569 @@
...
+   * query list notifies about the update. This is necessary because we cannot determine an
+   * appropriate chip that should receive focus until the array of chips updated completely.
+   */
+  public lastDestroyedChipIndex: number | null = null;
+
+  /** Subject that emits when the component has been destroyed. */
+  private destroyed = new Subject<void>();
+
+  /** Subscription to focus changes in the chips. */
+  private chipFocusSubscription!: Subscription | null;
```
benjamincharity: We should try to avoid using the `!` when possible (often it really is needed).

But here I think we can just set these all to `null` initially to avoid it.

terminus-ui/chip/src/chip-collection.component.ts
```diff
@@ -0,0 +1,569 @@
...
+  private chipBlurSubscription!: Subscription | null;
+
+  /** Subscription to selection changes in chips. */
+  private chipSelectionSubscription!: Subscription | null;
+
+  /** Subscription to remove changes in chips. */
+  private chipRemoveSubscription!: Subscription | null;
+
+  /** Uid of the chip collection */
+  public uid = `ts-chip-collection-${nextUniqueId++}`;
```
benjamincharity: This should be `protected`

terminus-ui/chip/src/chip-collection.component.ts
```diff
@@ -0,0 +1,569 @@
...
+    zone.runOutsideAngular(() => {
+      this._role = this.empty ? null : 'listbox';
+    });
+  }
+
+  public ngAfterViewInit() {
+    this.keyManager = new FocusKeyManager<TsChipComponent>(this.chips)
+      .withWrap()
+      .withVerticalOrientation()
+      .withHorizontalOrientation(this.dir ? this.dir.value : 'ltr');
```
benjamincharity, line 377: Ah didn't know we could init with both orientations 👍

terminus-ui/chip/src/chip-collection.component.ts
```diff
@@ -0,0 +1,569 @@
...
+        // checked, we need to defer it to the next tick.
+        Promise.resolve().then(() => {
+          this.syncChipsState();
+        });
+      }
+
+      this.resetChips();
+
+      // Check to see if we need to update our tab index
+      setTimeout(() => {
```
benjamincharity: Can we use `Promise.resolve` here? Feels slightly less hacky 😆

terminus-ui/chip/src/chip-collection.component.ts
```diff
@@ -0,0 +1,569 @@
...
+
+    });
+  }
+
+  public ngOnInit() {
+    this.selectionModel = new SelectionModel<TsChipComponent>(this.allowMultipleSelections, undefined, false);
+  }
+
+  public ngOnDestroy() {
+    this.destroyed.next();
```
benjamincharity: We can use the `untilComponentDestroyed` helper so we don't have to manually manage/test this

terminus-ui/chip/src/chip-collection.component.ts
```diff
@@ -0,0 +1,569 @@
...
+  }
+
+  private resetChips() {
+    this.dropSubscriptions();
+    this.listenToChipsFocus();
+    this.listenToChipsSelection();
+    this.listenToChipsRemoved();
+  }
+
+  private dropSubscriptions() {
```
benjamincharity: Same note here, we could use `untilComponentDestroyed` and avoid this entire method

terminus-ui/chip/src/chip-collection.component.ts
```diff
@@ -0,0 +1,569 @@
...
+    this._tabIndex = this._userTabIndex || (this.chips.length === 0 ? -1 : 0);
+  }
+
+  /**
+   * If the amount of chips changed, we need to update the
+   * key manager state and focus the next closest chip.
+   */
+  protected updateFocusForDestroyedChips() {
+    // Move focus to the closest chip. If no other chips remain, focus the chip-collection itself.
+    if (this.lastDestroyedChipIndex != null) {
```
benjamincharity: Do we explicitly need loose equality here?

terminus-ui/chip/src/chip-collection.component.ts
```diff
@@ -0,0 +1,569 @@
...
+      // the index in order to be able to determine an appropriate sibling chip that will
+      // receive focus.
+      if (this.isValidIndex(chipIndex) && chip.hasFocus) {
+        this.lastDestroyedChipIndex = chipIndex;
+      }
+    });
+  }
+
+  /** Checks whether any of the chips is focused. */
+  private hasFocusedChip() {
```
benjamincharity: This can be converted into a getter

terminus-ui/chip/src/chip-collection.component.ts
```diff
@@ -0,0 +1,569 @@
...
+    return this.chips.some(chip => chip.hasFocus);
+  }
+
+  /** Syncs the collection's state with the individual chips. */
+  private syncChipsState() {
+    if (this.chips) {
+      this.chips.forEach(chip => {
+        chip.isDisabled = this._disabled;
```
benjamincharity: we should reference the getter

terminus-ui/chip/src/chip-collection.component.ts
```diff
@@ -0,0 +1,569 @@
...
+    }
+  }
+
+  /**
+   * Function for tracking for-loops changes
+   *
+   * @param index - The item index
+   * @return The index
+   */
+  public trackByFn(index): number {
```
benjamincharity, line 637: I really wish angular just did this be default and we only override it if we need a better check

terminus-ui/chip/src/chip.component.scss
```diff
@@ -0,0 +1,77 @@
...
+$ts-chip-background-color: #e0e0e0;
+$ts-chip-focus-background-color: #cfcccc;
+$ts-chip-color: rgba(0, 0, 0, .87);
+$c-chip-top-margin: 7px;
+$c-chip-right-margin: 4px;
+$c-chip-padding: 25px;
+
+.ts-chip {
+  @include reset;
+  @include typography;
```
benjamincharity, line 22: We need to update our type styles to include a `14px` size, but until we do, let's just hardcode the style here with a note. That way the styles won't change in the UI when we release this and we don't have to make it breaking.

terminus-ui/chip/src/chip.component.scss
```diff
@@ -0,0 +1,77 @@
...
+  &:focus {
+    background-color: $ts-chip-focus-background-color;
+    transition: opacity 200ms cubic-bezier(.35, 0, .25, 1);
+  }
+
+  &.ts-chip-selected {
+    background-color: color(primary);
+  }
+
+  .ts-chip-remove {
```
benjamincharity, line 49: Once the font size is changed, we'll need to make this about 80% of it's current size. (this is on me for not really calling out styles in the ticket - I was too focused on functionality)

terminus-ui/autocomplete/src/autocomplete.component.html
```diff
@@ -15,38 +15,26 @@
...
-          <span class="ts-autocomplete-chip-value">
-            {{ option }}
-          </span>
-
-          <ts-icon matChipRemove>
-            cancel
-          </ts-icon>
-        </mat-chip>
-
+      <ts-chip-collection
```
benjamincharity, line 18: 🎉

terminus-ui/autocomplete/src/autocomplete.component.ts
```diff
@@ -236,8 +238,8 @@ export class TsAutocompleteComponent implements OnInit,
   /**
    * Access the chip list
    */
-  @ViewChild('chipList', {static: false})
-  public chipList: MatChipList | undefined;
+  @ViewChild('chipCollection', {static: false})
```
benjamincharity: We should be able to remove any matchip/matchiplist imports too

terminus-ui/autocomplete/src/autocomplete.component.ts
```diff
@@ -54,6 +54,8 @@ import {
   switchMap,
 } from 'rxjs/operators';
 
+import { KEYS } from '@terminus/ngx-tools/keycodes';
```
benjamincharity: Is this used anywhere?

terminus-ui/autocomplete/src/autocomplete.module.ts
```diff
@@ -39,6 +40,7 @@ export * from './autocomplete-panel/autocomplete-panel.component';
     MatRippleModule,
     OverlayModule,
     TsCheckboxModule,
+    TsChipModule,
```
benjamincharity, line 41: I think we can remove the mat chips module now too

demo/app/components/chip/chip.component.ts
```diff
@@ -0,0 +1,49 @@
...
+@Component({
+  selector: 'demo-chip',
+  templateUrl: './chip.component.html',
+})
+export class ChipComponent implements OnInit, AfterViewInit {
+
+  @ViewChild(TsChipCollectionComponent, { static: false, })
+  public list!: TsChipCollectionComponent;
+
+  @ViewChild('tsChipCollection', { static: true })
```
benjamincharity: Are both of these view queries needed?

terminus-ui/chip/src/chip-collection.component.ts
```diff
@@ -0,0 +1,569 @@
...
+  /** The chip components contained within this chip collection. */
+  @ViewChildren(TsChipComponent) public chips!: QueryList<TsChipComponent>;
+
+  /** Function when touched */
+  public onTouched = () => {};
+  public _role: string | null = null;
+
+  /** Function when changed */
+  public onChange: (value: string[]) => void = () => {};
+  private compareWith = (o1: string, o2: string) => o1 === o2;
```
benjamincharity: Is this used anywhere?

terminus-ui/chip/src/chip-collection.component.ts
```diff
@@ -0,0 +1,569 @@
...
+    return this.hasFocusedChip();
+  }
+
+  /** Determine whether there is at least one chip in collection */
+  public get empty(): boolean {
+    return this.chips && this.chips.length === 0;
+  }
+
+  /** The array of selected chips inside chip collection. */
+  public get selected(): TsChipComponent[] | TsChipComponent {
```
benjamincharity: Is this used anywhere?

terminus-ui/chip/src/chip.component.scss
```diff
@@ -0,0 +1,77 @@
...
+  transform: translateZ(0);
+  transition: box-shadow 280ms cubic-bezier(.4, 0, .2, 1);
+
+  &:focus {
+    background-color: $ts-chip-focus-background-color;
+    transition: opacity 200ms cubic-bezier(.35, 0, .25, 1);
+  }
+
+  &.ts-chip-selected {
+    background-color: color(primary);
```
benjamincharity, line 45: We should set the color to our pure light color. It's tough to read the black on this dark blue

terminus-ui/chip/src/chip.component.scss
benjamincharity: We should darken the remove icon background on hover to signal interaction

terminus-ui/chip/src/chip.component.scss
```diff
@@ -0,0 +1,77 @@
...
+  margin: -$ts-chips-chip-margin;
+  &.ts-chip-collection-stacked {
+    align-items: flex-start;
+    flex-direction: column;
+  }
+}
+
+.c-chip {
+  margin: $c-chip-top-margin $c-chip-right-margin;
+  padding-right: $c-chip-padding;
```
benjamincharity: We should not have this padding if removal is disabled. Can probably use a css :not selector.

Screen Shot 2019-08-28 at 11 03 48 AM

terminus-ui/chip/src/chip.component.spec.ts
```diff
@@ -0,0 +1,282 @@
...
+      testComponent = fixture.debugElement.componentInstance;
+    });
+
+    describe(`basic behaviors`, () => {
+
+      test('should add the `ts-chip` class', () => {
+        expect(chipNativeElement.classList).toContain('ts-chip');
+      });
+
+      test('should not add the `ts-basic-chip` class', () => {
```
benjamincharity: Where would this class ever be set?

terminus-ui/chip/src/chip.component.spec.ts
```diff
@@ -0,0 +1,282 @@
...
+        expect(chipInstance.id).toBeTruthy();
+
+        chipInstance.id = 'foo';
+        expect(chipInstance.id).toEqual('foo');
+
+        chipInstance.id = null as any;
+        expect(chipInstance.id).toEqual(expect.stringContaining('ts-chip-'));
+      })
+
+      test(`should allow removal`, () => {
```
benjamincharity, line 97: We should probably test that the removal icon is visible here

terminus-ui/chip/src/chip.component.spec.ts
```diff
@@ -0,0 +1,282 @@
...
+
+      test(`should toggle selected`, () => {
+        chipInstance.select();
+        expect(chipInstance.toggleSelected()).toBeFalsy();
+
+        chipInstance.deselect();
+        expect(chipInstance.toggleSelected()).toBeTruthy();
+      });
+
+      test(`should be able to disable ripples through ripple global options at runtime`, () => {
```
benjamincharity: Won't be needed once we remove global options

terminus-ui/chip/src/chip.component.spec.ts
```diff
@@ -0,0 +1,282 @@
...
+    const providers = [
+      {
+        provide: MAT_RIPPLE_GLOBAL_OPTIONS,
+        useFactory: () => globalRippleOptions,
+      },
+    ]
+    return createComponentInner(component, providers, moduleImports);
+  }
+
+  describe('TsChip', () => {
```
benjamincharity, line 48: I don't think we're testing the `destroyed` emitter anywhere

terminus-ui/chip/src/chip-collection.component.spec.ts
```diff
@@ -0,0 +1,422 @@
...
+  test(`should not allow keyboard focus when there's 0 chip`, fakeAsync(() => {
+    fixture = createComponent(testComponents.NoChip);
+    chipCollectionInstance = getChipCollectionInstance(fixture);
+    fixture.detectChanges();
+    tick(1000);
+    expect(chipCollectionInstance._tabIndex).toBe(-1);
+  }));
+
+
+  test('should not have the aria-selected attribute when is not selectable', () => {
```
benjamincharity, line 236: We should be testing all the aria labels (role/required/disabled/orientation/multiselectable/activedescendant/readonly)

terminus-ui/chip/src/chip-collection.component.spec.ts
```diff
@@ -0,0 +1,422 @@
...
+  }));
+
+  test(`should set value properly`, function() {
+    fixture = createComponent(testComponents.NoValueChip);
+    fixture.detectChanges();
+    const instance = getChipCollectionInstance(fixture);
+
+    expect(instance.value).toEqual([]);
+    instance.value = ['banana'];
+    expect(instance.value).toEqual(['banana']);
```
benjamincharity, line 132: Should we check the dom here to make sure it _really_ worked?

terminus-ui/chip/src/chip-collection.component.spec.ts
```diff
@@ -0,0 +1,422 @@
...
+    TsChipModule,
+    TsAutocompleteModule,
+    TsOptionModule,
+    NoopAnimationsModule,
+  ];
+
+  return createComponentInner(component, undefined, moduleImports);
+}
+
+describe(`TsChipCollection`, function() {
```
benjamincharity, line 50: Are we testing the mouse interaction anywhere? (shift+click)

terminus-ui/chip/src/chip-collection.component.spec.ts
```diff
@@ -0,0 +1,422 @@
...
+      expect(manager.activeItemIndex).toBe(-1);
+
+      // Press the BACKSPACE key
+      chipCollectionInstance.keydown(BACKSPACE_EVENT);
+      fixture.detectChanges();
+
+      // It focuses the last chip
+      expect(manager.activeItemIndex).toEqual(chips.length - 1);
+      chipCollectionInstance.keydown(BACKSPACE_EVENT_CHIP);
+      const nodeName = document.activeElement ? document.activeElement.nodeName : '';
```
benjamincharity: What's the purpose of the ternary here? afaik activeElement would be `null` or `body` so the check would still work.

Also, I'm wondering if we should be getting the nodeName _after_ calling detectChanges? Seems we might be getting a reference before the component settles.

terminus-ui/chip/src/chip-collection.component.spec.ts
```diff
@@ -0,0 +1,422 @@
...
+    }));
+  });
+
+  describe(`updateFocus`, () => {
+
+    beforeEach(() => {
+      setupStandardCollection('StandardChipCollection');
+    });
+
+    test(`should focus on last chip when END key pressed`, () => {
```
benjamincharity, line 383: I think this is a duplicate test from line 327

terminus-ui/chip/testing/src/test-helpers.ts
```diff
@@ -0,0 +1,178 @@
...
+  return debugElements.map(debugElement => debugElement.componentInstance);
+}
+
+/**
+ * Get chip collection instance when lives inside an autocomplete component
+ *
+ * @param fixture - The component fixture
+ * @return a chip collection instance
+ */
+
```
benjamincharity: Extra line break

terminus-ui/chip/testing/src/test-helpers.ts
```diff
@@ -0,0 +1,178 @@
...
+  const contentElement = debugElement.query(By.css('.ts-chip-value'));
+  return contentElement.nativeElement;
+}
+
+/**
+ * Get all the elements of a chip collection
+ * @param fixture - The component fixture
+ * @return an Array of chip elements
+ */
+export function getAllChipElements(fixture: ComponentFixture<any>): DebugElement[] {
```
benjamincharity: We should probably rename this to be more specific about what it returns. From the name I'd assume it was returning `HTMLElement[]`

terminus-ui/chip/src/chip-collection.component.ts
```diff
@@ -0,0 +1,569 @@
...
+    return this.allowMultipleSelections ? this.selectionModel.selected : this.selectionModel.selected[0];
+  }
+
+  // /** The ARIA role applied to the chip collection. */
+  // get role(): string | null { return this.empty ? null : 'collectionbox'; }
+
+  @Input()
+  public chipsInput: string[] = [];
```
benjamincharity: We should define all `@Input` arrays as `ReadonlyArray<string>` to ensure no mutation

terminus-ui/chip/src/chip-collection.component.ts
```diff
@@ -0,0 +1,569 @@
...
+    this._multiple = coerceBooleanProperty(value);
+    this.syncChipsState();
+  }
+
+
+  /**
+   * Get chip collection value
+   */
+  @Input()
+  public get value(): string[] {
```
benjamincharity: We should define all `@Input` arrays as `ReadonlyArray<string>` to ensure no mutation

benjamincharity (changes requested):
So pumped to get this in!! 🚀 	

giphy

As always with big PRs it seems like a lot of comments, but it's 90% style stuff.

Only have a few callouts I couldn't comment on directly:

1. This was a miss on my requirements, but if a user is focused on the chip collection within the autocomplete and hit `tab`, we should focus the input.
1. The primary readme component table should be updated
1. The component count badge should be updated (i'm exploring ways to automate this)

terminus-ui/chip/src/chip-collection.component.spec.ts
atlwendy: Do we need to explicitly test all of them? Since the coverage doesn't show that lacking, I had assumed it was covered.

terminus-ui/chip/src/chip-collection.component.spec.ts
atlwendy: ts wouldn't let me call `.nodeName` without the ternary. because it's possible to be `null`?
Noted on detectChanges,.... it worked as expected so I thought it got updated properly.

terminus-ui/chip/src/chip-collection.component.ts
atlwendy: Can we add this to the repo DEVELOPMENT markdown?

terminus-ui/chip/src/chip-collection.component.ts
atlwendy: it was, but logic probably has changed during the revolution 😂

terminus-ui/chip/src/chip-collection.component.ts
atlwendy: there was a duplication and forgot to delete this one..

terminus-ui/chip/src/chip-collection.component.ts
atlwendy: With the setter, can it be readonly?

terminus-ui/chip/src/chip-collection.component.ts
atlwendy: what'd be advantage of converting to getter?

terminus-ui/chip/src/chip.component.html
atlwendy: was wondering about this but went with "convention" lol

terminus-ui/chip/src/chip.component.spec.ts
atlwendy: It's covered by chip collection specs

terminus-ui/chip/src/chip.component.spec.ts
atlwendy: whoops i initially had it in, but removed later on...

terminus-ui/chip/src/chip.component.ts
atlwendy: Funny that I used to do what you say here, but started to add typing because I saw somewhere having it and thought it's convention now to add it lol
now undo what i did...

terminus-ui/chip/src/chip-collection.component.ts
benjamincharity: Yep! I'll open a ticket for myself. 

I bet it wouldn't be hard for us to write a custom lint rule that fixed this for us.. 🤔

terminus-ui/chip/src/chip-collection.component.ts
benjamincharity: (btw, this is only a suggestion, not a blocker for the PR)

IMO it makes the intention more clear. When I call a function, there is a 50/50 chance I'm triggering functionality. But when I interact with a getter I'm just requesting a piece of data. (of course, people can abuse getters to do more than return data) I've noticed many of these types of functions are even  named like a getter eg `getActiveChip` etc. 

Again, might be personal preference, but 
`myChip = foo.activeChip` 
seems slightly more concise than 
`myChip = foo.getActiveChip()`

Also I find it clear because a function may or may not return anything, but a getter must return something.

terminus-ui/chip/src/chip-collection.component.spec.ts
benjamincharity: That's interesting.. If we're not using snapshots and we aren't checking the values anywhere I'm not sure why it _would_ be covered?

We do need to ensure that functionality, but if you can see that we are already doing that somewhere, then we don't need to duplicate here.

terminus-ui/chip/src/chip-collection.component.ts
benjamincharity: I believe so. Because we should be cloning it internally if we need to mutate it at all.

terminus-ui/chip/src/chip-collection.component.ts
atlwendy: That'd be awesome! From time to time I found myself having some stuff there just to test out whether the idea works or not, but once it worked, usually forgot to move to more appropriate place..

terminus-ui/chip/src/chip-collection.component.ts
atlwendy: I was running into some weirdness when I first tested out the solution, so setting that outside the block kinda helped. But now I don't see the weirdness any more, so I'm removing it.

terminus-ui/chip/src/chip-collection.component.ts
atlwendy: Also, if there's life cycle hooks defined, they probably want to live right after constructor?

benjamincharity: Ah yes, good catch. I'll update the issue

terminus-ui/chip/src/chip.component.scss
atlwendy: Err... didn't think about that much... flex was probably just easier to get `stacked` working properly?

terminus-ui/chip/src/chip.component.spec.ts
atlwendy: now covered by the test above this.

terminus-ui/chip/src/chip-collection.component.spec.ts
atlwendy: Hmm... not really because those interactions are not explicitly written in the component? should those be covered?

terminus-ui/chip/src/chip.component.scss
benjamincharity: Meant the `fxLayout` directives (so still using flexbox - just doing via directive vs CSS)

terminus-ui/chip/src/chip-collection.component.spec.ts
benjamincharity: I'm not sure what you mean? It was a requirement for this component that `shift+click` should add/remove a chip from the selection group. So we should probably be verifying that it works.

demo/app/components/chip/chip.component.ts
```diff
@@ -0,0 +1,36 @@
...
+    // If not found
+    if (index < 0) {
+      return;
+    }
+    selections.splice(index, 1);
+    this.options = selections;
+  }
+
+  change(value: string | string[]) {
+    console.log('collection change triggered value: ', value);
```
benjamincharity: Let's prefix this message like we do with our other demo logs: `DEMO: my message`

terminus-ui/chip/src/chip-collection.component.spec.ts
benjamincharity: ?

terminus-ui/chip/src/chip-collection.component.ts
```diff
@@ -0,0 +1,588 @@
...
+  TsChipComponent,
+  TsChipEvent,
+  TsChipSelectionChange,
+} from './chip.component';
+
+
+// Increasing integer for generating unique ids for chip-collection components.
+let nextUniqueId = 0;
+
+/** Change event object that is emitted when the chip collection value has changed. */
```
benjamincharity: Incorrect JSDOC comment format.

terminus-ui/chip/src/chip-collection.component.ts
```diff
@@ -0,0 +1,588 @@
...
+export class TsChipCollectionChange {
+  constructor(
+    /** Chip collection that emitted the event. */
```
benjamincharity: We usually use inline comments inside an area like this (See: TsSelect)

terminus-ui/chip/src/chip-collection.component.ts
```diff
@@ -0,0 +1,588 @@
...
+  /**
+   * The FocusKeyManager which handles focus.
+   */
+  public keyManager!: FocusKeyManager<TsChipComponent>;
+
+  /**
+   * Function when touched
+   */
+  public onTouched = () => { };
+  public _role: string | null = null;
```
benjamincharity: Missing comment

terminus-ui/chip/src/chip-collection.component.ts
```diff
@@ -0,0 +1,588 @@
...
+
+  /**
+   * Function when changed
+   */
+  public onChange: (value: string[]) => void = () => { };
+
+  public selectionModel!: SelectionModel<TsChipComponent>;
```
benjamincharity, line 170: Missing comment

terminus-ui/chip/src/chip-collection.component.ts
```diff
@@ -0,0 +1,588 @@
...
+  }
+
+  /**
+   * Combined stream of all of the child chips' remove change events.
+   */
+  public get chipRemoveChanges(): Observable<TsChipEvent> {
+    return merge(...this.chips.map(chip => chip.destroyed));
+  }
+
+  /** Determine whether there is at least one chip in collection */
```
benjamincharity: JSDoc comment format

terminus-ui/chip/src/chip-collection.component.ts
```diff
@@ -0,0 +1,588 @@
...
+  public get empty(): boolean {
+    return this.chips && this.chips.length === 0;
+  }
+
+  /** Whether any chips has focus. */
```
benjamincharity: JSDoc comment format

terminus-ui/chip/src/chip-collection.component.ts
```diff
@@ -0,0 +1,588 @@
...
+  public get focused(): boolean {
+    return this.chips.some(chip => chip.hasFocus);
+  }
+
+  /** The ARIA role applied to the chip list. */
```
benjamincharity: JSDoc comment format

terminus-ui/chip/src/chip-collection.component.ts
```diff
@@ -0,0 +1,588 @@
...
+  public get role(): string | null {
+    return this._role;
+  }
+
+  /**
+   * The chip components contained within this chip collection.
+   */
+  @ViewChildren(TsChipComponent) public chips!: QueryList<TsChipComponent>;
+
+  @Input()
```
benjamincharity, line 230: Missing comment

terminus-ui/chip/src/chip-collection.component.ts
```diff
@@ -0,0 +1,588 @@
...
+  public chipsInput: ReadonlyArray<string> = [];
+
+  /** Whether the user should be allowed to select multiple chips. */
```
benjamincharity: JSDoc comment format

terminus-ui/chip/src/chip-collection.component.ts
```diff
@@ -0,0 +1,588 @@
...
+    this._multiple = coerceBooleanProperty(value);
+    this.syncChipsState();
+  }
+  public get allowMultipleSelections(): boolean {
+    return this._multiple;
+  }
+  private _multiple = false;
+
+  /**
+   * Get chip collection value
```
benjamincharity: Should this be set rather than get?

terminus-ui/chip/src/chip-collection.component.ts
```diff
@@ -0,0 +1,588 @@
...
+  public set value(value: string[]) {
+    this._value = value;
+  }
+  public get value(): string[] {
+    return this._value;
+  }
+  protected _value = [''];
+
+  /**
+   * Get chip collection id
```
benjamincharity: Should this be set rather than get?

terminus-ui/chip/src/chip-collection.component.ts
```diff
@@ -0,0 +1,588 @@
...
+        chip.chipCollectionSelectable = this._selectable;
+      });
+    }
+  }
+  public get isSelectable(): boolean {
+    return this._selectable;
+  }
+  protected _selectable = true;
+
+  @Input()
```
benjamincharity, line 320: Missing comment

terminus-ui/chip/src/chip-collection.component.ts
```diff
@@ -0,0 +1,588 @@
...
+  public orientation: 'horizontal' | 'vertical' = 'horizontal';
+
+  @Input()
```
benjamincharity, line 326: Missing comment

terminus-ui/chip/src/chip-collection.component.ts
```diff
@@ -0,0 +1,588 @@
...
+
+  /**
+   * The aria-describedby attribute on the chip collection for improved a11y.
+   */
+  public ariaDescribedby!: string;
+
+  /**
+   * Tab index for the chip collection.
+   */
+  public _tabIndex = 0;
```
benjamincharity: This should be down with the getter/setter

terminus-ui/chip/src/chip-collection.component.ts
```diff
@@ -0,0 +1,588 @@
...
+    return this._tabIndex;
+  }
+
+  /**
+   * Orientation of the chip collection.
+   */
+  @Input('aria-orientation')
+  public ariaOrientation: 'horizontal' | 'vertical' = 'horizontal';
+
+  /** Event emitted when the chip collection value has been changed by the user. */
```
benjamincharity: JSDoc comment format

terminus-ui/chip/src/chip-collection.component.ts
```diff
@@ -0,0 +1,588 @@
...
+      });
+
+      // Check to see if we have a destroyed chip and need to refocus
+      this.updateFocusForDestroyedChips();
+      this.propagateChanges();
+
+    });
+  }
+
+  public ngOnDestroy() { }
```
benjamincharity, line 408: We should add a comment that this empty function is needed for `untilComponentDestroyed`

terminus-ui/chip/src/chip-collection.component.spec.ts
atlwendy: Ah, I misread that requirement as switching focus (as the same as material does)... will add this on.

terminus-ui/chip/src/chip-collection.component.ts
```diff
@@ -0,0 +1,588 @@
...
+  /**
+   * When blurred, mark the field as touched when focus moved outside the chip collection.
+   */
+  public blur() {
+    if (!this.focused) {
+      this.keyManager.setActiveItem(-1);
+    }
+  }
+
+  public remove(event) {
```
benjamincharity, line 424: Missing comment

terminus-ui/chip/src/chip-collection.component.ts
```diff
@@ -0,0 +1,588 @@
...
+    }
+
+    if (this.chips.length > 0) {
+      this.keyManager.setFirstItemActive();
+    }
+  }
+
+  /**
+   * Utility to for whether input field is empty
+   * @param element
```
benjamincharity: Missing line break before `@param` and has extra line break between comment and function

terminus-ui/chip/src/chip-collection.component.ts
```diff
@@ -0,0 +1,588 @@
...
+   * @return boolean
+   */
+
+  private isInputEmpty(element: HTMLElement): boolean {
```
benjamincharity, line 448: Missing comment

terminus-ui/chip/src/chip-collection.component.ts
```diff
@@ -0,0 +1,588 @@
...
+    if (element && element.nodeName.toLowerCase() === 'input') {
+      const input = element as HTMLInputElement;
+      return !input.value;
+    }
+
+    return false;
+  }
+
+  /**
+   * Pass events to the keyboard manager.
```
benjamincharity, line 458: Missing param

terminus-ui/chip/src/chip-collection.component.ts
```diff
@@ -0,0 +1,588 @@
...
+   */
+  public keydown(event: KeyboardEvent) {
```
benjamincharity: Missing return type

terminus-ui/chip/src/chip-collection.component.ts
```diff
@@ -0,0 +1,588 @@
...
+    this.onChange(valueToEmit);
+    this.changeDetectorRef.markForCheck();
+  }
+
+  /**
+   * Check the tab index as you should not be allowed to focus an empty list.
+   */
+  protected updateTabIndex(): void {
+    // If we have 0 chips, we should not allow keyboard focus
+    this._tabIndex = this._userTabIndex || (this.chips.length === 0 ? -1 : 0);
```
benjamincharity: Any reason we are using the private property directly rather than the getter?

terminus-ui/chip/src/chip-collection.component.ts
```diff
@@ -0,0 +1,588 @@
...
+  }
+
+  /**
+   * If the amount of chips changed, we need to update the
```
benjamincharity: Comment wrapped really small for some reason

terminus-ui/chip/src/chip-collection.component.ts
```diff
@@ -0,0 +1,588 @@
...
+   * key manager state and focus the next closest chip.
+   */
+  protected updateFocusForDestroyedChips() {
```
benjamincharity: Missing return type

terminus-ui/chip/src/chip-collection.component.ts
```diff
@@ -0,0 +1,588 @@
...
+   *
+   * @param index The index to be checked.
+   * @returns True if the index is valid for our collection of chips.
+   */
+  private isValidIndex(index: number): boolean {
+    return index >= 0 && index < this.chips.length;
+  }
+
+
+  private resetChips() {
```
benjamincharity: Missing JSDOC comment.
Missing return type.

terminus-ui/chip/src/chip-collection.component.ts
```diff
@@ -0,0 +1,588 @@
...
+        this.keyManager.updateActiveItem(chipIndex);
+      }
+    });
+
+    this.chipBlurSubscription = this.chipBlurChanges.subscribe(() => {
+      this.blur();
+    });
+  }
+
+  private listenToChipsRemoved(): void {
```
benjamincharity, line 604: Missing JSDOC comment.

terminus-ui/chip/src/chip.component.scss
benjamincharity: Was this handled?

terminus-ui/chip/src/chip.component.scss
benjamincharity: Was this handled?

terminus-ui/chip/src/chip.component.scss
benjamincharity: Was this handled?

terminus-ui/chip/src/chip.component.scss
benjamincharity: ?

terminus-ui/chip/testing/src/test-helpers.ts
```diff
@@ -0,0 +1,175 @@
...
+  const debugElement = getChipDebugElement(fixture, index);
+  const contentElement = debugElement.query(By.css('.ts-chip-value'));
+  return contentElement.nativeElement;
+}
+
+/**
+ * Get all the elements of a chip collection
+ * @param fixture - The component fixture
+ * @return an Array of chip elements
+ */
```
benjamincharity: I believe this is legacy and can be removed

terminus-ui/autocomplete/src/autocomplete.component.ts
```diff
@@ -836,6 +836,9 @@ export class TsAutocompleteComponent implements OnInit,
     this.selectionChange.emit(new TsAutocompleteChange(this, this.autocompleteSelections));
   }
 
+  public focusInput(): void {
```
benjamincharity, line 853: Missing JSDOC comment.

terminus-ui/chip/src/chip.component.ts
```diff
@@ -0,0 +1,432 @@
...
+ * Unique ID for each instance
+ */
+let nextUniqueId = 0;
+
+
+/**
+ * Represents an event fired on an individual `ts-chip`.
+ */
+export interface TsChipEvent {
+  /** The chip the event was fired on. */
```
benjamincharity: Inline comment format

terminus-ui/chip/src/chip.component.ts
```diff
@@ -0,0 +1,432 @@
...
+  chip: TsChipComponent;
+}
+
+/**
+ * Event object emitted by TsChip when selected or deselected.
+ */
+export class TsChipSelectionChange {
+  constructor(
+    /** Reference to the chip that emitted the event. */
```
benjamincharity, line 60: Inline comment format

terminus-ui/chip/src/chip.component.ts
```diff
@@ -0,0 +1,432 @@
...
+    public source: TsChipComponent,
+    /** Whether the chip that emitted the event is selected. */
+    public selected: boolean,
+  ) { }
+}
+
+// Boilerplate for applying mixins to TsChip.
```
benjamincharity: Incorrect JSDOC comment format.

terminus-ui/chip/src/chip.component.ts
```diff
@@ -0,0 +1,432 @@
...
+ *              [isSelectable]="false"
+ *              [selected]="true"
+ *              (destroyed)="destroyed($event)"
+ *              (removed)="removed($event)"
+ *              (selectionChange)="selectionChange($event)"
+ * ></ts-chip>
+ *
+ * <example-url>https://getterminus.github.io/ui-demos-release/components/chip</example-url>
+ */
+
```
benjamincharity: Extra line break

terminus-ui/chip/src/chip.component.ts
```diff
@@ -0,0 +1,432 @@
...
+    '(blur)': 'blur()',
+  },
+})
+export class TsChipComponent extends TsChipMixinBase implements
+  FocusableOption,
+  OnDestroy,
+  CanColor,
+  CanDisable {
+
+  protected uid = `ts-chip-${nextUniqueId++}`;
```
benjamincharity, line 144: Missing JSDOC comment.

terminus-ui/chip/src/chip.component.ts
```diff
@@ -0,0 +1,432 @@
...
+    // NOTE: Remove the `aria-selected` when the chip is deselected in single-selection mode, because
+    // it adds noise to NVDA users where "not selected" will be read out for each chip.
+    return this.isSelectable && (this.chipCollectionMultiple || this.selected)
+      ? this.selected.toString() : null;
+  }
+
+  /**
+   * The chip's trailing icon.
+   */
+  @ContentChild(TsChipTrailingIconDirective, {static: false}) public trailingIcon!: TsChipTrailingIconDirective;
```
benjamincharity: Let's move the property below the decorator

terminus-ui/chip/src/chip.component.ts
```diff
@@ -0,0 +1,432 @@
...
+
+  /**
+   * The chip's remove toggler.
+   */
+  @ContentChild(forwardRef(() => TsChipRemoveDirective), {static: false}) public removeIcon!: TsChipRemoveDirective;
```
benjamincharity: Let's move the property below the decorator

terminus-ui/chip/src/chip.component.ts
```diff
@@ -0,0 +1,432 @@
...
+
+  @Input()
```
benjamincharity, line 197: Missing JSDOC comment.

terminus-ui/chip/src/chip.component.ts
```diff
@@ -0,0 +1,432 @@
...
+    return this._value === undefined
+      ? this.elementRef.nativeElement.textContent : this._value;
+  }
+  public set value(value: string) {
+    this._value = value;
+  }
+  protected _value = '';
+
+  /**
+   * Whether or not the chip is selectable. By default a chip is
```
benjamincharity: Comment seems to be wrapped very short

terminus-ui/chip/src/chip.component.ts
```diff
@@ -0,0 +1,432 @@
...
+  @Output()
+  public readonly destroyed = new EventEmitter<TsChipEvent>();
+
+  /**
+   * Emitted when a chip is to be removed.
+   */
+  @Output()
+  public readonly removed = new EventEmitter<TsChipEvent>();
+
+  constructor(public elementRef: ElementRef,
```
benjamincharity: Let's adjust this indention format to match our other constructors

terminus-ui/chip/src/chip.component.ts
```diff
@@ -0,0 +1,432 @@
...
+              private ngZone: NgZone,
+              private changeDetectorRef: ChangeDetectorRef) {
+    super(elementRef);
+
+  }
+
+  public ngOnDestroy() {
+    this.destroyed.emit({chip: this});
+    // eslint-disable-next-line no-underscore-dangle
```
benjamincharity: Legacy comment

terminus-ui/chip/src/chip.component.ts
```diff
@@ -0,0 +1,432 @@
...
+   * Informs any listeners of the removal request. Does not remove the chip from the DOM.
+   */
+  public remove(): void {
+    if (this.isRemovable) {
+      this.removed.emit({chip: this});
+    }
+  }
+
+  /**
+   * Handles click events on the chip.
```
benjamincharity, line 369: Need line break between title and param so our docs generation correctly lays it out

terminus-ui/chip/src/chip.component.ts
```diff
@@ -0,0 +1,432 @@
...
+   * @param event - click event
+   */
+  public handleClick(event: Event) {
```
benjamincharity: Missing return type.

terminus-ui/chip/src/chip.component.ts
```diff
@@ -0,0 +1,432 @@
...
+    if (this.isDisabled) {
+      event.preventDefault();
+    } else {
+      event.stopPropagation();
+    }
+  }
+
+  /**
+   * Handle custom key presses.
+   * @param event - keyboard event
```
benjamincharity, line 384: Need line break between title and param so our docs generation correctly lays it out

terminus-ui/chip/src/chip.component.ts
```diff
@@ -0,0 +1,432 @@
...
+        // Always prevent space from scrolling the page since the list has focus
+        event.preventDefault();
+        break;
+      default:
+        event.preventDefault();
+        break;
+    }
+  }
+
+  public blur(): void {
```
benjamincharity, line 417: Missing JSDOC comment.

terminus-ui/chip/src/chip.component.ts
```diff
@@ -0,0 +1,432 @@
...
+      .asObservable()
+      .pipe(take(1))
+      .subscribe(() => {
+        this.ngZone.run(() => {
+          this.hasFocus = false;
+        });
+      });
+  }
+
+  private dispatchSelectionChange() {
```
benjamincharity, line 432: Missing JSDOC comment.

terminus-ui/chip/src/chip.component.ts
```diff
@@ -0,0 +1,432 @@
...
+    this.selectionChange.emit(new TsChipSelectionChange(this, this._selected));
+  }
+}
+
+/**
+ * Applies proper (click) support and adds styling for use with "cancel" icon
+ *
+ * Example:
+ *
+ *     `<ts-chip>
```
benjamincharity: The first and last lines don't need to be indented, but all lines between need 14 space indention for our docs tool to indent properly:

Screen Shot 2019-09-06 at 11 46 34 AM

terminus-ui/chip/src/chip.component.ts
```diff
@@ -0,0 +1,432 @@
...
+  host: {
+    'class': 'ts-chip-remove ts-chip-trailing-icon',
+    '(click)': 'handleClick($event)',
+  },
+})
+export class TsChipRemoveDirective {
+  constructor(protected parentChip: TsChipComponent) {}
+
+  /**
+   * Calls the parent chip's public `remove()` method if applicable.
```
benjamincharity, line 458: Missing param

benjamincharity (changes requested):

terminus-ui/chip/src/chip-collection.component.ts
atlwendy: The comment is above - had extra line break though...

terminus-ui/chip/src/chip.component.scss
atlwendy: It's set right below...

terminus-ui/chip/src/chip.component.scss
atlwendy: Yes, it's the hover block below here...

terminus-ui/chip/src/chip.component.scss
atlwendy: I had thought since it's only one style block, putting in one style sheet would be okay?

terminus-ui/chip/src/chip.component.scss
atlwendy: No it's not - I wasn't sure whether you meant to hardcode the text size as `14px`?

terminus-ui/chip/src/chip.component.scss
atlwendy: I'll look into this.

terminus-ui/chip/src/chip-collection.component.spec.ts
atlwendy: Deferring the dom check once test on ci stabilizes - it's in my notes.

terminus-ui/chip/src/chip-collection.component.ts
benjamincharity: ah scanning too fast 🤦‍♂

terminus-ui/chip/src/chip.component.scss
benjamincharity: Yeah, Material chips are 14px but our typography scale doesn't currently include a 14px size. So for now I was suggesting to hardcode it with a note. Then when we update types styles (hopefully soon) we can refactor it out.

terminus-ui/chip/src/chip.component.scss
benjamincharity: Same comment with the rgba color below

terminus-ui/chip/src/chip.component.scss
benjamincharity: Weird.. confusing from the review mode.. makes it look like there are no styles other than background: 
Screen Shot 2019-09-09 at 5 53 37 AM

benjamincharity (approved):

benjamincharity (approved):

PR #1672, opened 2019-08-29
fix(Input): handle textual changes to date picker

deanterm:
The datepicker functionality has an issue when a user changes the text
value of the input after a date has been selected. Example of this:

Input value is set to:
08-22-2019

User decides they want change the date to:
08-23-2019

So, they put the cursor after the 22 and change the last 2 to a 3.
During this editing, if we update the value that the datepicker looks
at, it gets confused and the value ends up looking like:

8-232-019

This value cannot be parsed by the datepicker and we end up with a null
Date() value. However, the value inside the input looks correct. The
emitted value from the input, however, ends up being null.

We tried a few workarounds here.

1. Tried setting this.value = new Date(textualValue)
   This was a problem because the moment you removed the 2 in the
   example above, the value ended up looking like: 08-22-019. The
   2 from 2019 moved into the day section of the mask and that value
   is actually a valid date! The date turns out to parse into
   08-22-2019, which is what we already had.  This, in turn, would
   fire off an event on the datepicker which would re-update its
   internal stuff and kick the cursor to the end of the line.
   Making it literally impossible to change the date manually
   unless you deleted every character and reentered the whole
   date by hand.
2. Tried creating a hidden input element that the datepicker was tied to
   and then manually update the shown input based on changes.  This
   caused a lot of issues where we had to manually update the mask on
   every event and make sure we updated both values on the hidden and
   shown input.  This became messy quickly.
3. The solution that's being committed here.  We don't fire off a
   selected event on input changes. We instead just update the masked
   value and store the masked value in a separate store.  Then, on blur
   events, the change event will look to see if the date is null (due to
   the description of the original issue above) and if the textual value
   is not null. If so, the textual value wins and we create a new date
   object based on that info.  Otherwise, the datepicker value wins.
   This prevents the input text changing while the user is typing and
   throws out an event on blur with the updated value.

### DIFF
--- a/terminus-ui/input/src/input.component.ts
+++ b/terminus-ui/input/src/input.component.ts
@@ -394,7 +394,7 @@ export class TsInputComponent implements
       return true;
     }
 
-    return !!(input && !input.value && !this.isBadInput() && !this.autofilled);
+    return !!input && !input.value && !this.isBadInput() && !this.autofilled;
   }
 
   /**
@@ -844,6 +844,8 @@ export class TsInputComponent implements
   @Output()
   public readonly selected: EventEmitter<Date> = new EventEmitter();
 
+  // The textual value of the date entered into the input.
+  private textualDateValue = '';
 
   public constructor(
     private elementRef: ElementRef,
@@ -1111,6 +1113,7 @@ export class TsInputComponent implements
     } else {
       // Trigger the onTouchedCallback for blur events
       this.onTouchedCallback();
+      this.onDateChanged(this.value);
       this.inputBlur.emit(this.value);
     }
   }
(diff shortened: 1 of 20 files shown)


terminus-ui/input/src/input.component.ts
```diff
@@ -844,6 +844,8 @@ export class TsInputComponent implements
   @Output()
   public readonly selected: EventEmitter<Date> = new EventEmitter();
 
+  // The textual value of the date entered into the input.
+  private textualDateValue = '';
```
benjamincharity: Let's move this up with the other private properties at the top of this class (should go directly after the `textMaskInputElement` property

benjamincharity (changes requested):

benjamincharity (approved):

PR #1693, opened 2019-09-12
fix(Select): Text vertical alignment

### LINKED ISSUE #1639: Select: Text vertical alignment changes when a value is selected
Noticed in engage.

Select label is vertically aligned correctly with icons UNTIL a selection is made, then the text is too low:

atlwendy:
closes #1639 

image

### DIFF
--- a/terminus-ui/select/src/select.component.scss
+++ b/terminus-ui/select/src/select.component.scss
@@ -25,6 +25,12 @@ $ts-select-placeholder-arrow-space: 2 * ($ts-select-arrow-size + $ts-select-arro
   @include reset;
   display: block;
   outline: none;
+  .ts-form-field {
+    .ts-form-field__infix {
+      $select-form-field-padding-top: .6em;
+      padding: $select-form-field-padding-top 0 1em;
+    }
+  }
 
   // <div> trigger to open select
   .ts-select-trigger {


benjamincharity:
@atlwendy for UI changes like this, can you upload a screenshot/gif of the fix?

terminus-ui/select/src/select.component.scss
```diff
@@ -25,6 +25,12 @@ $ts-select-placeholder-arrow-space: 2 * ($ts-select-arrow-size + $ts-select-arro
   @include reset;
   display: block;
   outline: none;
+  .ts-form-field {
+    .ts-form-field__infix {
+      $select-form-field-padding: .6em;
```
benjamincharity: We should probably either a) add `-top` to to this variable name, or b) move all the padding numbers up into the definition.

benjamincharity (changes requested):
Apparently this comment input is required now... lame

terminus-ui/select/src/select.component.scss
atlwendy: I was actually wondering whether you'd call this out lol

Also, your overall comment made me think some about `input` component stuff is required when first glanced at it!

benjamincharity (approved):

PR #1695, opened 2019-09-16
feat(link): Allow email links without the external icon

### LINKED ISSUE #1627: Link: Allow email links without the external icon
When using a TsLink as an email or telephone link shows the external icon. Product has requested the ability to not show this for email and telephone links.

- [ ] Determine if the icon should show
  - Inside the setter for `destination`:
    - If `destination` is a string AND `mailto:` or `tel:` is found, set a flag to hide the icon
- [ ] Add test
- [ ] Add to usage docs as clarification below the `External links` header

atlwendy:
closes #1627 

mailto-tel

### DIFF
--- a/terminus-ui/link/src/link.component.ts
+++ b/terminus-ui/link/src/link.component.ts
@@ -4,6 +4,7 @@ import {
   Input,
   ViewEncapsulation,
 } from '@angular/core';
+import { isString } from '@terminus/ngx-tools';
 import { TsStyleThemeTypes } from '@terminus/ui/utilities';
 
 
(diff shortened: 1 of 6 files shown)


terminus-ui/link/src/link.component.ts
```diff
@@ -74,6 +83,12 @@ export class TsLinkComponent {
   @Input()
   public isExternal = false;
 
+  /**
+   * Decide whether an external icon should be shown
+   */
+  @Input()
```
benjamincharity: I don't believe we should expose this directly to consumers. They shouldn't have the option to remove the external icon for actual external links. I think what you've done using this flag internally is enough.

terminus-ui/link/src/link.component.ts
```diff
@@ -60,7 +60,16 @@ export class TsLinkComponent {
    * Define the link's destination
    */
   @Input()
-  public destination: string | string[] | undefined;
+  public get destination(): string | string[] | undefined {
+    return this._destination;
+  }
+  public set destination(value: string | string[] | undefined) {
```
benjamincharity, line 69: Our setters usually come before our getters 😬

terminus-ui/link/src/link.component.ts
```diff
@@ -60,7 +60,16 @@ export class TsLinkComponent {
...
+    if (typeof (value) === 'string' && (value.includes('mailto') || value.includes('tel'))) {
```
benjamincharity: We should use the ngx-tools `isString()` helper here

benjamincharity (changes requested):
_

benjamincharity (approved):
🎉

PR #1699, opened 2019-09-16
fix(select): text vertical alignment with or without float label

### LINKED ISSUE #1698: Select: text vertical alignment based on whether there's float label or not
When fixing select text vertical alignment, it wasn't taken into consideration that some select has float label, some doesn't. Text should always vertically aligned regardless the label position.

atlwendy:
closes #1698 

image (2)


benjamincharity:
@atlwendy Does this fix the issue with the text and icon alignment I posted to our channel? (I can't see the arrow in the screenshots)

benjamincharity (approved):

PR #1710, opened 2019-09-20
1688 consumer chip control

### LINKED ISSUE #1687: Chip: Documentation
- [ ] Add usage docs
- [ ] Increase component count badge number by 1
- [ ] Remove 'gzip size' badge and any related markdown links (unrelated but needs to be done)

### LINKED ISSUE #1688: Chip: Rework to allow ts-chip to be defined by the consumer
We want to allow consumers to use a `ts-chip` by itself OR to define the chips inside their template.

Example:

```html
<!-- Chip collection -->
<ts-chip-collection>
  <ts-chip *ngFor="let chip of myChips">
    {{ chip.value }}
  </ts-chip>
</ts-chip-collection>
```

- [x] Replace contents of `.ts-chip-collection-wrapper` with `<ng-content>`
- [x] Chip collection component no longer directly uses `ts-chip` in it's template
- [x] Autocomplete now passes `ts-chip` and `input` inside `ts-chip-collection`
- [x] Remove formatter functions from `ts-chip-collection`
- [x] Remove `chipsInput` `@Input` from the collection
- [x] Remove incorrectly formatted flex layout: `fxLyaout="row"` from the collection template
- [ ] Update usage docs as needed
- [ ] Update demo(s)
- [ ] Verify JSDocs include all the correct inputs/outputs for both the collection and chip
- [ ] Remove/update tests as needed to ensure coverage

benjamincharity:
Closes #1688 
Closes #1687

- chips now passed into collection
- removed material class mixins
- renamed blur functions and emitters to not use native event names or be prefixed with `on`
- reorganized public/private/static items
- removed directive for the remove icon since it is not controlled by the consumer
- selecting a chip via shift key now takes `isSelected` into account
- updated all test components
- updated demo and added more controls
- updated jsdoc comments
- added jsdoc links where applicable
- added usage docs
- converted errors to UI errors
- removed boolean coercion since we no longer do that for boolean inputs
- moved test components back to our standard import method to enable IDE support

### DIFF
--- a/terminus-ui/chip/src/chip-collection.component.spec.ts
+++ b/terminus-ui/chip/src/chip-collection.component.spec.ts
@@ -49,7 +49,6 @@ function createComponent<T>(component: Type<T>): ComponentFixture<T> {
 
 describe(`TsChipCollection`, function() {
   let fixture;
-
   let chipElements;
   let chipDebugElement;
   let chipNativeElement;
@@ -67,8 +66,8 @@ describe(`TsChipCollection`, function() {
   let BACKSPACE_EVENT: KeyboardEvent;
   let BACKSPACE_EVENT_CHIP: KeyboardEvent;
 
-  function setupStandardCollection(exampleComponent) {
-    fixture = createComponent(testComponents[exampleComponent]);
+  function setupStandardCollection(component) {
+    fixture = createComponent(component);
     fixture.detectChanges();
     chipCollectionElement = fixture.debugElement.query(By.directive(TsChipCollectionComponent));
     chipCollectionNativeElement = chipCollectionElement.nativeElement;
@@ -86,29 +85,32 @@ describe(`TsChipCollection`, function() {
   }
 
   test(`should exist`, function() {
-    setupStandardCollection('RegularChip');
+    setupStandardCollection(testComponents.RegularChip);
 
     expect(chipNativeElement).toBeTruthy();
   });
 
-  test(`should disable chip if isDisabled set`, function() {
-    setupStandardCollection('DisabledChip');
+  test(`should disable chip if isDisabled set`, fakeAsync(() => {
+    setupStandardCollection(testComponents.DisabledChip);
+    tick();
+    fixture.detectChanges();
 
     expect(chipInstance.isDisabled).toEqual(true);
-    expect(chipNativeElement.classList).toContain('ts-chip-disabled');
-  });
+    expect(chipNativeElement.classList).toContain('ts-chip--disabled');
+  }));
 
-  test(`should disable chip if isReadonly set`, function() {
-    setupStandardCollection('ReadonlyChip');
+  test(`should disable chip if isReadonly set`, fakeAsync(() => {
+    setupStandardCollection(testComponents.ReadonlyChip);
     chipCollectionInstance.isReadonly = true;
+    tick();
     fixture.detectChanges();
 
     expect(chipInstance.isRemovable).toBeFalsy();
-    expect(chipNativeElement.classList).not.toContain('c-chip-removable');
-  });
+    expect(chipNativeElement.classList).not.toContain('c-chip--removable');
+  }));
 
   test(`should be able to select`, fakeAsync(function() {
-    setupStandardCollection('RegularChip');
+    setupStandardCollection(testComponents.RegularChip);
 
     chipInstance.select();
     const chip1 = chipElements[0].nativeElement;
@@ -117,8 +119,8 @@ describe(`TsChipCollection`, function() {
     tick(1000);
     fixture.detectChanges();
 
-    expect(chip1.classList).toContain('ts-chip-selected');
-    expect(chip2.classList).not.toContain('ts-chip-selected');
+    expect(chip1.classList).toContain('ts-chip--selected');
+    expect(chip2.classList).not.toContain('ts-chip--selected');
   }));
 
   test(`should set value properly`, function() {
@@ -132,12 +134,25 @@ describe(`TsChipCollection`, function() {
     expect(instance.value).toEqual(['banana']);
   });
 
+  describe(`empty`, () => {
+
+    test(`should check the number of chips if the collection exists`, () => {
+      fixture = createComponent(testComponents.NoChip);
+      chipCollectionInstance = getChipCollectionInstance(fixture);
+      fixture.detectChanges();
+
+      expect(chipCollectionInstance.empty).toEqual(true);
+    });
+
+  });
+
   test(`should be able to set isSelectable from chip collection and populate to chips`, fakeAsync(function() {
-    setupStandardCollection('RegularChip');
+    setupStandardCollection(testComponents.RegularChip);
 
     chipCollectionInstance.isSelectable = true;
     const chip = chips.toArray()[0];
     fixture.detectChanges();
+
     chipInstance.select();
     tick(1000);
     fixture.detectChanges();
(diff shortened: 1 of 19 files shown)


terminus-ui/chip/src/chip-collection.component.spec.ts
```diff
@@ -67,8 +66,8 @@ describe(`TsChipCollection`, function() {
   let BACKSPACE_EVENT: KeyboardEvent;
   let BACKSPACE_EVENT_CHIP: KeyboardEvent;
 
-  function setupStandardCollection(exampleComponent) {
-    fixture = createComponent(testComponents[exampleComponent]);
+  function setupStandardCollection(component) {
```
benjamincharity, line 69: I thought about an enum here, but realized it wouldn't really get us anything vs what we're doing already.. basically same amount of typing. And using the string hampered the IDE ability to refactor or jump to the definition 😞

terminus-ui/chip/src/chip-collection.component.ts
```diff
@@ -222,41 +224,27 @@ export class TsChipCollectionComponent implements AfterViewInit, OnInit, OnDestr
...
-  @Input()
-  public set value(value: string[]) {
-    this._value = value;
-  }
-  public get value(): string[] {
-    return this._value;
-  }
-  protected _value = [''];
+  @Input('aria-orientation')
+  public ariaOrientation: 'horizontal' | 'vertical' = 'horizontal';
```
benjamincharity, line 247: Git seems a bit confused on some reorganization in this and the chip files - so be aware that many items were just moved rather than deleted.

terminus-ui/chip/src/chip-collection.component.ts
```diff
@@ -357,34 +346,43 @@ export class TsChipCollectionComponent implements AfterViewInit, OnInit, OnDestr
...
+
+  /**
+   * Initialize the key manager and listen for chip changes
+   */
+  public ngAfterViewInit(): void {
     this.keyManager = new FocusKeyManager<TsChipComponent>(this.chips)
       .withWrap()
       .withVerticalOrientation()
-      .withHorizontalOrientation(this.dir ? this.dir.value : 'ltr');
+      .withHorizontalOrientation('ltr');
```
benjamincharity, line 376: I ended up just removing directional support since we don't support that elsewhere. If we ever get into a `rtl` situation we'll need to address the library as a whole.

terminus-ui/chip/src/chip-collection.component.ts
```diff
@@ -597,44 +598,40 @@ export class TsChipCollectionComponent implements AfterViewInit, OnInit, OnDestr
...
+  private syncChipsState(): void {
+    // istanbul ignore else
+    if (this.chips && this.chips.length) {
       this.chips.forEach(chip => {
-        chip.isDisabled = this.isDisabled;
+        chip.allowMultiple = this.allowMultipleSelections;
         chip.chipCollectionMultiple = this.allowMultipleSelections;
+        chip.isDisabled = this.isDisabled;
         chip.isRemovable = !this.isReadonly && !this.isDisabled;
+        chip.isSelectable = this.isSelectable;
```
benjamincharity, line 632: Moved this here rather than the setter since it wasn't getting reflected to the chips if the flag was set before the chips where collected.

terminus-ui/chip/src/chip.component.spec.ts
```diff
@@ -17,61 +13,83 @@ import {
...
+    return createComponentInner(component, [], moduleImports);
   }
 
   describe('TsChip', () => {
-    let testComponent: testComponents.SingleChip;
-
-    beforeEach(() => {
-      fixture = createComponent(testComponents.SingleChip);
+    let testComponent;
+    function setup(component = testComponents.SingleChip) {
```
benjamincharity, line 40: The way this was set up originally required all the tests to use a single test host, so I needed to move to a setup function since I had a new test that needed a custom host.

terminus-ui/chip/src/chip-collection.component.spec.ts
```diff
@@ -86,29 +85,32 @@ describe(`TsChipCollection`, function() {
...
+  test(`should disable chip if isReadonly set`, fakeAsync(() => {
+    setupStandardCollection(testComponents.ReadonlyChip);
     chipCollectionInstance.isReadonly = true;
+    tick();
     fixture.detectChanges();
 
     expect(chipInstance.isRemovable).toBeFalsy();
-    expect(chipNativeElement.classList).not.toContain('c-chip-removable');
-  });
+    expect(chipNativeElement.classList).not.toContain('c-chip--removable');
```
atlwendy, line 109: Did you have to add `fakeAsync` and `tick` to get it class showing up?

terminus-ui/chip/src/chip-collection.component.spec.ts
atlwendy: 👍

terminus-ui/chip/src/chip-collection.component.spec.ts
```diff
@@ -132,44 +134,73 @@ describe(`TsChipCollection`, function() {
...
 
-    dispatchMouseEvent(chipRemovalButton, 'click');
-    tick(1000);
+    expect(chipCollectionInstance.chips.length).toEqual(2);
+
+    chipRemovalButton.click();
+    fixture.detectChanges();
+
+    expect(fixture.componentInstance.change.mock.calls[0][0].source.chips.length).toEqual(1);
+    expect(fixture.componentInstance.removed.mock.calls[0][0].chip.value).toEqual('banana');
```
atlwendy, line 176: What does `.mock.calls` return?

terminus-ui/chip/src/chip-collection.component.spec.ts
```diff
@@ -132,44 +134,73 @@ describe(`TsChipCollection`, function() {
...
+    expect.assertions(3);
+  });
+
+  test(`should emit the tabUpdateFocus event`, fakeAsync(() => {
+    setupStandardCollection(testComponents.Events);
+    fixture.componentInstance.tabbed = jest.fn();
+    const event = document.createEvent('KeyboardEvent');
+    event.initEvent('keydown', true, false);
+    Object.defineProperties(event, { code: { get: () => KEYS.TAB.code } });
+    Object.defineProperties(event, { keyCode: { get: () => KEYS.TAB.keyCode } });
```
atlwendy, line 186: We've been using this `.createEvent`, `.initEvent` and `Object.defineProperties` quite a bit to define a keyboard event... maybe could extract that out to ngx-tools? just a thought...

terminus-ui/chip/src/chip-collection.component.spec.ts
```diff
@@ -196,17 +228,27 @@ describe(`TsChipCollection`, function() {
...
     });
 
+    test(`should call the blur() method when a chip is blurred`, fakeAsync(() => {
+      setupStandardCollection(testComponents.RegularChip);
+      chipCollectionInstance.blur = jest.fn();
+      chipInstance.handleBlur();
+      fixture.detectChanges();
+      tick();
+
+      expect(chipCollectionInstance.blur).toHaveBeenCalled();
```
atlwendy, line 244: 👍 didn't write this out cause saw those lines were already covered...

terminus-ui/chip/src/chip-collection.component.ts
atlwendy: sounds good 👍

terminus-ui/chip/src/chip-collection.component.ts
```diff
@@ -357,34 +346,43 @@ export class TsChipCollectionComponent implements AfterViewInit, OnInit, OnDestr
...
-      .withHorizontalOrientation(this.dir ? this.dir.value : 'ltr');
+      .withHorizontalOrientation('ltr');
 
     this.keyManager.tabOut.pipe(untilComponentDestroyed(this)).subscribe(() => {
       this.tabUpdateFocus.emit();
     });
 
     // When the collection changes, re-subscribe
-    this.chips.changes.pipe(startWith(null), untilComponentDestroyed(this)).subscribe(() => {
+    this.chips.changes.pipe(startWith<void, null>(null), untilComponentDestroyed(this)).subscribe(() => {
```
atlwendy, line 383: What does adding `<void, null>` buy you?

terminus-ui/chip/src/chip-collection.component.ts
atlwendy: 👍

terminus-ui/chip/src/chip-collection.component.ts
```diff
@@ -491,30 +481,23 @@ export class TsChipCollectionComponent implements AfterViewInit, OnInit, OnDestr
...
-  public clickSelect(clickEvent) {
-    const shiftKey = clickEvent.event.shiftKey;
-    if (this.allowMultipleSelections && shiftKey) {
-      // Disable text highlight when shift click
-      document.onselectstart = function() {
-        return false;
-      };
-      clickEvent.chip.toggleSelected();
-      clickEvent.event.preventDefault();
-    }
```
atlwendy, line 503: Does shift + click work w/o this?
Oh I see you moved this to chip component....

terminus-ui/chip/src/chip.component.scss
```diff
@@ -4,53 +4,66 @@
...
-    background-color: color(primary);
+  &.ts-chip--selected {
     color: color(pure);
+    $themes: primary accent warn;
+
+    @each $theme in $themes {
+      &.ts-chip--#{$theme} {
+        background-color: color(#{$theme});
+      }
+    }
```
atlwendy, line 49: Nice loop... I haven't yet used loops in css so far...

terminus-ui/chip/src/chip.component.spec.ts
atlwendy: yep. at the time I didn't need a 2nd host component.. thus set up that way.

terminus-ui/chip/testing/src/test-components.ts
```diff
@@ -51,39 +54,34 @@ export class OneChip {
...
+    ></ts-chip>
   `,
 })
 export class SingleChip {
   public option = 'banana';
   public selected = false;
-  removed: (event?: TsChipEvent) => void = () => { }
-  chipDestroy: (event?: TsChipEvent) => void = () => { };
-  selectionChange: (event?: TsChipSelectionChange) => void = () => { };
+  public removed: (event?: TsChipEvent) => void = () => { };
```
atlwendy, line 66: I'm suprised lint didn't add `;` here...

atlwendy (approved):

terminus-ui/chip/src/chip-collection.component.spec.ts
benjamincharity: An array of the calls made to the spy https://jestjs.io/docs/en/mock-functions#using-a-mock-function

terminus-ui/chip/src/chip-collection.component.spec.ts
benjamincharity: yeah, the issue is that we already have this in the tools library, but seem to hit odd issues where we need to create our own. I haven't pinpointed what happens so I'm not sure what we'd need.

I did notice we are setting the `code` to both the `key` and the `code` here https://github.com/GetTerminus/ngx-tools/blob/release/ngx-tools/testing/src/utilities/events/create-keyboard-event.ts#L30-L32

And I noticed that the CDK was listening for the actual number code under the `keyCode` property. So possibly just adding that to the helper function may work.

terminus-ui/chip/src/chip-collection.component.ts
benjamincharity: It makes TypeScript happy. The version of `startWith` that doesn't have types passed in is deprecated while this version is not.

```typescript
startWith<T, D>(...array: Array<T | SchedulerLike>): OperatorFunction<T, T | D>
```

terminus-ui/chip/testing/src/test-components.ts
benjamincharity: Yeah, something is odd with our semicolon lint rules. Originally I couldn't get ESLint to fix that issue so I left TSLint in charge.. wondering if something is broken in this combined ES/TSLint stage..

terminus-ui/chip/src/chip-collection.component.spec.ts
benjamincharity: Yeah, I think it's because the initialization is in a Promise resolve now.

PR #1720, opened 2019-09-26
docs: add code style to dev docs

### LINKED ISSUE #1680: Docs: Add more conventions to the development readme
- Add new documentation to the development readme:
  - [x] member ordering
  - [x] underscore convention
  - [x] setters & getters
  - [x] decorators

### Member ordering

Currently most files have members in (an) order. We should formalize that pattern for consistency.

1. private properties
1. public properties
1. getters
    - getter before setter before private property
1. view references
1. inputs
1. outputs
1. constructor
1. lifecycle methods (in the order they are called by Angular)
1. public methods
1. protected/static methods
1. private methods

### Underscores

Prefixing private methods and properties is a common pattern. We've opted to rely on TypeScript's member access rather than the underscore. We should make it clear for contributors that we don't follow that pattern.

Caveat: The one place we do use the underscore is to prefix the private property that is used with a getter and setter. eg

```typescript
public set active(v: boolean) {
  this._foo = v;
}
public get active(): boolean {
  return this._foo;
}
private _foo = false;
```

### Setters & Getters

Our setters always come before the getter; which is in turn before the private property. (See above code example)

### Decorators

Decorators go on their own line above the item they are decorating.

```typescript
@Decorator()
public foo;
```

benjamincharity:
ISSUES CLOSED: #1680

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -19,6 +19,12 @@
   - [Releasing](#releasing)
   - [Code Comments](#code-comments)
   - [Usage Docs](#usage-docs)
+- [Code Style](#code-style)
+  - [Lint configurations](#lint-configurations)
+  - [Member Ordering](#member-ordering)
+  - [Underscores](#underscores)
+  - [Setters & Getters](#setters--getters)
+  - [Decorators](#decorators)
 - [Issues](#issues)
 
 <!-- END doctoc generated TOC please keep comment here to allow auto update -->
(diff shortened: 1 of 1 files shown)


DEVELOPMENT.md
```diff
@@ -316,6 +322,61 @@ When editing or creating usage documentation, add headlines and appropriate cont
...
+
+```typescript
+public set active(v: boolean) {
+  this._foo = v;
+}
+public get active(): boolean {
+  return this._foo;
+}
+private _foo = false;
+```
```
atlwendy, line 366: But you just said getter goes before setter 😂 
I think our code base has setter first, then getter..

benjamincharity: Whoops. Good catch.

benjamincharity: Updated

atlwendy (approved):

PR #1731, opened 2019-10-02
SelectionList bugs

### LINKED ISSUE #1727: SelectionList: Option should scroll into view when focused with the keyboard
Currently, if the user uses the arrow keys to move down the list of options, they can move their selection out of view:

The focused option should scroll into view as the user moves down the list.

See https://github.com/GetTerminus/terminus-ui/blob/b36c77a9d49fd07f96ddc25c36ca9f331400b589/terminus-ui/autocomplete/src/autocomplete-panel/autocomplete-trigger.directive.ts#L759-L772

---

- [ ] Add functionality to scroll focused option into view
- [ ] Add test

### LINKED ISSUE #1729: SelectionList: The first option should always be selected when querying
Notice, when typing a query to narrow results, hitting `enter` selects the originally focused option. The first option should always be selected when querying.

NOTE: May want to handle with #1728 as they are closely related.

- [x] As the options narrow, keep the first option selected until the user interacts with the list
- [ ] Add tests

### LINKED ISSUE #1730: SelectionList: Should have correct hover styles
The selection list should appear to allow interaction - currently the cursor is simply the default cursor.

- [x] Should have cursor hand (unless hovering actual `<input>`)

benjamincharity:
- cursor correct on hover. closes #1730 
- form field hover now has correct styles
- focused options are now scrolled into view. closes #1727 
- first option focused by default. closes #1729

### DIFF
--- a/terminus-ui/selection-list/src/selection-list-panel/selection-list-trigger.directive.ts
+++ b/terminus-ui/selection-list/src/selection-list-panel/selection-list-trigger.directive.ts
@@ -770,7 +770,8 @@ export class TsSelectionListTriggerDirective<ValueType = string> implements Cont
         // Create a new stream of panelClosingActions, replacing any previous streams that were created, and flatten it so our stream only
         // emits closing events...
         switchMap(() => {
-          this.resetActiveItem();
+          // Focus the first option when options change
+          this.selectionListPanel.keyManager.setActiveItem(0);
           this.selectionListPanel.setVisibility();
 
           return this.panelClosingActions;
@@ -820,14 +821,12 @@ export class TsSelectionListTriggerDirective<ValueType = string> implements Cont
       // top of the option, because it allows the user to read the top group's label.
       this.selectionListPanel.scrollTop = 0;
     } else {
-      const newScrollPosition = getOptionScrollPosition(
+      this.selectionListPanel.scrollTop = getOptionScrollPosition(
         index + labelCount,
         this.itemHeight,
         this.selectionListPanel.scrollTop,
         SELECTION_LIST_PANEL_MAX_HEIGHT,
       );
-
-      this.selectionListPanel.scrollTop = newScrollPosition;
     }
   }
 
--- a/terminus-ui/selection-list/src/selection-list-panel/selection-list-trigger.directive.ts
+++ b/terminus-ui/selection-list/src/selection-list-panel/selection-list-trigger.directive.ts
@@ -28,12 +28,11 @@ import { ControlValueAccessor } from '@angular/forms';
 import { TsDocumentService } from '@terminus/ngx-tools/browser';
 import { coerceBooleanProperty } from '@terminus/ngx-tools/coercion';
 import { KEYS } from '@terminus/ngx-tools/keycodes';
-import {
-  isUnset,
-  untilComponentDestroyed,
-} from '@terminus/ngx-tools/utilities';
+import { untilComponentDestroyed } from '@terminus/ngx-tools/utilities';
 import { TsFormFieldComponent } from '@terminus/ui/form-field';
 import {
+  countGroupLabelsBeforeOption,
+  getOptionScrollPosition,
   TsOptionComponent,
   TsOptionSelectionChange,
 } from '@terminus/ui/option';
@@ -72,6 +71,9 @@ export const TS_SELECTION_LIST_SCROLL_STRATEGY_FACTORY_PROVIDER = {
   deps: [Overlay],
   useFactory: TS_SELECTION_LIST_SCROLL_STRATEGY_FACTORY,
 };
+
+// The max height of the select's overlay panel
+export const SELECTION_LIST_PANEL_MAX_HEIGHT = 256;
 
 // Unique ID for each instance
 let nextUniqueId = 0;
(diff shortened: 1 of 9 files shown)


terminus-ui/selection-list/src/selection-list-panel/selection-list-trigger.directive.ts
```diff
@@ -73,6 +72,9 @@ export const TS_SELECTION_LIST_SCROLL_STRATEGY_FACTORY_PROVIDER = {
   useFactory: TS_SELECTION_LIST_SCROLL_STRATEGY_FACTORY,
 };
 
+// The max height of the select's overlay panel
+export const SELECTION_LIST_PANEL_MAX_HEIGHT = 256;
```
atlwendy, line 76: Is it the max height regardless current window/screen size?

terminus-ui/selection-list/src/selection-list-panel/selection-list-trigger.directive.ts
```diff
@@ -211,6 +213,18 @@ export class TsSelectionListTriggerDirective<ValueType = string> implements Cont
...
+  /**
+   * Calculates the height of the options
+   *
+   * Only called if at least one option exists
+   */
+  private get itemHeight(): number {
+    // Try to use the 2nd option in case the first option is blank or a filter etc. Fall back to the first item if needed.
+    const options = this.selectionListPanel.options.toArray();
+    const option = options[1] || options[0];
+    return option.elementRef.nativeElement.offsetHeight;
```
atlwendy, line 225: Can the option height be defined by consumers?

terminus-ui/selection-list/src/selection-list-panel/selection-list-trigger.directive.ts
```diff
@@ -752,7 +770,8 @@ export class TsSelectionListTriggerDirective<ValueType = string> implements Cont
         // Create a new stream of panelClosingActions, replacing any previous streams that were created, and flatten it so our stream only
         // emits closing events...
         switchMap(() => {
-          this.resetActiveItem();
+          // Focus the first option when options change
+          this.selectionListPanel.keyManager.setActiveItem(0);
```
atlwendy, line 774: Does `resetActiveItem` also do the same thing - set the first option active? although, this way is clearer...

terminus-ui/selection-list/src/selection-list.component.spec.ts
```diff
@@ -965,4 +965,24 @@ describe(`TsSelectionListComponent`, function() {
     expect(fixture.componentInstance.clicked).toHaveBeenCalled();
   }));
 
+  // Note: Simply dispatching the arrow down key on the trigger repeatedly did not wor
```
atlwendy: your keyboard might have cut `k` out...

terminus-ui/selection-list/src/selection-list.component.spec.ts
```diff
@@ -965,4 +965,24 @@ describe(`TsSelectionListComponent`, function() {
...
+    expect(instance.panel.keyManager.activeItemIndex).toEqual(-1);
+
+    const input = getSelectionListInput(fixture);
+    const states = fixture.componentInstance.states;
+    const name = states[3].name.substring(0, 2);
+    typeInElement(name, input);
+    tick(1000);
+    fixture.detectChanges();
+
+    expect(instance.panel.keyManager.activeItemIndex).toEqual(0);
```
atlwendy, line 985: This however doesn't test using arrow key that option list should scroll when list is longer than viewport height - do notice that arrow key doesn't work.... we do have other specs that using arrow down key, right? wonder how they're different..

atlwendy (approved):
how could the coverage go down? 😱

benjamincharity:
> how could the coverage go down?

Cause we are unfortunately under the gun currently to get some items out, so I'm leaving several `test.todo`s in the code which we will need to circle back on.

terminus-ui/selection-list/src/selection-list-panel/selection-list-trigger.directive.ts
benjamincharity: yep, just like select and autocomplete.

terminus-ui/selection-list/src/selection-list-panel/selection-list-trigger.directive.ts
benjamincharity: It pretty much already is. We are checking the rendered height, so if the consumer puts in `200px` tall content, that's what it will be. Our default height is just created by our CSS. So the template drives the height.

terminus-ui/selection-list/src/selection-list-panel/selection-list-trigger.directive.ts
benjamincharity: `resetActiveItem` clears the selection completely. Which mentally seems fine until you start using the component. The UX of being able to just type to filter and hit `enter` is just better imo

terminus-ui/selection-list/src/selection-list.component.spec.ts
benjamincharity: I wouldn't expect it to? This test is called `should focus first option when the options collection changes` and that is all it tests. There is a TODO test above this to cover that called `should update panel scroll position when focusing an out-of-view option`

terminus-ui/selection-list/src/selection-list.component.spec.ts
benjamincharity: Fixed

terminus-ui/selection-list/src/selection-list.component.spec.ts
atlwendy: Okay, I didn't see a todo and that's all the specs you added so just wanted to point it out...

PR #1732, opened 2019-10-03
1728 selectionlist select text

### LINKED ISSUE #1728: SelectionList: Focusing the input should select the entire text
Currently, clicking into the single-mode version just places the cursor at the end of the input text rather than selecting it all which causes more work than necessary if the user wants to change the value.

Note: When tabbing to the single mode, all text is selected - just not when clicking into it.

---

- [ ] When user focuses the list in single mode, all the input text should be highlighted.
- [ ] Add test

#### Resources

- https://stackoverflow.com/a/4067488/722367

benjamincharity:
closes #1728 

NOTE: Jest currently doesn't have support for testing document ranges. I've added a todo test with a link to the Jest issue. In the copy component we got around this limitation by testing our implementation rather than the actual document selection. In this case we don't really have any implementation to test.

### DIFF
--- a/terminus-ui/selection-list/src/selection-list.component.spec.ts
+++ b/terminus-ui/selection-list/src/selection-list.component.spec.ts
@@ -985,4 +985,8 @@ describe(`TsSelectionListComponent`, function() {
     expect(instance.panel.keyManager.activeItemIndex).toEqual(0);
   }));
 
+  // TODO: Not currently supported by Jest
+  //   Waiting for https://github.com/jsdom/jsdom/issues/317 to land
+  test.todo(`should select all existing text when the input is focused`);
+
 });
--- a/terminus-ui/selection-list/src/selection-list.component.ts
+++ b/terminus-ui/selection-list/src/selection-list.component.ts
@@ -653,7 +653,7 @@ export class TsSelectionListComponent implements
         const controlValue = this.ngControl.value;
         // If the input value doesn't match the selection, then the user must have edited the input value
         // istanbul ignore else
-        if (inputValue !== ((controlValue && controlValue.length > 0) ? this.displayFormatter(controlValue[0]) : '')) {
+        if (controlValue && inputValue !== ((controlValue && controlValue.length > 0) ? this.displayFormatter(controlValue[0]) : '')) {
           this.ngControl.value.length = 0;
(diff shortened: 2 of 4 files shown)


terminus-ui/selection-list/src/selection-list.component.spec.ts
```diff
@@ -985,4 +985,8 @@ describe(`TsSelectionListComponent`, function() {
     expect(instance.panel.keyManager.activeItemIndex).toEqual(0);
   }));
 
+  // TODO: Not currently supported by Jest
+  //   Waiting for https://github.com/jsdom/jsdom/issues/317 to land
+  test.todo(`should select all existing text when the input is focused`);
```
atlwendy, line 990: Ah this todo line is what I meant in last pr, I saw this line once but it was gone when reviewed again...

terminus-ui/selection-list/src/selection-list.component.ts
```diff
@@ -653,7 +653,7 @@ export class TsSelectionListComponent implements
         const controlValue = this.ngControl.value;
         // If the input value doesn't match the selection, then the user must have edited the input value
         // istanbul ignore else
-        if (inputValue !== ((controlValue && controlValue.length > 0) ? this.displayFormatter(controlValue[0]) : '')) {
+        if (controlValue && inputValue !== ((controlValue && controlValue.length > 0) ? this.displayFormatter(controlValue[0]) : '')) {
```
atlwendy, line 656: I'm surprised that ts wasn't yelling at you w/o that check..

atlwendy (approved):

terminus-ui/selection-list/src/selection-list.component.spec.ts
benjamincharity: Ahh weird.. I definitely left it in!

terminus-ui/selection-list/src/selection-list.component.ts
benjamincharity: I think it's cause we always assume the control has a string value. I only noticed cause the demo had controls initialized to `null` so this broke as `cannot read length of null`. I'd hope TS would yell at people for this, but did this just in case.

PR #1741, opened 2019-10-08
1617 table column sizing

### LINKED ISSUE #1617: Table: Dev controlled column width
Product is requesting the ability for a) consumers to better control the column widths, and to allow users to resize columns manually.

#### Consumer control over width

Our goal is to extend the ability of the current `columns` input.

- Create a new input called `columnsDef: ReadonlyArray<TsColumn>`
  - We should clone this array so that we can work on it freely.
  - When set, the column names are pulled out into their own array and assigned to the property `columns` (this is to enable our row directive to continue working with the CDK row directive as it expects this property)
- Each column should default to `100px` for its width definition when none is set.
  - This base value should be overwritten by the `width` param of a column definition.
- Defined widths are used by setting it as the value for `flex-basis` for a column.
  - eg `flex-basis: 100px`

```TypeScript
export type TsColumn = {
  name: string;   // ie 'Name'
  width?: string; // ie '200px' or '14rem' etc
}
```

#### User resize ability

Moved to #1740

benjamincharity:
Closes #1617 

BREAKING CHANGE:

- `TsColumn` definitions must be passed to `TsTable`.
- `minWidth` support has been removed in favor of `TsColumn` definitions.
- Scroll/overflow now managed by the consumer. So if always-visible scrollbars are desired, the consumer can add the mixin `visible-scrollbars` to their SCSS.

### DIFF
--- a/terminus-ui/table/src/cell.ts
+++ b/terminus-ui/table/src/cell.ts
@@ -1,5 +1,5 @@
-// NOTE(B$): In this case, we want an actual selector (so content can be nested inside) for our
-// directive. So we are disabling the `directive-selector` rule
+// NOTE: In this case, we want an actual selector (so content can be nested inside) for our directive. So we are disabling the
+// `directive-selector` rule
 // tslint:disable: directive-selector
 import {
   CdkCell,
@@ -11,11 +11,21 @@ import {
 import {
   Directive,
   ElementRef,
+  Host,
   Input,
   isDevMode,
+  Optional,
   Renderer2,
+  SkipSelf,
 } from '@angular/core';
+import { TsUILibraryError } from '@terminus/ui/utilities';
 
+import { TsTableComponent } from './table.component';
+
+
+/**
+ * Allowed column alignments
+ */
 export type TsTableColumnAlignment
   = 'left'
   | 'center'
@@ -25,7 +35,11 @@ export type TsTableColumnAlignment
 /**
  * An array of the allowed {@link TsTableColumnAlignment} for checking values
  */
-export const tsTableColumnAlignmentTypesArray: TsTableColumnAlignment[] = ['left', 'center', 'right'];
+export const tsTableColumnAlignmentTypesArray: TsTableColumnAlignment[] = [
+  'left',
+  'center',
+  'right',
+];
 
 
 /**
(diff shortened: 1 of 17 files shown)


benjamincharity:
smh CodeCov.. Select coverage supposedly went down with this PR but no select files are included in this PR..

terminus-ui/table/src/cell.ts
```diff
@@ -71,20 +85,25 @@ export class TsHeaderCellDefDirective extends CdkHeaderCellDef {}
 })
 export class TsHeaderCellDirective extends CdkHeaderCell {
   constructor(
-    columnDef: CdkColumnDef,
-    elementRef: ElementRef,
+    public columnDef: CdkColumnDef,
     public renderer: Renderer2,
+    private elementRef: ElementRef,
+    @Optional() @Host() @SkipSelf() parent: TsTableComponent<unknown>,
```
atlwendy, line 91: This means that this child component needs to access parent component's properties?

terminus-ui/table/src/cell.ts
```diff
@@ -100,44 +119,64 @@ export class TsHeaderCellDirective extends CdkHeaderCell {
...
   ) {
     super(columnDef, elementRef);
-
-    // NOTE: We are adding `noWrap` to the column in `TsColumnDefDirective` which doesn't exist
-    // in the `CdkColumnDef` so we cast it to 'any'.
-    // tslint:disable-next-line no-any
-    const column: any = columnDef;
+    // NOTE: Changing the type in the constructor from `CdkColumnDef` to `TsColumnDefDirective` doesn't seem to play well with the CDK.
+    // Coercing the type here is a hack, but allows us to reference properties that do not exist on the underlying `CdkColumnDef`.
+    this.column = columnDef as TsColumnDefDirective;
```
atlwendy, line 133: can `columnDef` be defined as `TsColumnDefDirective`? oh I guess you'd need to use some `CdkColumnDef` built in method

terminus-ui/table/src/row.ts
```diff
@@ -11,6 +11,7 @@ import {
   Directive,
   ViewEncapsulation,
 } from '@angular/core';
+import { TsTableComponent } from './table.component';
```
atlwendy, line 14: was wondering how this could have worked without the import,.. then found out it lives in the comments..

atlwendy (approved):

terminus-ui/table/src/cell.ts
benjamincharity: Correct. This is what I had to refactor in the feature I'm working on now. So in the next PR you'll see only parent refer to child.

terminus-ui/table/src/cell.ts
benjamincharity: I honestly thought it could since mine is an extension of theirs, but everything broke when I did. This doesn't feel great, but I was trying to keep moving and technically `columnDef` _is_ `TsColumnDefDirective` so I don't think I'm cheating _too_ much

PR #1742, opened 2019-10-08
feat(Input): Emit when user pastes into the input

coreyterminator:
Adds a paste output attribute to the ts-select control. It simply acts as a passthrough, emitting a `ClipboardEvent` on `inputPaste` when the native input element emits a paste event.


terminus-ui/input/src/input.component.ts
```diff
@@ -843,6 +843,9 @@ export class TsInputComponent implements
   @Output()
   public readonly inputFocus: EventEmitter<boolean> = new EventEmitter();
 
+  @Output()
```
benjamincharity, line 849: Need a comment matching the format of other Outputs.

Also need a test for this. (should be able to find an example by searching for one of the other Outputs in the spec file)

terminus-ui/input/src/input.component.md
```diff
@@ -320,6 +320,7 @@ Supported event emitters:
   (cleared)="inputWasBlurred($event)"
   (inputFocus)="inputGainedFocus($event)"
   (selected)="newDateSelected($event)"
+  (inputPaste)="inputValuePasted($event)"
```
benjamincharity, line 323: Just below this our usage docs state `Each emitter is called with the current input value.` - can you just remove this line for now? (I'll go in the future and update the docs format to show the payload for each output to match our newer docs format)

benjamincharity (changes requested):
Also, typo in the commit message: `attriute`. (sorry to be picky, but this commit will become the release note!)

benjamincharity (approved):

PR #1743, opened 2019-10-16
01: 1740 user resizable table columns

### LINKED ISSUE #1740: Table: User resizable columns
Splitting into it's own ticket here.

#### User resize ability

- This should not require a 3rd party library, but several examples have been added below for inspiration.
- A user should be able to resize a column by dragging the right border of a header cell.
- We should use the resize cursor when hovered.
- The user should not have to grab the 1px border - there should be a larger target area.
- The entire column should resize in real time while the cell edge is dragged.
- As a cell edge is resized, we should be updating our internal column definition.
- Anytime a drag ends, an event should be fired that exposes the current columns definition through an event emitter `columnResizeEnd: TsColumn[]` (this allows consumers to store the current column settings so they can be restored on the next visit.
- *Please Test*: How easy/hard is it for a user to resize the last column in a row?

### Open Questions

- How is the last column in a row resized?
- Is column resize ever disabled? (currently assuming no)
  - If so, we will need the UI to clearly indicate when this is/is not allowed.
  - If not, we can remove the columnDef input.

### Resources

- Example resizing with angular components table: https://stackblitz.com/edit/angular-rtfc5v
  - Related repo: https://github.com/lujian98/Angular-material-table-Resize
  - Resize is based off total space so doesn't account for scrollable container
- Example code from issue: https://github.com/angular/components/issues/8312#issuecomment-441729871
- Example with flexbox: https://codepen.io/Reklino/pen/raRaXq/
- Simple example: https://stackblitz.com/edit/angular-resizer-directive?file=app%2Fresizable.ts
- Example using 3rd party library: https://stackblitz.com/edit/mat-table-resize
  - Not as smooth since the whole column isn't resized during drag
  - Library: https://mattlewis92.github.io/angular-resizable-element/
- https://github.com/lujian98/Angular-material-table-Resize
  - https://stackblitz.com/edit/angular-rtfc5v

benjamincharity:
Closes #1740 

- Table: User resizable columns
- Demo: Update table demo with more functionality
- CSS: Allow CSS cursor `col-resize`
- Autocomplete: Add missing export
- Sort: Better visual representation of the ability to sort

tableResizeDemo

### DIFF
--- a/demo/app/components/table/table.component.ts
+++ b/demo/app/components/table/table.component.ts
@@ -4,12 +4,16 @@ import {
   Component,
   ViewChild,
 } from '@angular/core';
+import { DomSanitizer, SafeHtml } from '@angular/platform-browser';
 import {
   TsPaginatorComponent,
   TsPaginatorMenuItem,
 } from '@terminus/ui/paginator';
 import { TsSortDirective } from '@terminus/ui/sort';
-import { TsTableDataSource } from '@terminus/ui/table';
+import {
+  TsColumn,
+  TsTableDataSource,
+} from '@terminus/ui/table';
 import {
   merge,
   Observable,
@@ -76,10 +80,14 @@ const COLUMNS_SOURCE_GITHUB = [
 
 export interface GithubApi {
   items: GithubIssue[];
+  // NOTE: Format controlled by GitHub
+  // eslint-disable-next-line camelcase
   total_count: number;
 }
 
 export interface GithubIssue {
+  // NOTE: Format controlled by GitHub
+  // eslint-disable-next-line camelcase
   created_at: string;
   number: string;
   state: string;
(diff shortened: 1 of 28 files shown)


demo/app/components/table/table.component.ts
```diff
@@ -108,35 +118,70 @@ export class ExampleHttpDao {
...
+      width: '400px',
+    },
+    {
+      name: 'number',
+    },
+    {
+      name: 'updated',
+      // width: '300px',
+    },
+    // { name: 'comments' },
```
benjamincharity, line 152: Leaving these commented lines in so that when we're developing we can easily change column definitions without rewriting them each time.

terminus-ui/table/src/table.component.scss
```diff
@@ -1,51 +1,129 @@
...
+  --header-bg: #{color(utility, xlight)};
+  --header-text-color: #{color(utility, dark)};
+  --border-color: #{color(utility, light)};
+  $primary: color(primary);
+  --row-bg--hover: #{desaturate(lighten($primary, 66%), 70%)};
+  // NOTE: This must be above 40 as that is the number of header cell z-indexes set
+  --header-cell-sticky-z-index: 50;
+  --row-cell-sticky-z-index: 3;
+  --z-index-negative: -1;
+  --z-index-base-context: 1;
```
benjamincharity: Just a heads up, I've started using CSS variables more rather than only SCSS variables. I think this will help us skin items easier in the future (ie rather than generating separate styles for primary button, accent button, etc, we can just change the CSS variable).

benjamincharity:
NOTE: I just realized the add/remove columns select is broken. I'm currently working on a new implementation for that in my next issue. I'll leave broken for now since the refactor will go into `v15` prior to release.

atlwendy (approved):
Is this the rework of pr 1741?

benjamincharity:
Each of these will have some rework, but #1741 was more about letting the consumer pass in widths and this one is more about the user resizing. But very much all touching the same code

PR #1749, opened 2019-10-21
TABLE 02: re-orderable columns

### LINKED ISSUE #1717: Menu: Add ability to drag'n'drop items
Originally was `Table: Re-orderable columns` but we are now enabling this functionality through drag'n'drop functionality in the column selection menu.

Support via CDK drag'n'drop.

- [x] Add support for the `move` cursor to the cursors mixin
- [x] Draggable items should have visual representation on the far right side using the icon `drag_indicator`
- [x] Cursor should change to `move` when hovering a movable item
- [ ] Drag position should be locked so it can only move on the `y` axis
- [ ] Items must be dragged by the handle (not the entire list item)
- [ ] Should have custom placeholder design
- [ ] Dragging should be restricted to inside the menu panel

#### Resources

- https://material.angular.io/cdk/drag-drop/overview

benjamincharity:
Closes #1717 

- SCSS: Add cursor support for `move` and `ns-resize`
- Table: Update label styles
- Table: Support reorderable columns
- Table: Update demo to show reorderable columns
- Table: Add usage docs for resizable columns

Kapture 2019-10-21 at 15 21 36

### DIFF
--- a/terminus-ui/checkbox/src/checkbox.component.scss
+++ b/terminus-ui/checkbox/src/checkbox.component.scss
@@ -60,11 +60,20 @@ $ts-select-item-height: 3em !default;
   }
 
   // Target checkboxes inside a material menu dropdown panel
-  .mat-menu-panel & {
+  .ts-menu__panel & {
     // Match the material button line-height and spacing
     .c-checkbox {
       line-height: 2em;
       padding: 0 spacing(default);
+    }
+    
+    // By default the label doesn't fill the width even though the cursor signifies interaction
+    .mat-checkbox-layout {
+      width: 100%;
+    }
+
+    .mat-checkbox-label {
+      flex: 1;
     }
   }
 
--- a/terminus-ui/checkbox/src/checkbox.component.scss
+++ b/terminus-ui/checkbox/src/checkbox.component.scss
@@ -66,7 +66,13 @@ $ts-select-item-height: 3em !default;
       line-height: 2em;
       padding: 0 spacing(default);
     }
-    
+
+    &.cdk-drag {
+      .c-checkbox {
+        padding: 0 spacing(large, 1) 2px spacing(default);
+      }
+    }
+
     // By default the label doesn't fill the width even though the cursor signifies interaction
     .mat-checkbox-layout {
       width: 100%;
--- a/terminus-ui/sort/src/sort-header.component.html
+++ b/terminus-ui/sort/src/sort-header.component.html
@@ -7,6 +7,7 @@
     type="button"
     [attr.aria-label]="_intl.sortButtonLabel(id)"
     [attr.disabled]="_isDisabled() || null"
+    (click)="_handleClick()"
   >
     <ng-content></ng-content>
 
--- a/terminus-ui/table/src/cell.ts
+++ b/terminus-ui/table/src/cell.ts
@@ -18,11 +18,16 @@ import {
   Output,
   Renderer2,
 } from '@angular/core';
+import { untilComponentDestroyed } from '@terminus/ngx-tools';
 import {
   TsDocumentService,
   TsWindowService,
 } from '@terminus/ngx-tools/browser';
 import { TsUILibraryError } from '@terminus/ui/utilities';
+import {
+  fromEvent,
+  Subscription,
+} from 'rxjs';
 
 import {
   TsColumnDefDirective,
@@ -30,6 +35,7 @@ import {
 } from './column';
 
 
+export const TS_TABLE_MIN_COLUMN_WIDTH = 70;
 
 /**
  * Set the column alignment styles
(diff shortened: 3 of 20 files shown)


benjamincharity:
NOTE: Found 1 small bug that I'll try to get fixed prior to merge. Not a blocker but nice to have so it shouldn't block the review.

benjamincharity:
Alright, just pushed a fix for the small issue I found. Consumer must trigger `cdr.detectChanges()` when they update the column order. Otherwise the custom column widths don't get re-set until the menu closes. 

Added to the demo and the usage docs.

terminus-ui/checkbox/src/checkbox.component.scss
```diff
@@ -60,12 +60,27 @@ $ts-select-item-height: 3em !default;
...
+    }
+
+    // By default the label doesn't fill the width even though the cursor signifies interaction
+    .mat-checkbox-layout {
+      width: 100%;
+    }
+
+    .mat-checkbox-label {
+      flex: 1;
+    }
```
atlwendy, line 83: Do we overwrite material css here?

terminus-ui/sort/src/sort-header.component.html
```diff
@@ -7,6 +7,7 @@
     type="button"
     [attr.aria-label]="_intl.sortButtonLabel(id)"
     [attr.disabled]="_isDisabled() || null"
+    (click)="_handleClick()"
```
atlwendy, line 10: curious why you have `_` here?

terminus-ui/table/src/cell.ts
```diff
@@ -237,18 +213,25 @@ export class TsHeaderCellDirective extends CdkHeaderCell implements AfterViewIni
     resizeElement.title = 'Drag to resize column.';
     this.renderer.appendChild(this.elementRef.nativeElement, resizeElement);
 
-    // Remove any existing event listeners
-    this.removeAllEventListeners();
-    this.resizeListenerRemover = this.renderer.listen(resizeElement, 'mousedown', e => this.onResizeColumn(e));
+    // Handle column resize events
+    fromEvent<MouseEvent>(resizeElement, 'mousedown')
+      .pipe(untilComponentDestroyed<MouseEvent>(this))
+      .subscribe(e => this.onResizeColumn(e));
```
atlwendy, line 219: like how clean this is 👍

terminus-ui/table/src/table.component.ts
```diff
@@ -14,7 +14,7 @@ import {
   Inject,
   Input,
   IterableDiffers,
-  NgZone,
+  NgZone, OnDestroy,
```
atlwendy: Should this go down one line? webstorm seems always trying to sneak these in lol

terminus-ui/table/src/table.component.ts
```diff
@@ -229,31 +241,30 @@ export class TsTableComponent<T = any> extends CdkTable<T> implements OnInit, Af
...
 
   /**
    * Update the column width definitions when the rows change
    */
   public ngAfterViewChecked(): void {
+    this.viewChange.subscribe(v => {
+      this.initHeaderSubscriptions();
+    });
+
     this.rows.changes.subscribe(() => {
```
atlwendy, line 257: is this `this.rows.changes` subset of that `this.viewChange`? So any time rows changed would also trigger `viewChange`?

atlwendy (approved):
other than that minor format, it looks good.

terminus-ui/checkbox/src/checkbox.component.scss
benjamincharity: Yep. Checkbox is still 95% material. For some reason they change the cursor to `pointer` for the entire label wrapper even though the actual clickable section is much smaller. This fixes that slightly confusing issue.

terminus-ui/sort/src/sort-header.component.html
benjamincharity: Legacy material code. When sort was originally pulled into the library I didn't have time to fully refactor etc.

There is a `FIXME` note at the top of the file to refactor out the dangles and remove the lint rule for it... someday....

terminus-ui/table/src/cell.ts
benjamincharity: Yeah, this wasn't my first pass at it 😆 but really dig using RxJS for DOM events. My first pass was at least double the line count

terminus-ui/table/src/table.component.ts
benjamincharity: Argghhhh this gets me all the time. I even have started doing PR pre-reviews specifically for this. And _still_ missing it.. maybe if we get some slow cycles in December I can get around to writing that lint rule.

terminus-ui/table/src/table.component.ts
benjamincharity: This took me forever to get working and I'm still not 100% sure why...

When columns are reordered I lose my event bindings, so I need to re-initialize them. With the inspector open you can visually see the rows DOM get updated, but for some reason the `rows.changes` doesn't fire (when using a trackBy or not). I ended up finding `viewChange` by digging through the CDK source code and it worked. I believe it's a CDR issue with `rows.changes` but I'm honestly not entirely sure. I've seen a couple places now that Material/CDK uses an internal Subject to alert others that the internals have changed rather than relying only on Angular cycles.

terminus-ui/table/src/table.component.ts
benjamincharity: Fixed

PR #1760, opened 2019-10-25
1759 table icon

benjamincharity:
- Add support for new table editing icon
  - Slight code cleanup for custom icons
- Use new icon in table demo and docs


terminus-ui/icon/src/custom-icons/csv.ts
```diff
@@ -2,8 +2,33 @@
...
+        <path d="M10,0 L2,0 C0.9,0 0.01,0.9 0.01,2 L0,18 C0,19.1 0.89,20 1.99,20 L14,20 C15.1,20 16,19.1 16,18 L16,6 L10,0 Z M2,18 L2,2 L9,2
```
benjamincharity, line 4: FYI All of these internal SVG changes are just me formatting them to respect our max-line-length

terminus-ui/table/src/table.component.html
```diff
@@ -1,8 +0,0 @@
-<div class="ts-table__wrapper">
```
benjamincharity, line 1: This was missed in my previous PR. I plan on rebasing this into the correct commit once this PR is merged into v15

atlwendy: this block not needed? just making sure...

atlwendy (approved):

terminus-ui/table/src/table.component.html
benjamincharity: Yeah we used to need a custom wrapper so we defined our own template. Now we don't need that so we're just using the CDK table template. I had already changed the component to reference the CDK template, but forgot to remove the old template file.

benjamincharity: Yeah we used to need a custom wrapper so we defined our own template. Now we don't need that so we're just using the CDK table template. I had already changed the component to reference the CDK template, but forgot to remove the old template file.

PR #1763, opened 2019-10-25
1117 drawer

### LINKED ISSUE #1117: Drawer
We need a sidebar component for detail views in tables, unified nav, collapsable settings columns, etc.

### Container

- [x] Need a directive to define the container `ts-drawer-container`
- [x] Should have `position: relative`
- [x] Should add/remove a class that defines if a child drawer is open
  - `ts-drawer-container--expanded`
- [x] Should close the drawer if the user clicks the backdrop
  - [ ] Can disable auto-collapsing:
    - Input `disableAutoCollapse: boolean` default to `false`
- [x] Should support an optional backdrop to catch clicks outside of the drawer
  - [ ] Should be transparent
  - [ ] Input: `hasBackdrop = false` default to `false`
- [x] Multiple drawers can be added to the same side
  - [ ] Throw a library error when multiple drawers opening on the same side all with `push` mode
- [ ] Should emit an event anytime the backdrop is clicked (even if nothing should happen on click)
- [ ] When the drawer is in 'push' mode, the page content should be shrunk to fit the drawer (rather than pushing page content out of the viewport)

### Drawer

- [x]  Need a component `<ts-drawer>` - `TsDrawerComponent`
  - [ ] Export as `tsDrawer`
- [x]  Can set to overlay content or push content
  - [ ] Input: `mode: TsDrawerModes = 'overlay'`
  - `TsDrawerModes = 'overlay' | 'push'`
  - Default to `overlay`
- [x] Must be added to a `ts-drawer-container`
- [x]  Can be placed before or after content:
  - Input: `position: TsDrawerPositions`
  - `TsDrawerPositions: 'start' | 'end'`
  - [ ] Defaults to `start`
- [x]  Should be collapsable + expandable
- [ ]  Can contain rich content (HTML) via `ng-content`
- [ ]  Consumer can set the expanded and collapsed sizes
  - [ ]  We default to `3.75rem` and `12.5rem`
  - [ ] Inputs: `collapsedSize: string` & `expandedSize: string`
- [ ]  Should have public triggers to expand and collapse the panel.
  - [ ]  `expand()`
  - [ ]  `collapse()`
  - [ ]  `toggle()`
- [ ]  Should have correct aria `role` label (consumer can set based on their usage)
  - [ ]  Input `role`
  - [ ] Defaults to nothing
- [ ] Should have `aria-expanded` set to `false` when collapsed and `true` when expanded
- Should fire events
  - [ ] Should fire event when expanding starts: `expandedStart: void`
  - [ ] Should fire event when collapsing starts: `collapsedStart: void`
  - [ ] Should fire event when state change **ends** `expandedChange: 'expanded' | 'collapsed'` (ie when the animation ends)
- [ ]  Content should be vertically scrollable.
- [ ]  Should support being expanded on load
  - [ ]  Input `isExpanded: boolean`
  - [ ]  Defaults to `false`
- [ ]  Only run change detection when escape is hit - not every key:
  - See: [angular/components](https://github.com/angular/components/blob/master/src/material/sidenav/drawer.ts#L268-L279)
- [ ] Should use the same level shadow as a dropdown panel
  - Shadow should not be seen at top or bottom
- [ ] Should add/remove the classes that define the current UI state
  - `ts-drawer--expanded`
  - `ts-drawer--collapsed`
- [ ] Drawer should animate when expanding/collapsing
  - We should use Angular animations so that we can trigger events from animations.
  - Grab animations from Material: https://github.com/angular/components/blob/master/src/material/sidenav/drawer-animations.ts
- [ ] Add another item to our z-index map. The drawer should be above everything _except_ the global header. NOTE: We are adding this item to the map so that consumers can use it - but we aren't actually applying the z-index within the component.

#### Usage Example

```html
<ts-drawer-container>
  <ts-drawer #drawer1="tsDrawer">
    My 1st drawer content!

    <button (click)="drawer1.close()">Close</button>
  </ts-drawer>

  <ts-drawer-content>
    My page content!
  </ts-drawer-content>

  <ts-drawer #drawer2="tsDrawer">
    My 2nd drawer content!

    <button (click)="drawer2.close()">Close</button>
  </ts-drawer>
</ts-drawer-container>
```

---

<details>
  <summary>Examples in the wild</summary>

datadogdrawer

screen shot 2018-11-13 at 9 54 31 am
</details>

### Resources

- https://github.com/angular/components/blob/master/src/material/sidenav/drawer.ts
- https://material.angular.io/components/sidenav/overview

atlwendy:
closes #1117

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ The library of UI components used for Terminus applications.
 [![Github release][gh-release-badge]][gh-releases]
 [![Library size][file-size-badge]][raw-distribution-js]
 <br>
-[![components](https://img.shields.io/badge/Components-37-%2316a085.svg)](#components)
+[![components](https://img.shields.io/badge/Components-38-%2316a085.svg)](#components)
 [![pipes](https://img.shields.io/badge/Pipes-6-%2316a085.svg)](#pipes)
 [![services](https://img.shields.io/badge/Services-1-%2316a085.svg)](#services)
 [![style helpers](https://img.shields.io/badge/StyleHelpers-14-%2316a085.svg)](#style-helpers)
@@ -104,6 +104,7 @@ for each feature.
 | [date-range][src-date-range]               | Dual inputs with calendar pop-ups                         | :white_check_mark: | [Demo][demo-date-range]         | [Docs][date-range-docs]        | [Usage][date-range-usage]        |
 | dialog                                     |                                                           |                    |                                 |                                |                                  |
 | divider                                    |                                                           |                    |                                 |                                |                                  |
+| [drawer][src-drawer]                       | Drawer                                                    | :white_check_mark: | [Demo][demo-drawer]             | [Docs][drawer-docs]            | [Usage][drawer-usage]            |
 | [expansion-panel][src-expansion-panel]     | Expansion panel with accordion & stepper functionality    | :white_check_mark: | [Demo][demo-expansion-panel]    | [Docs][expansion-panel-docs]   | [Usage][expansion-panel-usage]   |
 | [file-upload][src-file-upload]             | File upload with drag and drop                            | :white_check_mark: | [Demo][demo-file-upload]        | [Docs][file-upload-docs]       | [Usage][file-upload-usage]       |
 | [icon][src-icon]                           | Supported icons: https://material.io/icons                | :white_check_mark: | [Demo][demo-icon]               | [Docs][icon-docs]              | [Usage][icon-usage]              |
@@ -356,6 +357,7 @@ This project follows the [all-contributors](https://github.com/kentcdodds/all-co
 [demo-csv-entry]:        https://getterminus.github.io/ui-demos-release/components/csv-entry
 [demo-date-range]:       https://getterminus.github.io/ui-demos-release/components/date-range
 [demo-datepicker]:       https://getterminus.github.io/ui-demos-release/components/datepicker
+[demo-drawer]:           https://getterminus.github.io/ui-demos-release/components/drawer
 [demo-expansion-panel]:  https://getterminus.github.io/ui-demos-release/components/expansion-panel
 [demo-file-upload]:      https://getterminus.github.io/ui-demos-release/components/file-upload
 [demo-icon-button]:      https://getterminus.github.io/ui-demos-release/components/icon-button
@@ -405,6 +407,7 @@ This project follows the [all-contributors](https://github.com/kentcdodds/all-co
 [src-copy]:               ./terminus-ui/copy/src/
 [src-csv-entry]:          ./terminus-ui/csv-entry/src/
 [src-date-range]:         ./terminus-ui/date-range/src/
+[src-drawer]:             ./terminus-ui/drawer/src/
 [src-expansion-panel]:    ./terminus-ui/expansion-panel/src/
 [src-file-upload]:        ./terminus-ui/file-upload/src/
 [src-icon-button]:        ./terminus-ui/icon-button/src/
@@ -451,6 +454,7 @@ This project follows the [all-contributors](https://github.com/kentcdodds/all-co
 [csv-entry-docs]:       http://uilibrary-docs.terminus.ninja/release/components/TsCSVEntryComponent.html
 [date-range-docs]:      http://uilibrary-docs.terminus.ninja/release/components/TsDateRangeComponent.html
 [datepicker-docs]:      http://uilibrary-docs.terminus.ninja/release/components/TsDatepickerComponent.html
+[drawer-docs]:          http://uilibrary-docs/terminus.ninja/release/components/TsDrawerComponent.html
 [expansion-panel-docs]: http://uilibrary-docs.terminus.ninja/release/components/TsExpansionPanelComponent.html
 [file-upload-docs]:     http://uilibrary-docs.terminus.ninja/release/components/TsFileUploadComponent.html
 [icon-button-docs]:     http://uilibrary-docs.terminus.ninja/release/components/TsIconButtonComponent.html
@@ -500,6 +504,7 @@ This project follows the [all-contributors](https://github.com/kentcdodds/all-co
 [csv-entry-usage]:         http://uilibrary-docs.terminus.ninja/release/components/TsCSVEntryComponent.html#readme
 [date-range-usage]:        http://uilibrary-docs.terminus.ninja/release/components/TsDateRangeComponent.html#readme
 [datepicker-usage]:        http://uilibrary-docs.terminus.ninja/release/components/TsDatepickerComponent.html#readme
+[drawer-usage]:            http://uilibrary-docs.terminus.ninja/release/components/TsDrawerComponent.html#readme
 [expansion-panel-usage]:   http://uilibrary-docs.terminus.ninja/release/components/TsExpansionPanelComponent.html#readme
 [file-upload-usage]:       http://uilibrary-docs.terminus.ninja/release/components/TsFileUploadComponent.html#readme
 [icon-button-usage]:       http://uilibrary-docs.terminus.ninja/release/components/TsIconButtonComponent.html#readme
--- a/demo/app/components/drawer/drawer.component.html
+++ b/demo/app/components/drawer/drawer.component.html
@@ -1,0 +1,101 @@
+  <div tsVerticalSpacing>
+    <ts-radio-group
+      label="Position:"
+      [options]="positions"
+      [formControl]="positionControl"
+      [formatUILabelFn]="uiFormatter"
+      [formatModelValueFn]="modelFormatter"
+    >
+    </ts-radio-group>
+  </div>
+
+  <ts-drawer-container
+    class="drawer-container"
+    [hasBackdrop]="false"
+    #drawerContainer
+  >
+    <ts-drawer
+      [mode]="modeControl.value"
+      [position]="positionControl.value"
+      [isExpanded]="false"
+      collapsedSize="2rem"
+      expandedSize="12rem"
+      class="first-drawer"
+      #drawer
+    >
+      <div tsVerticalSpacing class="content">
+        <p tsVerticalSpacing>
+          <ts-button
+            (click)="drawer.toggle()"
+          >First Drawer</ts-button>
+        </p>
+        <p>
+          <ts-radio-group
+            label="Mode:"
+            [options]="items"
+            [formControl]="modeControl"
+            [formatUILabelFn]="uiFormatter"
+            [formatModelValueFn]="modelFormatter"
+            class="radio"
+          >
+          </ts-radio-group>
+        </p>
+      </div>
+    </ts-drawer>
+
+    <ts-drawer
+      [mode]="mode2Control.value"
+      [position]="position2Control.value"
+      class="second-drawer"
+      collapsedSize="0rem"
+      #drawer2
+    >
+      <div tsVerticalSpacing>
+        <p tsVerticalSpacing>
+          <ts-button
+            (click)="drawer2.toggle()"
+          >Second Drawer</ts-button>
+        </p>
+        <p>
+          <ts-radio-group
+            label="Mode2:"
+            [options]="items"
+            [formControl]="mode2Control"
+            [formatUILabelFn]="uiFormatter"
+            [formatModelValueFn]="modelFormatter"
+          >
+          </ts-radio-group>
+        </p>
+      </div>
+    </ts-drawer>
+
+    <ts-drawer-content>
+      <div style="margin: 0px 100px;">
+        <div tsVerticalSpacing>
+          <ts-button
+            [theme]="theme"
+            [isDisabled]="false"
+            (clicked)="drawer.toggle()"
+          >Toggle first drawer</ts-button>
+        </div>
+
+        <div tsVerticalSpacing>
+          <ts-button
+            [theme]="theme"
+            [isDisabled]="false"
+            (clicked)="drawer2.toggle()"
+          >Toggle 2nd drawer</ts-button>
+        </div>
+
+        <div tsVerticalSpacing>
+          This is the drawer content...
+          The #1 software development tool used by agile teams.
+          Visual regression testing is the ability to cross-reference snapshots of a product, highlighting pixel differences — returning fast feedback on the current state of the screen.
+        </div>
+      </div>
+    </ts-drawer-content>
+  </ts-drawer-container>
+
+
+
+
--- a/demo/app/components/drawer/drawer.component.ts
+++ b/demo/app/components/drawer/drawer.component.ts
@@ -1,0 +1,73 @@
+import {
+  Component,
+  NgZone,
+  ViewChild
+} from '@angular/core';
+import {
+  FormBuilder,
+  FormControl,
+  FormGroup,
+  Validators,
+} from '@angular/forms';
+import {
+  TsDrawerContainerComponent,
+} from '@terminus/ui/drawer';
+import { TsRadioFormatFn } from '@terminus/ui/radio-group';
+import { TsStyleThemeTypes } from '@terminus/ui/utilities';
+
+@Component({
+  selector: 'demo-drawer',
+  templateUrl: './drawer.component.html',
+  styleUrls: ['./drawer.component.scss'],
+})
+export class DrawerComponent {
+  public myForm: FormGroup = this.formBuilder.group({
+    isSmall: false,
+    isCentered: true,
+    myRadioGroup: [
+      null,
+      [
+        Validators.required,
+      ],
+    ],
+    myRadioGroup2: [
+      'bar2_value',
+      [
+        Validators.required,
+      ],
+    ],
+  });
+  public modeControl = new FormControl('push');
+  public mode2Control = new FormControl('overlay');
+  public items = [{
+    display: 'Overlay the Content',
+    value: 'overlay',
+  }, {
+    display: 'Push Over the Content',
+    value: 'push',
+  }];
+
+  public positionControl = new FormControl('start');
+  public position2Control = new FormControl('end');
+  public positions = [{
+    display: 'First drawer opens on the left',
+    value: 'start',
+  }, {
+    display: 'First drawer opens on the right',
+    value: 'end',
+  }];
+
+  public uiFormatter: TsRadioFormatFn = v => v.display;
+  public modelFormatter: TsRadioFormatFn = v => v.value;
+
+  public theme: TsStyleThemeTypes = 'warn';
+  public position = 'end';
+
+  @ViewChild('drawerContainer', { static: false })
+  public drawerContainer!: TsDrawerContainerComponent;
+
+  constructor(
+    private formBuilder: FormBuilder,
+    public zone: NgZone,
+  ) { }
+}
--- a/demo/app/components/drawer/drawer.module.ts
+++ b/demo/app/components/drawer/drawer.module.ts
@@ -1,0 +1,33 @@
+import { CommonModule } from '@angular/common';
+import { NgModule } from '@angular/core';
+import {
+  FormsModule,
+  ReactiveFormsModule,
+} from '@angular/forms';
+import { TsCardModule } from '@terminus/ui/card';
+import { TsSpacingModule } from '@terminus/ui/spacing';
+
+import { TsButtonModule } from '@terminus/ui/button';
+import { TsDrawerModule } from '@terminus/ui/drawer';
+import { TsRadioGroupModule } from '@terminus/ui/radio-group';
+import { DrawerRoutingModule } from './drawer-routing.module';
+import { DrawerComponent } from './drawer.component';
+
+
+@NgModule({
+  imports: [
+    CommonModule,
+    DrawerRoutingModule,
+    FormsModule,
+    ReactiveFormsModule,
+    TsCardModule,
+    TsRadioGroupModule,
+    TsSpacingModule,
+    TsDrawerModule,
+    TsButtonModule,
+  ],
+  declarations: [
+    DrawerComponent,
+  ],
+})
+export class DrawerModule {}
--- a/terminus-ui/drawer/src/drawer-animations.ts
+++ b/terminus-ui/drawer/src/drawer-animations.ts
@@ -1,0 +1,32 @@
+import {
+  animate,
+  AnimationTriggerMetadata,
+  state,
+  style,
+  transition,
+  trigger,
+} from '@angular/animations';
+
+/**
+ * Animations used by the {@link TsDrawerComponent}.
+ */
+export const tsDrawerAnimations: {
+  readonly transformDrawer: AnimationTriggerMetadata;
+} = {
+  // Animation that expands and collapses a drawer.
+  transformDrawer: trigger('transform', [
+    state('open, open-instant', style({
+      transform: 'none',
+      visibility: 'visible',
+      width: '{{ expandedSize }}',
+    }), { params: { expandedSize: '12.5rem' } }),
+    state('void', style({
+      'box-shadow': 'none',
+      'visibility': 'visible',
+      'width': '{{ collapsedSize }}',
+    }), { params: { collapsedSize: '3.75rem' } }),
+    transition('void => open-instant', animate('0ms')),
+    transition('void <=> open, open-instant => void, left-close <=> open, right-close <=> open',
+      animate('400ms cubic-bezier(0.25, 0.8, 0.25, 1)')),
+  ]),
+};
--- a/terminus-ui/drawer/src/drawer-container.component.html
+++ b/terminus-ui/drawer/src/drawer-container.component.html
@@ -1,0 +1,11 @@
+<div
+  class="ts-drawer__backdrop"
+  (click)="onBackdropClicked()"
+  *ngIf="hasBackdrop"
+  [class.ts-drawer__backdrop--visible]="isShowingBackdrop"
+></div>
+
+<ng-content select="ts-drawer"></ng-content>
+
+<ng-content select="ts-drawer-content">
+</ng-content>
--- a/terminus-ui/drawer/src/drawer-container.component.spec.ts
+++ b/terminus-ui/drawer/src/drawer-container.component.spec.ts
@@ -1,0 +1,220 @@
+import { PlatformModule } from '@angular/cdk/platform';
+import { ScrollingModule } from '@angular/cdk/scrolling';
+import { CommonModule } from '@angular/common';
+import { Type } from '@angular/core';
+import {
+  async,
+  ComponentFixture,
+  fakeAsync,
+  flush,
+  tick,
+} from '@angular/core/testing';
+import {
+  FormsModule,
+  ReactiveFormsModule,
+} from '@angular/forms';
+import { MatRippleModule } from '@angular/material/core';
+import { By } from '@angular/platform-browser';
+import { BrowserAnimationsModule } from '@angular/platform-browser/animations';
+import { createComponent as createComponentInner } from '@terminus/ngx-tools/testing';
+import * as testComponents from '@terminus/ui/drawer/testing';
+
+import { TsDrawerComponent } from './drawer.component';
+import { TsDrawerModule } from './drawer.module';
+
+function createComponent<T>(component: Type<T>): ComponentFixture<T> {
+  const moduleImports = [
+    BrowserAnimationsModule,
+    CommonModule,
+    FormsModule,
+    ReactiveFormsModule,
+    MatRippleModule,
+    ScrollingModule,
+    PlatformModule,
+    TsDrawerModule,
+  ];
+
+  return createComponentInner(component, undefined, moduleImports);
+}
+
+describe(`TsDrawerContainerComponent`, () => {
+  let fixture;
+  let component;
+  let drawers;
+
+  function setUpTestComponent(testComponent) {
+    fixture = createComponent(testComponent);
+    fixture.detectChanges();
+    component = fixture.debugElement.queryAll(By.css('ts-drawer-container'))[0].componentInstance;
+    drawers = component.drawers;
+  }
+
+  test(`should exist`, () => {
+    setUpTestComponent(testComponents.RegularDrawer);
+    expect(fixture.debugElement.query(By.css('.ts-drawer-container'))).toBeTruthy();
+  });
+
+  test('should be able to open and close all drawers', fakeAsync(() => {
+    setUpTestComponent(testComponents.RegularDrawer);
+    const allDrawers = fixture.debugElement.queryAll(By.directive(TsDrawerComponent));
+
+    expect(allDrawers.every(drawer => drawer.componentInstance.isExpanded)).toBeFalsy();
+
+    component.expand();
+    fixture.detectChanges();
+    tick();
+
+    expect(allDrawers.every(drawer => drawer.componentInstance.isExpanded)).toBe(true);
+
+    component.collapse();
+    fixture.detectChanges();
+    flush();
+
+    expect(allDrawers.every(drawer => drawer.componentInstance.isExpanded)).toBeFalsy();
+  }));
+
+  describe(`should set size accordingly`, () => {
+
+    test('should calculate the left margin', fakeAsync(() => {
+      setUpTestComponent(testComponents.SetMargins);
+      const drawerElement = fixture.debugElement.nativeElement.querySelector('.ts-drawer');
+
+      fixture.detectChanges();
+      Object.defineProperty(drawerElement, 'offsetWidth', {
+        configurable: true,
+        value: 500,
+      });
+      fixture.componentInstance.drawer.expand();
+      fixture.detectChanges();
+      tick();
+      expect(component.contentMargins.left).toEqual(204);
+
+    }));
+
+    test(`should calculate the right margin`, fakeAsync(() => {
+      setUpTestComponent(testComponents.SetMargins);
+      const drawerElement = fixture.debugElement.nativeElement.querySelector('.ts-drawer');
+
+      fixture.debugElement.componentInstance.position = 'end';
+      fixture.detectChanges();
+      Object.defineProperty(drawerElement, 'offsetWidth', {
+        configurable: true,
+        value: 500,
+      });
+      fixture.componentInstance.drawer.expand();
+      fixture.detectChanges();
+      tick();
+      expect(component.contentMargins.right).toEqual(204);
+    }));
+
+    test(`should have no right margin in overlay mode expanding on end position`, () => {
+      setUpTestComponent(testComponents.MultipleDrawer);
+      const drawer2 = component.drawers.last;
+      const contentElement = fixture.debugElement.nativeElement.querySelector('.ts-drawer-content');
+      const content = component.content;
+      Object.defineProperty(contentElement, 'rightMargin', {
+        configurable: true,
+        value: 20,
+      });
+      Object.defineProperty(content, 'rightMargin', {
+        configurable: true,
+        value: 20,
+      });
+
+      drawer2.expand();
+      fixture.detectChanges();
+
+      expect(component.contentMargins.right).toBeFalsy();
+    });
+
+    test(`should not move the margin if 2nd drawer is overlay on top of first drawer`, async(() => {
+      setUpTestComponent(testComponents.MultipleDrawer);
+      const drawer1 = component.drawers.first;
+      const drawer2 = component.drawers.last;
+      const contentElement = fixture.debugElement.nativeElement.querySelector('.ts-drawer-content');
+      const content = component.content;
+      Object.defineProperty(contentElement, 'rightMargin', {
+        configurable: true,
+        value: 20,
+      });
+      Object.defineProperty(content, 'rightMargin', {
+        configurable: true,
+        value: 20,
+      });
+      Object.defineProperty(component, 'contentMargins', {
+        configurable: true,
+        value: {
+          left: 20,
+          right: 20,
+        },
+      });
+      drawer1.expand();
+      fixture.detectChanges();
+
+      drawer2.expand();
+      fixture.detectChanges();
+      // When second drawer with overlay mode, the content margin should not change
+      expect(component.contentMargins.right).toEqual(20);
+    }));
+  });
+
+  describe(`backdrop`, () => {
+    beforeEach(() => {
+      setUpTestComponent(testComponents.BasicDrawer);
+    });
+
+    test('should be able to toggle whether the container has a backdrop', fakeAsync(() => {
+      expect(fixture.nativeElement.querySelector('.ts-drawer__backdrop')).toBeTruthy();
+
+      fixture.componentInstance.hasBackdrop = false;
+      fixture.detectChanges();
+
+      expect(fixture.nativeElement.querySelector('.ts-drawer__backdrop')).toBeFalsy();
+    }));
+
+    test(`should dismiss the opening drawer if backdrop is enabled and click on container`, async(() => {
+      const backdrop = document.querySelector('.ts-drawer__backdrop')! as HTMLElement;
+      expect(fixture.debugElement.componentInstance.drawer.isExpanded).toBeFalsy();
+      component.expand();
+      fixture.detectChanges();
+      fixture.whenStable().then(() => {
+        backdrop.click();
+        fixture.detectChanges();
+        expect(fixture.debugElement.componentInstance.drawer.isExpanded).toBeFalsy();
+      });
+    }));
+  });
+
+  describe(`open on load`, () => {
+    test(`should not animated`, async(() => {
+      setUpTestComponent(testComponents.OpenOnLoadDrawer);
+      fixture.detectChanges();
+      fixture.whenStable().then(() => {
+        const container = fixture.debugElement.nativeElement.querySelector('.ts-drawer-container');
+        expect(container.classList).not.toContain('ts-drawer-transition');
+      });
+    }));
+  });
+
+  test('should throw when multiple drawers have the same position', () => {
+    setUpTestComponent(testComponents.MultiDrawerSameSidePush);
+    fixture.detectChanges();
+    const setup = () => {
+      component.validateDrawers();
+    };
+    expect(setup).toThrowError();
+  });
+
+  test(`should readjust drawers when mode changed`, () => {
+    setUpTestComponent(testComponents.SetSizeDrawer);
+    component.validateDrawers = jest.fn();
+    component.updateContentMargins = jest.fn();
+    component.changeDetectorRef.markForCheck = jest.fn();
+    fixture.debugElement.componentInstance.mode = 'push';
+    fixture.detectChanges();
+    expect(component.validateDrawers).toHaveBeenCalled();
+    expect(component.updateContentMargins).toHaveBeenCalled();
+    expect(component.changeDetectorRef.markForCheck).toHaveBeenCalled();
+  });
+});
+
--- a/terminus-ui/drawer/src/drawer-container.component.ts
+++ b/terminus-ui/drawer/src/drawer-container.component.ts
@@ -1,0 +1,430 @@
+// tslint:disable: template-no-call-expression
+import {
+  AfterContentInit,
+  ChangeDetectionStrategy,
+  ChangeDetectorRef,
+  Component,
+  ContentChild,
+  ContentChildren,
+  DoCheck,
+  ElementRef,
+  EventEmitter,
+  Inject,
+  Input,
+  NgZone,
+  OnDestroy,
+  Optional,
+  Output,
+  QueryList,
+  Renderer2,
+  ViewEncapsulation,
+} from '@angular/core';
+import { ANIMATION_MODULE_TYPE } from '@angular/platform-browser/animations';
+import { untilComponentDestroyed } from '@terminus/ngx-tools';
+import { TsUILibraryError } from '@terminus/ui/utilities';
+import { Subject } from 'rxjs';
+import {
+  debounceTime,
+  filter,
+  startWith,
+  take,
+  takeUntil,
+} from 'rxjs/operators';
+
+import { TsDrawerContentComponent } from './drawer-content.component';
+import { TsDrawerComponent } from './drawer.component';
+
+/**
+ * Throws an exception when two TsDrawer are matching the same position with same mode.
+ *
+ * @params - drawer's position
+ */
+export function throwTsDuplicatedDrawerError(position: string): void {
+  throw new TsUILibraryError(`TsDrawerContainer: Only one drawer can be in push mode on '"${position}" position'.`);
+}
+
+/**
+ * Define content left right margin interface
+ */
+export interface ContentLeftRightMargin {
+  left: number | null;
+  right: number | null;
+}
+
+/**
+ * This is used to convert margin measurement from rem to px.
+ */
+const MARGIN_SIZE_CONVERSION = 16;
+
+/**
+ * The drawer container UI Component
+ *
+ * @example
+ * <ts-drawer-container
+ *              [hasBackdrop]="hasBackdrop"
+ *              (backdropClick)="backdropClicked($event)"
+ * ></ts-drawer-container>
+ *
+ * <example-url>https://getterminus.github.io/ui-demos-release/components/drawer</example-url>
+ */
+@Component({
+  selector: 'ts-drawer-container',
+  templateUrl: './drawer-container.component.html',
+  styleUrls: ['./drawer.component.scss'],
+  host: {
+    'class': 'ts-drawer-container',
+    '[class.ts-drawer-container-explicit-backdrop]': 'hasBackdrop',
+  },
+  changeDetection: ChangeDetectionStrategy.OnPush,
+  encapsulation: ViewEncapsulation.None,
+  exportAs: 'tsDrawerContainer',
+})
+export class TsDrawerContainerComponent implements AfterContentInit, DoCheck, OnDestroy {
+
+  /**
+   * The drawer at the start/end position.
+   */
+  private start: TsDrawerComponent | null = null;
+  private end: TsDrawerComponent | null = null;
+  private openDrawerCount = 0;
+
+  /**
+   * Emits on every ngDoCheck. Used for debouncing reflows.
+   */
+  private readonly doCheckSubject = new Subject<void>();
+
+  /**
+   * Whether the container is being pushed to the side by one of the drawers.
+   */
+  private get isPushed(): boolean {
+    return (TsDrawerContainerComponent.isDrawerOpen(this.start) && this.start.mode === 'push')
+      || (TsDrawerContainerComponent.isDrawerOpen(this.end) && this.end.mode === 'push');
+  }
+
+  /**
+   * Define whether backdrop is shown.
+   */
+  public get isShowingBackdrop(): boolean {
+    return (TsDrawerContainerComponent.isDrawerOpen(this.start) && this.hasBackdrop)
+      || (TsDrawerContainerComponent.isDrawerOpen(this.end) && this.hasBackdrop);
+  }
+
+  /**
+   * Access the child drawer {@link TsDrawerComponent}
+   */
+  @ContentChildren(TsDrawerComponent)
+  public drawers!: QueryList<TsDrawerComponent>;
+
+  /**
+   * Access the child content {@link TsDrawerContentComponent}
+   */
+  @ContentChild(TsDrawerContentComponent, { static: false })
+  public content!: TsDrawerContentComponent;
+
+  /**
+   * Margins to be applied to the content. These are used to push / shrink the drawer content when a
+   * drawer is open. We use margin rather than transform even for push mode because transform breaks
+   * fixed position elements inside of the transformed element.
+   */
+  public contentMargins: ContentLeftRightMargin = {
+    left: null,
+    right: null,
+  };
+
+  /**
+   * Used to notify anyone that's listening to content margin changes.
+   */
+  public readonly contentMarginChanges = new Subject<ContentLeftRightMargin>();
+
+  /**
+   * Whether the drawer container should have a backdrop while one of the drawers is open.
+   * If explicitly set to `true`, the backdrop will be enabled for drawers.
+   */
+  @Input()
+  public set hasBackdrop(value: boolean) {
+    this._hasBackdrop = value;
+  }
+  public get hasBackdrop(): boolean {
+    return this._hasBackdrop;
+  }
+  public _hasBackdrop = false;
+
+  /**
+   * Event emitted when the drawer backdrop is clicked.
+   */
+  @Output()
+  private readonly backdropClick = new EventEmitter<void>();
+
+  constructor(
+    private element: ElementRef<HTMLElement>,
+    private ngZone: NgZone,
+    public renderer: Renderer2,
+    private changeDetectorRef: ChangeDetectorRef,
+    @Optional() @Inject(ANIMATION_MODULE_TYPE) private animationMode?: string,
+  ) {
+  }
+
+  /**
+   * After drawer initiated, subscribe to drawer and content changes.
+   */
+  public ngAfterContentInit(): void {
+    const contentEle = this.content ? this.content.getElementRef().nativeElement : null;
+    this.drawers.changes.pipe(
+      startWith<void, null>(null),
+      untilComponentDestroyed(this),
+    ).subscribe(() => {
+      this.validateDrawers();
+      this.drawers.forEach((drawer: TsDrawerComponent) => {
+        this.adjustDrawerOnChange(drawer);
+        this.drawerPositionOnChange(drawer);
+        this.drawerModeOnChange(drawer);
+      });
+
+      if (!this.drawers.length
+        || TsDrawerContainerComponent.isDrawerOpen(this.start)
+        || TsDrawerContainerComponent.isDrawerOpen(this.end)) {
+        this.updateContentMargins();
+      }
+      this.changeDetectorRef.markForCheck();
+    });
+
+    // Arbitrary debounce time, less than a frame at 60fps
+    this.doCheckSubject.pipe(
+      debounceTime(1),
+      untilComponentDestroyed(this)
+    ).subscribe(() => this.updateContentMargins());
+
+    this.contentMarginChanges.subscribe(e => {
+      if (this.content) {
+        this.content.leftMargin = e.left;
+        this.content.rightMargin = e.right;
+        this.renderer.setStyle(contentEle, 'margin-left', e.left ? `${e.left.toString()}px` : 0);
+        this.renderer.setStyle(contentEle, 'margin-right', e.right ? `${e.right.toString()}px` : 0);
+        this.changeDetectorRef.markForCheck();
+      }
+    });
+  }
+
+  /**
+   * Run outside the NgZone, otherwise the debouncer will throw us into an infinite loop.
+   */
+  public ngDoCheck(): void {
+    if (this.isPushed) {
+      this.ngZone.runOutsideAngular(() => this.doCheckSubject.next());
+    }
+  }
+
+  /**
+   * Complete the observable on destroy
+   */
+  public ngOnDestroy(): void {
+    this.contentMarginChanges.complete();
+    this.doCheckSubject.complete();
+  }
+
+  /**
+   * Calls `expand` of drawers
+   */
+  public expand(): void {
+    this.drawers.forEach(drawer => drawer.expand());
+  }
+
+  /**
+   * Calls `collapse` of drawers
+   */
+  public collapse(): void {
+    this.drawers.forEach(drawer => drawer.collapse());
+  }
+
+  /**
+   * Recalculates and updates the inline styles for the content.
+   */
+  public updateContentMargins(): void {
+    let left: number | null = 0;
+    let right: number | null = 0;
+    left = this.start ? this.marginCalculation(this.start, 'left', 0) : left;
+    right = this.end ? this.marginCalculation(this.end, 'right', 0) : right;
+
+    // If either `right` or `left` is zero, don't set a style to the element. This
+    // allows users to specify a custom size via CSS class in SSR scenarios where the
+    // measured widths will always be zero. Note that we reset to `null` here, rather
+    // than below, in order to ensure that the types in the `if` below are consistent.
+    left = left || null;
+    right = right || null;
+    if (left !== this.contentMargins.left || right !== this.contentMargins.right) {
+      this.contentMargins = {
+        left,
+        right,
+      };
+
+      // Pull back into the NgZone since in some cases we could be outside.
+      this.ngZone.run(() => this.contentMarginChanges.next(this.contentMargins));
+    }
+  }
+
+  /**
+   * For drawers in `overlay` mode, they don't affect the content.
+   * For drawers in `push` mode they should shrink the content. We do this by adding to the
+   * left margin (for left drawer) or right margin (for right the drawer).
+   * If the content margin has already been set and there are more than 1 opening windows on the same size, do not update.
+   *
+   * @param drawer: drawer component,
+   * @param side: left of right side where the drawer expands
+   * @param margin: margin on that side of drawer
+   * @return - calculated margin on the specific side.
+   */
+  private marginCalculation(drawer: TsDrawerComponent | null, side: string, margin: number): number {
+    let sideMargin;
+    if (this.content) {
+      sideMargin = side === 'left' ? this.content.leftMargin : this.content.rightMargin;
+    }
+    if (drawer && drawer.isExpanded) {
+      if (drawer.mode === 'push') {
+        const width = parseFloat(drawer.expandedSize.replace(/[^0-9.]/g, ''));
+        const measurement = drawer.expandedSize.match(/[a-zA-Z]+/g) || '';
+        const coefficient = measurement.includes('px') ? 1 : MARGIN_SIZE_CONVERSION;
+        margin += width * coefficient;
+      } else if (this.content && sideMargin && this.openDrawerCount <= 1) {
+        margin = 0;
+      } else if (this.content && sideMargin && this.openDrawerCount > 1) {
+        margin = this.contentMargins[side] || 0;
+      }
+    }
+    return margin;
+  }
+
+  /**
+   * Subscribes to drawer events in order to set a class on the main container element when the
+   * drawer is open and the backdrop is visible. This ensures any overflow on the container element
+   * is properly hidden.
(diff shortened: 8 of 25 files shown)


.gitignore
```diff
@@ -1,3 +1,4 @@
+junit.xml
```
benjamincharity: This file shouldn't be in here. If you rebase release to get the latest, this file will be generated in the correct location and won't need to be ignored.

terminus-ui/drawer/src/drawer-container.component.ts
```diff
@@ -0,0 +1,345 @@
...
+
+  public readonly contentMarginChanges = new Subject<{left: number|null; right: number|null}>();
+
+  /**
+   * Whether the drawer container should have a backdrop while one of the drawers is open.
+   * If explicitly set to `true`, the backdrop will be enabled for drawers.
+   */
+  @Input()
+  public set hasBackdrop(value: boolean) {
+    this._hasBackdrop = value ? coerceBooleanProperty(value) : false;
```
benjamincharity: We don't need to coerce our boolean inputs anymore.

terminus-ui/drawer/src/drawer-container.component.ts
```diff
@@ -0,0 +1,345 @@
...
+  }
+  public _hasBackdrop = false;
+
+  /**
+   * Event emitted when the drawer backdrop is clicked.
+   */
+  @Output()
+  private readonly backdropClick: EventEmitter<void> = new EventEmitter<void>();
+
+  constructor(private element: ElementRef<HTMLElement>,
```
benjamincharity: We should format this like our other constructors with a break after opening and before closing

terminus-ui/drawer/src/drawer.component.ts
```diff
@@ -0,0 +1,343 @@
...
+  styleUrls: ['./drawer.component.scss'],
+  animations: [tsDrawerAnimations.transformDrawer],
+  host: {
+    'class': 'ts-drawer',
+    // must prevent the browser from aligning text based on value
+    '[attr.align]': 'null',
+    '[attr.role]': 'role',
+    '[class.ts-drawer-end]': 'position === "end"',
+    '[class.ts-drawer-overlay]': 'mode === "overlay"',
+    '[class.ts-drawer-push]': 'mode === "push"',
```
benjamincharity: We should use our BEM format here eg `ts-drawer--end`

terminus-ui/drawer/src/drawer.component.ts
```diff
@@ -0,0 +1,343 @@
...
+ * This component corresponds to a drawer that can be opened on the drawer container.
+ */
+@Component({
+  selector: 'ts-drawer',
+  templateUrl: './drawer.component.html',
+  styleUrls: ['./drawer.component.scss'],
+  animations: [tsDrawerAnimations.transformDrawer],
+  host: {
+    'class': 'ts-drawer',
+    // must prevent the browser from aligning text based on value
```
benjamincharity: Can we clarify more on the why vs what? Something like `NOTE: To avoid ____ we must ___` etc

terminus-ui/drawer/src/drawer.component.ts
```diff
@@ -0,0 +1,343 @@
...
+    factory: TS_DRAWER_DEFAULT_AUTOSIZE_FACTORY,
+  });
+
+export function TS_DRAWER_DEFAULT_AUTOSIZE_FACTORY(): boolean {
+  return false;
+}
+
+
+/**
+ * This component corresponds to a drawer that can be opened on the drawer container.
```
benjamincharity: When referencing another class/component we should use the JSDoc link so that our docs are generated with a cross link:

`This component corresponds to a drawer that is nested inside a {@link TsDrawerContainerComponent}`

terminus-ui/drawer/src/drawer-container.component.ts
```diff
@@ -0,0 +1,345 @@
...
+import {
+  TS_DRAWER_DEFAULT_AUTOSIZE,
+  TsDrawerComponent,
+} from './drawer.component';
+
+/**
+ * Throws an exception when two TsDrawer are matching the same position with same mode.
+ */
+export function throwTsDuplicatedDrawerError(position: string) {
+  throw new TsUILibraryError(`A drawer was already declared for 'position="${position}"' with push mode.`);
```
benjamincharity: Two comments:

a) I've been prefixing error messages with the name of the component that throws it. eg `TsDrawerContainer: My message`

b) We should probably reword this message to be a bit more about what is wrong vs what is set (that sentence is not clear so I'll add an example). eg `TsDrawerContainer: Only 1 drawer can be registered in 'push' mode to a single side of a container.` etc

demo/app/components/drawer/drawer.component.html
```diff
@@ -0,0 +1,102 @@
...
+    <ts-radio-group
+      label="Position:"
+      [options]="positions"
+      [formControl]="positionControl"
+      [formatUILabelFn]="uiFormatter"
+      [formatModelValueFn]="modelFormatter"
+    >
+    </ts-radio-group>
+  </div>
+
```
benjamincharity: I'd love to make it more clear where the drawer is registered. Possibly moving the demo controls out of the card? Or maybe into it's own card.

Then try laying out the demo drawers without using absolute positioning? Just trying to make it as close to how the consumer will likely use it in app.

demo/app/components/drawer/drawer.component.html
```diff
@@ -0,0 +1,102 @@
...
+            [formatUILabelFn]="uiFormatter"
+            [formatModelValueFn]="modelFormatter"
+            class="radio"
+          >
+          </ts-radio-group>
+        </p>
+      </div>
+    </ts-drawer>
+
+    <ts-drawer
```
benjamincharity, line 46: Thoughts on possibly moving this drawer to the right side? I feel like it may show more options. Plus it'll make it a bit more clear between the two drawers since they are styled the same.

demo/app/components/drawer/drawer.component.html
```diff
@@ -0,0 +1,102 @@
...
+      [mode]="mode2Control.value"
+      [position]="position2Control.value"
+      class="second-drawer"
+      collapsedSize="0rem"
```
benjamincharity, line 50: Assuming we can remove this as `0` is default?

demo/app/components/drawer/drawer.component.html
```diff
@@ -0,0 +1,102 @@
...
+    #drawerContainer
+  >
+    <ts-drawer
+      [mode]="modeControl.value"
+      [position]="positionControl.value"
+      isExpanded="false"
+      collapsedSize="2rem"
+      expandedSize="12rem"
+      class="first-drawer"
+      #drawer
```
benjamincharity, line 24: Also, maybe we default the left drawer open?

terminus-ui/drawer/src/drawer-animations.ts
```diff
@@ -0,0 +1,32 @@
...
+  trigger,
+} from '@angular/animations';
+
+/**
+ * Animations used by the Drawer component.
+ */
+export const tsDrawerAnimations: {
+  readonly transformDrawer: AnimationTriggerMetadata;
+} = {
+  // Animation that slides a drawer in and out.
```
benjamincharity: Wording: slides vs expand/collapse

terminus-ui/drawer/src/drawer-animations.ts
```diff
@@ -0,0 +1,32 @@
...
+  animate,
+  AnimationTriggerMetadata,
+  state,
+  style,
+  transition,
+  trigger,
+} from '@angular/animations';
+
+/**
+ * Animations used by the Drawer component.
```
benjamincharity: Let's use the JSDoc link here too: `Animations used by the {@link TsDrawerComponent}.`

terminus-ui/drawer/src/drawer-container.component.html
```diff
@@ -0,0 +1,15 @@
+<div
+  class="ts-drawer-backdrop"
+  (click)="onBackdropClicked()"
+  *ngIf="hasBackdrop"
+  [class.ts-drawer-shown]="isShowingBackdrop()"
```
benjamincharity: BEM style: 

1. `ts-drawer__backdrop`
2. `ts-drawer__backdrop--visible`

terminus-ui/drawer/src/drawer-container.component.html
```diff
@@ -0,0 +1,15 @@
...
+></div>
+
+<ng-content select="ts-drawer"></ng-content>
+
+<ng-content select="ts-drawer-content">
```
benjamincharity, line 10: BEM style: `ts-drawer__content`

terminus-ui/drawer/src/drawer.component.scss
```diff
@@ -0,0 +1,115 @@
...
+
+// Mixin that creates a new stacking context.
+@mixin ts-drawer-stacking-context($z-index: 1) {
+  position: relative;
+  // Use a z-index to create a new stacking context. (We can't use transform because it breaks fixed
+  // positioning inside of the transformed element).
+  z-index: $z-index;
+}
+
+@mixin ts-fill {
```
benjamincharity: We've got this mixin already: https://github.com/GetTerminus/terminus-ui/blob/release/terminus-ui/scss/docs/layout.md#take-space

terminus-ui/drawer/src/drawer.component.scss
```diff
@@ -0,0 +1,115 @@
...
+      timing-function: cubic-bezier($g-material-background-easing);
+    }
+  }
+}
+
+.ts-drawer {
+  $high-contrast-border: solid 1px currentColor;
+
+  background-color: color(utility, light);
+
```
benjamincharity: Your linter should be yelling about `@include` not coming before `background-color`.

We also should need a line break between the last style and the opening of the next style block (ie after `z-index`)

demo/app/components/drawer/drawer.component.html
benjamincharity: Seeing now that absolute positioning was needed to get around the card padding. Maybe we don't use a card for the drawers? Just create a random div with an outline so it's clear what's happening?

demo/app/components/drawer/drawer.component.ts
```diff
@@ -0,0 +1,90 @@
...
+    //     thisdrawer.open();
+    //   });
+    // }
+  }
+
+  ngAfterViewChecked() {
+    // this.zone.runOutsideAngular(() => {
+    //   this.drawerContainer.open();
+    // });
+  }
```
benjamincharity: Remove legacy commented code here and in the constructor.

Also, can we add a few methods here to log out all drawer events (only need for one drawer - not both). Log format: `DEMO: what happened: EMITTER_PAYLOAD`

terminus-ui/drawer/src/drawer-container.component.html
```diff
@@ -0,0 +1,15 @@
...
+</ng-content>
+
+<ts-drawer-content *ngIf="!content">
```
benjamincharity: Do we need/want this? It seems like we'd want to force them to use our content container vs doing it for them?

demo/app/components/drawer/drawer.module.ts
```diff
@@ -0,0 +1,37 @@
...
+    DrawerRoutingModule,
+    FormsModule,
+    ReactiveFormsModule,
+    TsCardModule,
+    TsRadioGroupModule,
+    TsSpacingModule,
+    TsRadioGroupModule,
+    TsDrawerModule,
+    TsButtonModule,
+    TsDrawerModule,
```
benjamincharity: Duplicate import

benjamincharity: Same for `TsRadioGroupModule`

demo/app/components/drawer/drawer.module.ts
```diff
@@ -0,0 +1,37 @@
+import { CommonModule } from '@angular/common';
+import { NgModule } from '@angular/core';
+import {
+  FormsModule,
+  ReactiveFormsModule,
+} from '@angular/forms';
+import { TsCardModule } from '@terminus/ui/card';
+import { TsCohortDateRangeModule } from '@terminus/ui/cohort-date-range';
+import { TsDateRangeModule } from "@terminus/ui/date-range";
```
benjamincharity: I don't think `TsDateRangeModule` or `TsCohortDateRangeModule` is used?

terminus-ui/drawer/src/drawer-container.component.ts
```diff
@@ -0,0 +1,345 @@
...
+  }
+
+  public closeModalDrawer() {
+    // Close all open drawers where closing is not disabled.
+    [this.start, this.end]
+      .filter(drawer => drawer && this.hasBackdrop)
+      .forEach(drawer => drawer && drawer.collapse());
+  }
+
+  public isShowingBackdrop(): boolean {
```
benjamincharity: I think this would work great as a getter 👍

terminus-ui/drawer/src/drawer-container.component.ts
```diff
@@ -0,0 +1,345 @@
...
+    });
+    if (hasDuplicates) {
+      throwTsDuplicatedDrawerError(seen.values().next().value);
+    }
+
+    this.changeDetectorRef.markForCheck();
+  }
+
+  // Whether the container is being pushed to the side by one of the drawers.
+  private isPushed() {
```
benjamincharity: I think this would work great as a getter 👍

terminus-ui/drawer/src/drawer-container.component.ts
```diff
@@ -0,0 +1,345 @@
...
+  },
+  changeDetection: ChangeDetectionStrategy.OnPush,
+  encapsulation: ViewEncapsulation.None,
+  exportAs: 'tsDrawerContainer',
+})
+export class TsDrawerContainerComponent implements AfterContentInit, DoCheck, OnDestroy {
+
+  // The drawer at the start/end position.
+  private start!: TsDrawerComponent | null;
+  private end!: TsDrawerComponent | null;
```
benjamincharity: Seems like we could remove the non-null-assertions here and default to `null`?

terminus-ui/drawer/src/drawer-container.component.ts
```diff
@@ -0,0 +1,345 @@
...
+  throw new TsUILibraryError(`A drawer was already declared for 'position="${position}"' with push mode.`);
+}
+
+@Component({
+  selector: 'ts-drawer-container',
+  templateUrl: './drawer-container.component.html',
+  styleUrls: ['./drawer.component.scss'],
+  host: {
+    'class': 'ts-drawer-container',
+    '[class.ts-drawer-container-explicit-backdrop]': '_hasBackdrop',
```
benjamincharity: We should reference the getter vs the private property

terminus-ui/drawer/src/drawer-container.component.ts
```diff
@@ -0,0 +1,345 @@
...
+      }
+      this.changeDetectorRef.markForCheck();
+    });
+
+    // Arbitrary debounce time, less than a frame at 60fps
+    this.doCheckSubject.pipe(
+      debounceTime(1),
+      untilComponentDestroyed(this)
+    ).subscribe(() => this.updateContentMargins());
+  }
```
benjamincharity: Can you explain this and the related code a bit? I'm not sure I exactly follow what this is solving..

terminus-ui/drawer/src/drawer-container.component.ts
```diff
@@ -0,0 +1,345 @@
...
+
+  constructor(private element: ElementRef<HTMLElement>,
+              private ngZone: NgZone,
+              private changeDetectorRef: ChangeDetectorRef,
+              @Inject(TS_DRAWER_DEFAULT_AUTOSIZE) defaultAutosize = false,
+              @Optional() @Inject(ANIMATION_MODULE_TYPE) private animationMode?: string) {
+  }
+
+  public ngAfterContentInit() {
+    this.drawers.changes.pipe(startWith(null)).subscribe(() => {
```
benjamincharity: Missing `untilComponentDestroyed(this)`

terminus-ui/drawer/src/drawer-container.component.ts
```diff
@@ -0,0 +1,345 @@
...
+  /**
+   * For drawers in `overlay` mode, they don't affect the content.
+   * For drawers in `push` mode they should shrink the content. We do this by adding to the
+   * left margin (for left drawer) or right margin (for right the drawer).
+   * If the content margin has already been set and there are more than 1 opening windows on the same size, do not update.
+   *
+   * @params - drawer component, side of the drawer and margin on that side
+   * @return - calculated margin on the specific side.
+   */
+  private marginAdjustment(drawer: TsDrawerComponent | null, side: string, margin: number): number {
```
benjamincharity: Would it be possible to make the function names more explicit? From a glance it's unclear how this function is different from `updateContentMargins`

terminus-ui/drawer/src/drawer-container.component.ts
```diff
@@ -0,0 +1,345 @@
...
+  public get hasBackdrop() {
+    return this._hasBackdrop;
+  }
+  public _hasBackdrop = false;
+
+  /**
+   * Event emitted when the drawer backdrop is clicked.
+   */
+  @Output()
+  private readonly backdropClick: EventEmitter<void> = new EventEmitter<void>();
```
benjamincharity: We can omit the type on the property here and just do `backdropClick = new EventEmitter<void>();` now

terminus-ui/drawer/src/drawer-container.component.ts
```diff
@@ -0,0 +1,345 @@
...
+   * Margins to be applied to the content. These are used to push / shrink the drawer content when a
+   * drawer is open. We use margin rather than transform even for push mode because transform breaks
+   * fixed position elements inside of the transformed element.
+   */
+  public contentMargins: { left: number|null; right: number|null } = {
+    left: null,
+    right: null,
+  };
+
+  public readonly contentMarginChanges = new Subject<{left: number|null; right: number|null}>();
```
benjamincharity: This type `{left: number|null; right: number|null}` is used in more than one place, so we should probably pull it out into a 'real' type and reference it when needed.

terminus-ui/drawer/src/drawer-container.component.ts
benjamincharity: Also, I believe this is the deprecated usage, I _think_ we need to provide two types: `startWith<void, null>(null)`. (See `ChipCollectionComponent`)

terminus-ui/drawer/src/drawer-container.component.ts
```diff
@@ -0,0 +1,345 @@
...
+  /**
+   * Event emitted when the drawer backdrop is clicked.
+   */
+  @Output()
+  private readonly backdropClick: EventEmitter<void> = new EventEmitter<void>();
+
+  constructor(private element: ElementRef<HTMLElement>,
+              private ngZone: NgZone,
+              private changeDetectorRef: ChangeDetectorRef,
+              @Inject(TS_DRAWER_DEFAULT_AUTOSIZE) defaultAutosize = false,
```
benjamincharity: I don't seem to see this in use anywhere.. is it legacy code?

terminus-ui/drawer/src/drawer-container.component.ts
```diff
@@ -0,0 +1,345 @@
...
+      .filter(drawer => drawer && this.hasBackdrop)
+      .forEach(drawer => drawer && drawer.collapse());
+  }
+
+  public isShowingBackdrop(): boolean {
+    return (this.isDrawerOpen(this.start) && this.hasBackdrop)
+      || (this.isDrawerOpen(this.end) && this.hasBackdrop);
+  }
+
+  private isDrawerOpen(drawer: TsDrawerComponent | null): drawer is TsDrawerComponent {
```
benjamincharity: Are we relying on the coercion of the type to `TsDrawerComponent`? It seems from the usage I see we only need to return a boolean?

benjamincharity: Also, I believe this method can be `static`

terminus-ui/drawer/src/drawer.component.html
```diff
@@ -0,0 +1,5 @@
+<div
+  class="ts-drawer-inner-container"
```
benjamincharity: a) BEM style: `ts-drawer__inner-container`
b) This can all be on one line since there is only a single HTML attribute

terminus-ui/drawer/src/drawer-content.component.ts
```diff
@@ -0,0 +1,51 @@
...
+  template: '<ng-content></ng-content>',
+  host: {
+    'class': 'ts-drawer-content',
+    '[style.margin-left.px]': 'container.contentMargins.left',
+    '[style.margin-right.px]': 'container.contentMargins.right',
+  },
+  changeDetection: ChangeDetectionStrategy.OnPush,
+  encapsulation: ViewEncapsulation.None,
+})
+export class TsDrawerContentComponent extends CdkScrollable implements AfterContentInit {
```
benjamincharity: What do we gain by extending `CdkScrollable`?

terminus-ui/drawer/src/drawer.component.scss
```diff
@@ -0,0 +1,115 @@
...
+  // We need a stacking context here so that the backdrop and drawers are clipped to the
+  // TsDrawerContainer. This creates a new z-index stack so we use low numbered z-indices.
+  // We create another stacking context in the '.ts-drawer-content' and in each drawer so that
+  // the application content does not get messed up with our own CSS.
+  @include ts-drawer-stacking-context;
+  box-sizing: border-box;
+  // Need this to take up space in the layout.
+  display: block;
+  -webkit-overflow-scrolling: touch;
+  &[fullscreen] {
```
benjamincharity: What is this getting us? If someone adds the word `fullscreen` to the DOM we blow it out to cover everything?

terminus-ui/drawer/src/drawer.component.scss
```diff
@@ -0,0 +1,115 @@
...
+  // transitions, while visibility will interpolate transitions properly.
+  visibility: hidden;
+  z-index: $ts-drawer-backdrop-z-index;
+  &.ts-drawer-shown {
+    visibility: visible;
+  }
+
+  .ts-drawer-transition & {
+    transition: {
+      duration: 400ms;
```
benjamincharity: We use this duration in a few places so we should probably pull it to the top as a variable

terminus-ui/drawer/src/drawer.component.scss
```diff
@@ -0,0 +1,115 @@
...
+      property: background-color, visibility;
+      timing-function: cubic-bezier($g-material-background-easing);
```
benjamincharity, line 61: 🎉

terminus-ui/drawer/src/drawer.component.scss
```diff
@@ -0,0 +1,115 @@
...
+  $high-contrast-border: solid 1px currentColor;
+
+  background-color: color(utility, light);
+
+  @include ts-drawer-stacking-context($ts-drawer-over-drawer-z-index);
+  @include elevation-element(menu);
+  bottom: 0;
+  box-sizing: border-box;
+  display: block;
+  outline: 0;
```
benjamincharity: What is this doing? I don't think divs receive outline?

terminus-ui/drawer/src/drawer.component.ts
```diff
@@ -0,0 +1,343 @@
...
+    if (isOpen) {
+      this.animationState = this.enableAnimations ? 'open' : 'open-instant';
+      // this.elementRef.nativeElement.focus();
+    } else {
+      this.animationState = 'void';
+    }
+
+    return new Promise<TsDrawerToggleResult>(resolve => {
+      this.expandedChange.pipe(take(1)).subscribe(open => resolve(open ? 'open' : 'close'));
+    });
```
benjamincharity, line 351: It doesn't seem like anyone is using this promise.. should the method just return void?

Also, need to add `untilComponentDestroyed`

terminus-ui/drawer/src/drawer.component.ts
```diff
@@ -0,0 +1,343 @@
...
+    return this.elementRef.nativeElement ? (this.elementRef.nativeElement.offsetWidth || 0) : 0;
+  }
+
+  // aria role label, default to nothing
+  @Input()
+  public role = '';
+
+  // Collapsed drawer width
+  @Input()
+  public set collapsedSize(value: string) {
```
benjamincharity, line 144: For `collapsedSize` and `expandedSize` we should abstract out the default to a constant above, then if the user passes in `null` or `undefined` we can fall back to the default. (Our utility `isUnset` would work well here.)

terminus-ui/drawer/src/drawer.component.ts
```diff
@@ -0,0 +1,343 @@
...
+  }
+  private _mode: TsDrawerModes = 'overlay';
+
+  /**
+   * Whether the drawer is opened. We overload this because we trigger an event when it
+   * starts or end.
+   */
+  @Input()
+  public set isExpanded(value: boolean) {
+    this.toggle(coerceBooleanProperty(value));
```
benjamincharity: No need to coerce the boolean

terminus-ui/drawer/src/drawer.component.ts
```diff
@@ -0,0 +1,343 @@
...
+  }
+  public get isExpanded(): boolean {
+    return this._isExpanded;
+  }
+  private _isExpanded = false;
+
+  // Event emitted when the drawer open state is changed.
+  @Output() public readonly expandedChange: EventEmitter<boolean> =
+    // Note this has to be async in order to avoid some issues with two-bindings.
+    new EventEmitter<boolean>(/* isAsync */true);
```
benjamincharity: Does this mean that the `isAsync` comment actually has an impact on the compilation?

Also, can use the simpler method of defining emitters where the property doesn't have a defined type: `expandedChange = new EventEmitter<boolean>(true);`

terminus-ui/drawer/src/drawer.component.ts
```diff
@@ -0,0 +1,343 @@
...
+      filter(e => e.fromState !== e.toState && e.toState === 'void'),
+      untilComponentDestroyed(this),
+      map(() => {})
+    );
+  }
+
+  // Event emitted when the drawer's position changes.
+  // tslint:disable-next-line:no-output-on-prefix
+  @Output('positionChanged')
+  public readonly onPositionChanged: EventEmitter<void> = new EventEmitter<void>();
```
benjamincharity: Simpler emitter format viable here

terminus-ui/drawer/src/drawer.component.ts
```diff
@@ -0,0 +1,343 @@
...
+
+  constructor(public elementRef: ElementRef<HTMLElement>,
+              private platform: Platform,
+              private ngZone: NgZone,
+              public renderer: Renderer2,
+              // tslint:disable-next-line:no-any
+              @Optional() @Inject(DOCUMENT) private doc: any) {
```
benjamincharity: We can use `TsDocumentService` here

terminus-ui/drawer/src/drawer.component.ts
```diff
@@ -0,0 +1,343 @@
...
+
+    /**
+     * Listen to `keydown` events outside the zone so that change detection is not run every
+     * time a key is pressed. Re-enter the zone only if the `ESC` key is pressed
+     * and we don't have close disabled.
```
benjamincharity: I don't believe the 'close disabled' part is still valid

terminus-ui/drawer/src/drawer.component.ts
```diff
@@ -0,0 +1,343 @@
...
+
+    // We need a Subject with distinctUntilChanged, because the `done` event
+    // fires twice on some browsers.
+    this.animationEnd.pipe(
+      distinctUntilChanged((x, y) => x.fromState === y.fromState && x.toState === y.toState),
+      untilComponentDestroyed(this),
+    ).subscribe((event: AnimationEvent) => {
+      const { fromState, toState } = event;
+
+      if ((toState.indexOf('open') === 0 && fromState === 'void')
```
benjamincharity: This can fit on one line and may be a bit more readable

terminus-ui/drawer/src/drawer.component.ts
```diff
@@ -0,0 +1,343 @@
...
+        || (toState === 'void' && fromState.indexOf('open') === 0)
+      ) {
+        this.expandedChange.emit(this.isExpanded);
+      }
+    });
+    this.renderer.setStyle(this.elementRef.nativeElement, 'width', this._expandedSize);
```
benjamincharity: We should use the getter vs the private property

terminus-ui/drawer/src/drawer.component.ts
```diff
@@ -0,0 +1,343 @@
...
+    return this.toggle(true);
+  }
+
+  /**
+   * Collapse the drawer.
+   *
+   * @return Promise<TsDrawerToggleResult>
+   */
+  public collapse(): Promise<TsDrawerToggleResult> {
+    this.elementRef.nativeElement.classList.add('ts-drawer--collapsed');
```
benjamincharity: Seems like class addition/removal would be better suited as part of the `toggle` method?

terminus-ui/drawer/src/drawer.component.ts
```diff
@@ -0,0 +1,343 @@
...
+  @HostListener('@transform.start', ['$event'])
+  public animationStartListener(event: AnimationEvent) {
+    this.animationStarted.next(event);
+  }
+  //
+  // We have to use a `HostListener` here in order to support both Ivy and ViewEngine.
+  // In Ivy the `host` bindings will be merged when this class is extended, whereas in
+  // ViewEngine they're overwritten.
+  // TODO: move this back into `host` once Ivy is turned on by default.
+  // tslint:disable-next-line:no-host-decorator-in-concrete
```
benjamincharity: I'm unfamiliar with `no-host-decorator-in-concrete`.. is that a custom lint rule?

benjamincharity (changes requested):
This review does **not** include spec files.

General comments here to avoid noise:
- Missing JSDoc comments
- Missing return types
- Missing space between 3rd and 1st party imports

benjamincharity:
Forgot to add one other review item:

It seems that the animation is delayed on the content:

drawerAnimationTiming

I would expect them to animate in sync

demo/app/components/drawer/drawer.component.html
atlwendy: Based on the requirements, default is `3.75rem`

terminus-ui/drawer/src/drawer-container.component.html
atlwendy: But `TsDrawerContent` is actually a component, which has its selector as `ts-drawer-content`

benjamincharity: Ah yeah I misread.. you are correct.

terminus-ui/drawer/src/drawer-container.component.ts
atlwendy: It's referenced by `isPushed()` method. If it returns `boolean` only, `this.start` after `&&` could be null. But if we put the guard as `TsDrawerComponent`, that'd be fine.

terminus-ui/drawer/src/drawer.component.scss
atlwendy: yeah I think so. We don't have to have it though...

terminus-ui/drawer/src/drawer.component.scss
atlwendy: It doesn't? doesn't it draw outline of the element?

terminus-ui/drawer/src/drawer.component.ts
atlwendy: Don't think that comment has any code/compile impact. It's more of explanation,  what `true` is for. Could move to the line above.

terminus-ui/drawer/src/drawer.component.ts
atlwendy: I guess it's fine either way... had preferred here because it was more clear here versus in toggle, it'd live in a condition block.

terminus-ui/drawer/src/drawer.component.ts
atlwendy: Does it need `untilComponentDestroyed` when it's just `take(1)`?

terminus-ui/drawer/src/drawer.component.ts
benjamincharity: I only say that because this function is not actually in control of collapsing the drawer. It just calls another function that is actually in charge of adjusting the drawer.

terminus-ui/drawer/src/drawer.component.ts
benjamincharity: Ah! Nope! I must have blown right by that

terminus-ui/drawer/src/drawer.component.scss
benjamincharity: I think we can lose it.. since our drawers can be inside any container `fullscreen` likely wouldn't actually be fullscreen anyway. If product ends up requesting the ability to fill the container we can add an `@Input` etc

terminus-ui/drawer/src/drawer.component.scss
benjamincharity: You can manually add an outline via CSS: `outline: 1px solid red` but afaik the browser only adds a outline automatically to 'focusable' elements such as buttons and links.

So unless someone upstream is setting an outline that we are cancelling out I don't believe it's needed.

terminus-ui/drawer/src/drawer-container.component.ts
benjamincharity: If I'm seeing it correctly, `isPushed()` just needs a boolean, correct? So couldn't we just set the return type to `boolean` vs `drawer is TsDrawerComponent`? (maybe I'm missing where we need the type to be coerced?)

terminus-ui/drawer/src/drawer.component.ts
benjamincharity: Ah. Yeah, if this is a needed note about the Emitter, I'd move it to the Emitter comment above. IMO the inline component is odd just cause we don't do that anywhere.

.gitignore
benjamincharity: ☝️

README.md
```diff
@@ -13,7 +13,7 @@ The library of UI components used for Terminus applications.
 [![Github release][gh-release-badge]][gh-releases]
 [![Library size][file-size-badge]][raw-distribution-js]
 <br>
-[![components](https://img.shields.io/badge/Components-35-%2316a085.svg)](#components)
+[![components](https://img.shields.io/badge/Components-38-%2316a085.svg)](#components)
```
benjamincharity, line 16: I _think_ this should probably just increase by 1? So far we've been counting only items that can be used alone. eg Tabs has several components/directives, but to the consumer we are really offering 1 more thing.

demo/app/components/drawer/drawer.component.ts
```diff
@@ -0,0 +1,77 @@
...
+  Validators,
+} from '@angular/forms';
+import {
+  TsDrawerContainerComponent,
+  TsDrawerModes,
+} from '@terminus/ui/drawer';
+import { TsRadioFormatFn } from '@terminus/ui/radio-group';
+import { TsStyleThemeTypes } from '@terminus/ui/utilities';
+
+const currentDate: Date = new Date();
```
benjamincharity: I don't think this is used anywhere

demo/app/components/drawer/drawer.component.ts
```diff
@@ -0,0 +1,77 @@
+import {
+  AfterViewChecked,
```
benjamincharity: Unused

demo/app/components/drawer/drawer.component.ts
```diff
@@ -0,0 +1,77 @@
...
+} from '@angular/core';
+import {
+  FormBuilder,
+  FormControl,
+  FormGroup,
+  Validators,
+} from '@angular/forms';
+import {
+  TsDrawerContainerComponent,
+  TsDrawerModes,
```
benjamincharity: Unused import

terminus-ui/drawer/src/drawer-container.component.spec.ts
```diff
@@ -0,0 +1,218 @@
...
+import {
+  FormsModule,
+  ReactiveFormsModule,
+} from '@angular/forms';
+import { MatRippleModule } from '@angular/material/core';
+import { By } from '@angular/platform-browser';
+import { BrowserAnimationsModule } from '@angular/platform-browser/animations';
+import { createComponent as createComponentInner } from '@terminus/ngx-tools/testing';
+import * as testComponents from '@terminus/ui/drawer/testing';
+import { TsDrawerComponent } from './drawer.component';
```
benjamincharity, line 22: Line break before 1st party imports

terminus-ui/drawer/src/drawer-container.component.spec.ts
```diff
@@ -0,0 +1,218 @@
...
+
+describe(`TsDrawerContainerComponent`, () => {
+  let fixture;
+  let component;
+  let drawers;
+
+  function setUpTestComponent(testComponent) {
+    fixture = createComponent(testComponent);
+    fixture.detectChanges();
+    component = fixture.debugElement.queryAll(By.css('ts-drawer-container'))[0].componentInstance;
```
benjamincharity, line 48: I don't think we should delay this PR any more, but for future reference, I think we should abstract the queries you need into that test-helpers file so that consumers and our other specs can use them too.

terminus-ui/drawer/src/drawer-container.component.spec.ts
```diff
@@ -0,0 +1,218 @@
...
+        configurable: true,
+        value: 500,
+      });
+      fixture.componentInstance.drawer.expand();
+      fixture.detectChanges();
+      tick();
+      expect(component.contentMargins.left).toEqual(204);
+
+    }));
+    test(`should calculate the right margin`, fakeAsync(() => {
```
benjamincharity: Missing line break

terminus-ui/drawer/src/drawer-container.component.ts
```diff
@@ -0,0 +1,368 @@
...
+export function throwTsDuplicatedDrawerError(position: string): void {
+  throw new TsUILibraryError(`TsDrawerContainer: Only one drawer can be in push mode on '"${position}" position'.`);
+}
+
+export interface ContentLeftRightMargin {
+  left: number | null;
+  right: number | null;
+}
+
+const MARGIN_SIZE_CONVERSION = 16;
```
benjamincharity: I know you mentioned this already, but since it's my last review I'm going to call them out 😬 

Missing comments

terminus-ui/drawer/src/drawer-container.component.ts
```diff
@@ -0,0 +1,368 @@
...
+  startWith,
+  take,
+  takeUntil,
+} from 'rxjs/operators';
+
+import { TsDrawerContentComponent } from './drawer-content.component';
+import { TsDrawerComponent } from './drawer.component';
+
+/**
+ * Throws an exception when two TsDrawer are matching the same position with same mode.
```
benjamincharity, line 38: This should be a real JSDoc with param

terminus-ui/drawer/src/drawer-container.component.ts
benjamincharity: ☝️

terminus-ui/drawer/src/drawer-container.component.ts
```diff
@@ -0,0 +1,368 @@
...
+  styleUrls: ['./drawer.component.scss'],
+  host: {
+    'class': 'ts-drawer-container',
+    '[class.ts-drawer-container-explicit-backdrop]': '_hasBackdrop',
+  },
+  changeDetection: ChangeDetectionStrategy.OnPush,
+  encapsulation: ViewEncapsulation.None,
+  exportAs: 'tsDrawerContainer',
+})
+export class TsDrawerContainerComponent implements AfterContentInit, DoCheck, OnDestroy {
```
benjamincharity, line 82: Missing JSDocs throughout

terminus-ui/drawer/src/drawer-container.component.ts
```diff
@@ -0,0 +1,368 @@
...
+    private element: ElementRef<HTMLElement>,
+    private ngZone: NgZone,
+    public renderer: Renderer2,
+    private changeDetectorRef: ChangeDetectorRef,
+    @Optional() @Inject(ANIMATION_MODULE_TYPE) private animationMode?: string
+  ) {
+  }
+
+  public ngAfterContentInit() {
+    const contentEle = this.element.nativeElement.querySelector('.ts-drawer-content');
```
benjamincharity: Is there a reason we need to do a query here rather than using the view reference created earlier?

terminus-ui/drawer/src/drawer-container.component.ts
```diff
@@ -0,0 +1,368 @@
...
+  public collapse(): void {
+    this.drawers.forEach(drawer => drawer.collapse());
+  }
+
+  /**
+   * Recalculates and updates the inline styles for the content.
+   */
+  public updateContentMargins() {
+    let left = 0;
+    let right = 0;
```
benjamincharity: I think these should be typed to `number | null`?

terminus-ui/drawer/src/drawer-container.component.ts
```diff
@@ -0,0 +1,368 @@
...
+    right = this.end ? this.marginCalculation(this.end, 'right', 0) : right;
+
+    // If either `right` or `left` is zero, don't set a style to the element. This
+    // allows users to specify a custom size via CSS class in SSR scenarios where the
+    // measured widths will always be zero. Note that we reset to `null` here, rather
+    // than below, in order to ensure that the types in the `if` below are consistent.
+    // tslint:disable-next-line:no-non-null-assertion
+    left = left || null!;
+    // tslint:disable-next-line:no-non-null-assertion
+    right = right || null!;
```
benjamincharity: I think we don't need these non-null-assertions here.

terminus-ui/drawer/src/drawer-container.component.ts
```diff
@@ -0,0 +1,368 @@
...
+   * Subscribes to drawer events in order to set a class on the main container element when the
+   * drawer is open and the backdrop is visible. This ensures any overflow on the container element
+   * is properly hidden.
+   *
+   * @params drawer - TsDrawerComponent
+   */
+  private adjustDrawerOnChange(drawer: TsDrawerComponent): void {
+    drawer.animationStarted.pipe(
+      filter(event => event.fromState !== event.toState),
+      takeUntil(this.drawers.changes),
```
benjamincharity, line 306: I _think_ we still need `untilComponentDestroyed` here?

terminus-ui/drawer/src/drawer-container.component.ts
```diff
@@ -0,0 +1,368 @@
...
+    drawer.positionChanged.pipe(takeUntil(this.drawers.changes)).subscribe(() => {
+      this.ngZone.onMicrotaskEmpty.asObservable().pipe(take(1)).subscribe(() => {
+        this.validateDrawers();
+      });
+    });
+  }
+
+  // Subscribes to changes in drawer mode so we can run change detection.
+  private drawerModeOnChange(drawer: TsDrawerComponent): void {
+    if (drawer) {
```
benjamincharity, line 347: We can set `// istanbul ignore else` above this if

terminus-ui/drawer/src/drawer-container.component.ts
```diff
@@ -0,0 +1,368 @@
...
+      drawer.modeChanged
+        .pipe(
+          takeUntil(this.drawers.changes),
+          untilComponentDestroyed(this)
```
benjamincharity: Trailing comma? Might need to run the linter?

terminus-ui/drawer/src/drawer-content.component.ts
```diff
@@ -0,0 +1,34 @@
...
+  selector: 'ts-drawer-content',
+  template: '<ng-content></ng-content>',
+  host: { class: 'ts-drawer-content' },
+  changeDetection: ChangeDetectionStrategy.OnPush,
+  encapsulation: ViewEncapsulation.None,
+})
+export class TsDrawerContentComponent extends CdkScrollable {
+  public leftMargin: number | null = null;
+  public rightMargin: number | null = null;
+  constructor(
```
benjamincharity, line 26: line break before constructor

terminus-ui/drawer/src/drawer.component.md
```diff
@@ -0,0 +1,131 @@
...
+
+```html
+<ts-drawer-container>
+  <ts-drawer
+    mode="push"
+    posistion="start"
+  >
+  <div>
+    Inside drawer
+  </div>
```
benjamincharity: Should be indented

terminus-ui/drawer/src/drawer.component.md
```diff
@@ -0,0 +1,131 @@
...
+  
+  </ts-drawer>
+</ts-drawer-container>
+```
+
+`hasBackdrop` defaults to `false`.
+
+### Mode
+
+For string based collections, the value input can be disregarded as the value will be pulled directly from the DOM.
```
benjamincharity: This seems to be a copy paste from chips

terminus-ui/drawer/src/drawer.component.md
```diff
@@ -0,0 +1,131 @@
...
+| Event              | Description                                    | Payload                  |
+|:-------------------|:-----------------------------------------------|:-------------------------|
+| `backdropClick`    | Fired when backdrop is clicked                 |  |
+
+### Drawer events
+
+| Event             | Description                                 | Payload                 |
+|:------------------|:--------------------------------------------|:------------------------|
+| `expandedStart`   | Fired when the drawer expansion starts      |                         |
+| `collapsedStart`  | Fired when the drawer collapse starts       |                         |
```
benjamincharity: We should define the payload as `void` in these cases

terminus-ui/drawer/src/drawer.component.scss
benjamincharity: ☝️

terminus-ui/drawer/src/drawer.component.scss
benjamincharity: ☝️

terminus-ui/drawer/src/drawer.component.spec.ts
```diff
@@ -0,0 +1,133 @@
...
+import {
+  TsDrawerComponent, TsDrawerModule,
+} from './drawer.module';
+import {
+  createComponent as createComponentInner,
+  dispatchKeyboardEvent,
+} from '@terminus/ngx-tools/testing';
+import * as testComponents from '@terminus/ui/drawer/testing';
+import { By } from '@angular/platform-browser';
+import { KEYS } from '@terminus/ngx-tools';
```
benjamincharity: Import ordering. Fixing tslint issues should solve this

terminus-ui/drawer/src/drawer.component.spec.ts
```diff
@@ -0,0 +1,133 @@
...
+
+    describe(`basic behaviors`, () => {
+      test(`should add ts-drawer class`, () => {
+        setup();
+        expect(drawerNativeElement.classList).toContain('ts-drawer');
+      });
+
+      test(`should allow direct setting of mode`, () => {
+
+      });
```
benjamincharity: Empty test

terminus-ui/drawer/src/drawer.component.spec.ts
```diff
@@ -0,0 +1,133 @@
...
+      test('should resolve the open method promise with the new state of the drawer', fakeAsync(() => {
+        setup(testComponents.SimpleDrawer);
+        fixture.detectChanges();
+
+        drawer.expand().then(result => expect(result).toBe('open'));
+        fixture.detectChanges();
+        tick();
+      }));
+    });
+    test('should close when pressing escape', fakeAsync(() => {
```
benjamincharity: linebreak

terminus-ui/drawer/src/drawer.component.ts
```diff
@@ -0,0 +1,317 @@
...
+import {
+  distinctUntilChanged,
+  filter,
+  map,
+  take,
+} from 'rxjs/operators';
+
+import { tsDrawerAnimations } from './drawer-animations';
+
+// Result of the toggle promise that indicates the state of the drawer.
```
benjamincharity: General: JSDoc comments from here on

terminus-ui/drawer/src/drawer.component.ts
```diff
@@ -0,0 +1,317 @@
...
+   * Toggle this drawer.
+   *
+   * @param isOpen - whether the drawer should be open.
+   * @return  Promise<TsDrawerToggleResult>
+   */
+  public toggle(isOpen: boolean = !this.isExpanded): Promise<TsDrawerToggleResult> {
+    this._isExpanded = isOpen;
+    if (isOpen) {
+      this.animationState = this.enableAnimations ? 'open' : 'open-instant';
+      // this.elementRef.nativeElement.focus();
```
benjamincharity: Old code

terminus-ui/drawer/src/drawer.component.ts
```diff
@@ -0,0 +1,317 @@
...
+    } else {
+      this.animationState = 'void';
+    }
+
+    return new Promise<TsDrawerToggleResult>(resolve => {
```
benjamincharity, line 349: It doesn't seem like anyone is using this promise.. should the method just return void?

terminus-ui/drawer/testing/src/test-components.ts
```diff
@@ -0,0 +1,248 @@
+import {
+  Component, ElementRef, NgModule, ViewChild,
```
benjamincharity: Multiple lines

terminus-ui/drawer/testing/src/test-components.ts
```diff
@@ -0,0 +1,248 @@
...
+} from '@angular/core';
+
+import {
+  TsDrawerComponent,
+  TsDrawerModes, TsDrawerModule,
+  TsDrawerPosition,
+  TsDrawerContainerComponent, TsDrawerContentComponent,
+} from '@terminus/ui/drawer';
```
benjamincharity: a) Multiple lines.
b) It doesn't look like `TsDrawerContainerComponent` is being used
c) Imports not ordered

benjamincharity (changes requested):
Mostly cleanup stuffs. Looks like we need to rebase this against `release` and run all linters.

README.md
atlwendy: I thought I have seen other pr that had 37, so thinking this pr might just be behind. Will get this resolved once rebase to release. I'm purposely delaying rebase because want to get this approved first, then squash, rebase once to make final cut...

benjamincharity: Yeah rebasing sooner than later is probably good cause we will still need to look it over again after the rebase.

terminus-ui/drawer/src/drawer.component.ts
atlwendy: For example, expand call will resolve this to the expandedChange being at `open` state. Probably not a critical usage. But it does have a chain of actions.

terminus-ui/drawer/src/drawer-container.component.ts
atlwendy: My thought was that `takeUntil` would stop when `drawer` is destroyed and no changes can be emitted. So there is no need to add that. But it doesn't hurt either...

benjamincharity: I'm not sure that would actually stop. If drawers were de-registered during the destroy lifecycle, `this.drawers` would still be an empty query list so the subscribe would happen again.

terminus-ui/drawer/src/drawer-container.component.ts
atlwendy: The only reason is that content child `content` isn't always there. So to use `this.content`, a check needs to be put there. This way if it's not there, it'd just return null..

benjamincharity: > The only reason is that content child content isn't always there

This logic should probably be moved into `ngAfterViewInit` then. QueryLists don't exists prior to that lifecycle.

Or I guess lose the querylist and just use this though it doesn't feel as 'nice'

terminus-ui/drawer/src/drawer-container.component.ts
benjamincharity: Still missing comments and return types throughout

terminus-ui/drawer/src/drawer.component.ts
benjamincharity: Same comments format issue. Should be aligned with the rest of our components

benjamincharity (changes requested):
Looks like just return types and comment formats left 🎉

terminus-ui/drawer/src/drawer.component.ts
```diff
@@ -0,0 +1,373 @@
...
+  filter,
+  map,
+  take,
+} from 'rxjs/operators';
+
+import { tsDrawerAnimations } from './drawer-animations';
+
+/**
+ * Result of the toggle promise that indicates the state of the drawer.
+ */
```
atlwendy, line 40: I was under the false impression that if the comment is longer than one line we use `/** */` style and if it's within one line we use `//` instead.... guess it's not.
It's all updated now.

terminus-ui/drawer/src/drawer-container.component.ts
```diff
@@ -0,0 +1,400 @@
...
+/**
+ * Throws an exception when two TsDrawer are matching the same position with same mode.
+ *
+ * @params - drawer's position
+ */
+export function throwTsDuplicatedDrawerError(position: string): void {
+  throw new TsUILibraryError(`TsDrawerContainer: Only one drawer can be in push mode on '"${position}" position'.`);
+}
+
+export interface ContentLeftRightMargin {
```
benjamincharity, line 49: Missing JSDOC comment.

terminus-ui/drawer/src/drawer-container.component.ts
```diff
@@ -0,0 +1,400 @@
...
+  exportAs: 'tsDrawerContainer',
+})
+export class TsDrawerContainerComponent implements AfterContentInit, DoCheck, OnDestroy {
+
+  // The drawer at the start/end position.
+  private start: TsDrawerComponent | null = null;
+  private end: TsDrawerComponent | null = null;
+  private openDrawerCount = 0;
+
+  // Emits on every ngDoCheck. Used for debouncing reflows.
```
benjamincharity: Any specific reason these comments aren't formatted like our other components? Are you hoping for a change in our formats?

terminus-ui/drawer/src/drawer-container.component.ts
```diff
@@ -0,0 +1,400 @@
...
+  private readonly doCheckSubject = new Subject<void>();
+
+  // Whether the container is being pushed to the side by one of the drawers.
+  private get isPushed(): boolean {
+    return (TsDrawerContainerComponent.isDrawerOpen(this.start) && this.start.mode === 'push')
+      || (TsDrawerContainerComponent.isDrawerOpen(this.end) && this.end.mode === 'push');
+  }
+
+  public get isShowingBackdrop(): boolean {
```
benjamincharity, line 107: Missing JSDOC comment.

terminus-ui/drawer/src/drawer-container.component.ts
```diff
@@ -0,0 +1,400 @@
...
+    return (TsDrawerContainerComponent.isDrawerOpen(this.start) && this.hasBackdrop)
+      || (TsDrawerContainerComponent.isDrawerOpen(this.end) && this.hasBackdrop);
+  }
+
+  /**
+   * Access the child drawer {@link TsDrawerComponent}
+   */
+  @ContentChildren(TsDrawerComponent)
+  public drawers!: QueryList<TsDrawerComponent>;
```
benjamincharity, line 116: Missing line break

terminus-ui/drawer/src/drawer-container.component.ts
```diff
@@ -0,0 +1,400 @@
...
+   * Margins to be applied to the content. These are used to push / shrink the drawer content when a
+   * drawer is open. We use margin rather than transform even for push mode because transform breaks
+   * fixed position elements inside of the transformed element.
+   */
+  public contentMargins: ContentLeftRightMargin = {
+    left: null,
+    right: null,
+  };
+
+  public readonly contentMarginChanges = new Subject<ContentLeftRightMargin>();
```
benjamincharity, line 137: Missing JSDOC comment.

terminus-ui/drawer/src/drawer-container.component.ts
```diff
@@ -0,0 +1,400 @@
...
+
+  /**
+   * Whether the drawer container should have a backdrop while one of the drawers is open.
+   * If explicitly set to `true`, the backdrop will be enabled for drawers.
+   */
+  @Input()
+  public set hasBackdrop(value: boolean) {
+    this._hasBackdrop = value;
+  }
+  public get hasBackdrop() {
```
benjamincharity: Missing return type.

terminus-ui/drawer/src/drawer-container.component.ts
```diff
@@ -0,0 +1,400 @@
...
+   */
+  @Output()
+  private readonly backdropClick = new EventEmitter<void>();
+
+  constructor(
+    private element: ElementRef<HTMLElement>,
+    private ngZone: NgZone,
+    public renderer: Renderer2,
+    private changeDetectorRef: ChangeDetectorRef,
+    @Optional() @Inject(ANIMATION_MODULE_TYPE) private animationMode?: string
```
benjamincharity: Trailing comma still is not added. Let me know if you need help getting the linter to work on your machine. It's pretty important to get that stuff automated so we can focus reviews on functionality

terminus-ui/drawer/src/drawer-container.component.ts
```diff
@@ -0,0 +1,400 @@
...
+  ) {
+  }
+
+  public ngAfterContentInit() {
```
benjamincharity: Missing JSDOC comment.

Missing return type.

terminus-ui/drawer/src/drawer-container.component.ts
```diff
@@ -0,0 +1,400 @@
...
+        this.content.leftMargin = e.left;
+        this.content.rightMargin = e.right;
+        this.renderer.setStyle(contentEle, 'margin-left', e.left ? `${e.left.toString()}px` : 0);
+        this.renderer.setStyle(contentEle, 'margin-right', e.right ? `${e.right.toString()}px` : 0);
+        this.changeDetectorRef.markForCheck();
+      }
+    });
+  }
+
+  public ngDoCheck() {
```
benjamincharity: Missing JSDOC comment.

Missing return type.

terminus-ui/drawer/src/drawer-container.component.ts
```diff
@@ -0,0 +1,400 @@
...
+    if (this.isPushed) {
+      // Run outside the NgZone, otherwise the debouncer will throw us into an infinite loop.
+      this.ngZone.runOutsideAngular(() => this.doCheckSubject.next());
+    }
+  }
+
+  public ngOnDestroy() {
```
benjamincharity: Missing JSDOC comment.

Missing return type.

terminus-ui/drawer/src/drawer-container.component.ts
```diff
@@ -0,0 +1,400 @@
...
+    }
+  }
+
+  /**
+   * For drawers in `overlay` mode, they don't affect the content.
+   * For drawers in `push` mode they should shrink the content. We do this by adding to the
+   * left margin (for left drawer) or right margin (for right the drawer).
+   * If the content margin has already been set and there are more than 1 opening windows on the same size, do not update.
+   *
+   * @params - drawer component, side of the drawer and margin on that side
```
benjamincharity: Params should be on individual lines.

terminus-ui/drawer/src/drawer-container.component.ts
```diff
@@ -0,0 +1,400 @@
...
+    // NOTE: We need to wait for the microtask queue to be empty before validating,
+    // since both drawers may be swapping positions at the same time.
+    drawer.positionChanged.pipe(takeUntil(this.drawers.changes)).subscribe(() => {
+      this.ngZone.onMicrotaskEmpty.asObservable().pipe(take(1)).subscribe(() => {
+        this.validateDrawers();
+      });
+    });
+  }
+
+  // Subscribes to changes in drawer mode so we can run change detection.
```
benjamincharity: Incorrect JSDOC comment format.

terminus-ui/drawer/src/drawer-container.component.ts
```diff
@@ -0,0 +1,400 @@
...
+          this.updateContentMargins();
+          this.changeDetectorRef.markForCheck();
+        });
+    }
+  }
+
+  /**
+   * Toggles the 'ts-drawer--expanded' class on the main 'ts-drawer-container' element.
+   *
+   * @params isAdd: boolean
```
benjamincharity: Params don't need to define the type since we have TypeScript, but they do need to describe the param content/purpose.

terminus-ui/drawer/src/drawer.component.ts
```diff
@@ -0,0 +1,374 @@
...
+  public role = '';
+
+  /**
+   * Collapsed drawer width
+   */
+  @Input()
+  public set collapsedSize(value: string) {
+    this._collapsedSize = isUnset(value) ? TS_DRAWER_DEFAULT_COLLAPSE_SIZE : value;
+  }
+  public get collapsedSize() {
```
benjamincharity: Missing return type.

terminus-ui/drawer/src/drawer.component.ts
```diff
@@ -0,0 +1,374 @@
...
+  public _collapsedSize = '3.75rem';
+
+  /**
+   * Expanded drawer width
+   */
+  @Input()
+  public set expandedSize(value: string) {
+    this._expandedSize = isUnset(value) ? TS_DRAWER_DEFAULT_EXPAND_SIZE : value;
+  }
+  public get expandedSize() {
```
benjamincharity: Missing return type.

terminus-ui/drawer/src/drawer.component.ts
```diff
@@ -0,0 +1,374 @@
...
+    );
+  }
+
+  /**
+   * Event emitted when the drawer's position changes.
+   */
+  // tslint:disable-next-line:no-output-on-prefix
+  // tslint:disable-next-line:no-output-rename
+  @Output('positionChanged')
+  public readonly positionChanged: EventEmitter<void> = new EventEmitter<void>();
```
benjamincharity: We can remove the type from this declaration

terminus-ui/drawer/src/drawer.component.ts
```diff
@@ -0,0 +1,374 @@
...
+
+    // We need a Subject with distinctUntilChanged, because the `done` event fires twice on some browsers.
+    this.animationEnd.pipe(
+      distinctUntilChanged((x, y) => x.fromState === y.fromState && x.toState === y.toState),
+      untilComponentDestroyed(this),
+    ).subscribe((event: AnimationEvent) => {
+      const { fromState, toState } = event;
+
+      if ((toState.indexOf('open') === 0 && fromState === 'void') || (toState === 'void' && fromState.indexOf('open') === 0)
+      ) {
```
benjamincharity: These two: `) {` should be on the previous line

terminus-ui/drawer/src/drawer.component.ts
```diff
@@ -0,0 +1,374 @@
...
+        this.expandedChange.emit(this.isExpanded);
+      }
+    });
+    this.renderer.setStyle(this.elementRef.nativeElement, 'width', this.expandedSize);
+  }
+
+  public ngAfterContentChecked() {
```
benjamincharity: Missing JSDOC comment.

Missing return type.

terminus-ui/drawer/src/drawer.component.ts
```diff
@@ -0,0 +1,374 @@
...
+    // Enable the animations after the lifecycle hooks have run, in order to avoid animating
+    // drawers that are open by default.
+    if (this.platform.isBrowser) {
+      this.enableAnimations = true;
+    }
+  }
+
+  public ngOnDestroy() {
```
benjamincharity: Missing JSDOC comment.

Missing return type.

terminus-ui/drawer/src/drawer.component.ts
benjamincharity: We use the JSDoc format for all items that should appear in our generated documentation. Inline comments are used for explanations, notes, fixme's, etc

terminus-ui/drawer/src/drawer.component.ts
benjamincharity: Missing return type.

terminus-ui/drawer/src/drawer.component.ts
benjamincharity: Missing return type.

terminus-ui/drawer/src/drawer.component.ts
```diff
@@ -0,0 +1,374 @@
...
+  public get isExpanded(): boolean {
+    return this._isExpanded;
+  }
+  private _isExpanded = false;
+
+  /**
+   * Event emitted when the drawer open state is changed.
+   */
+  @Output()
+  public readonly expandedChange: EventEmitter<boolean> =
```
benjamincharity: This should be one line.

We can also remove the type definition.

terminus-ui/drawer/src/drawer.component.ts
benjamincharity: Unneeded type

terminus-ui/drawer/src/drawer.component.ts
benjamincharity: Missing JSDOC comment.

Missing return type.

terminus-ui/drawer/src/drawer.component.ts
benjamincharity: Missing JSDOC comment.

Missing return type.

benjamincharity (changes requested):

benjamincharity (approved):
🚀 All we need now is a rebase on `release` and we can merge!

PR #1764, opened 2019-10-29
1682 form control size change

### LINKED ISSUE #1682: Input: Decrease padding to match the height of buttons
Our inputs are much taller than our buttons. We are attempting to align the height with our buttons to make our UI more consistent.

- [ ] Adjust styles as needed so that the input _visually_ matches the button height.
- [ ] Verify layout in all use-cases

### LINKED ISSUE #1683: SelectionList: Update styles to match the height of our buttons
Our selection list component is much taller than our buttons. We are attempting to align the height with our buttons to make our UI more consistent.

- [ ] Adjust styles as needed so that the selection list _visually_ matches the button height.
- [ ] Verify layout in all use-cases

### LINKED ISSUE #1685: CSV: Match cell height from the updated table styles
Once the 'compact' cell layout is released, we should align this component to the new padding/height.

### LINKED ISSUE #1707: Autocomplete: Adjust spacing to align with our buttons
We need to align our form control heights to match our button heights

### LINKED ISSUE #1708: Select: Adjust spacing to align with our buttons
We need to align our form control heights to match our button heights

### LINKED ISSUE #1716: DateRange: Change dash color
Update the color of the dash between input boxes to match the border color.

- [ ] Change the dash color to `color(utility, light)`
- [ ] Fix misspelling in CSS class `c-date-range__seperator` > `c-date-range__separator`

benjamincharity:
Closes epic #1684 

- CSVEntry: padding decreased. 
    - Closes #1685 
- FormField: padding decreased and alignments tweaked.   
    - Closes #1682 
    - Closes #1683 
    - Closes #1708 
    - Closes #1707 
- Input: Correctly reflect disabled status to native input
- SelectionList: Now correctly seeds the query stream on load.
- DateRange: Fix dash position and color
    - Closes #1716


terminus-ui/chip/src/chip.component.scss
```diff
@@ -5,13 +5,20 @@
...
+  --chip-backgroundColor-hover: #{color(utility)};
+  --chip-borderRadius: 16px;
+  --chip-color: #{color(pure, dark)};
+  --chip-remove-size: #{spacing(large)};
+
   &:focus,
   &:hover {
     .c-chip {
-      background-color: $ts-chip-focus-background-color;
+      --chip-backgroundColor: var(--chip-backgroundColor-hover);
```
atlwendy, line 21: what's the difference between setting a variable like before and this?

terminus-ui/input/src/input.component.html
```diff
@@ -148,7 +148,7 @@
       [attr.autocapitalize]="autocapitalize ? 'on' : 'off'"
       [attr.autocomplete]="autocomplete"
       [attr.autofocus]="isFocused ? '' : null"
-      [attr.disabled]="shouldBeDisabled ? '' : null"
+      [disabled]="shouldBeDisabled"
```
atlwendy, line 151: you sure we don't need `null` here?

atlwendy (approved):
a lot of css changes!!

terminus-ui/chip/src/chip.component.scss
benjamincharity: The `$foo` is a Sass variable which is only available at compile time. While the `var(--foo)` format is a CSS custom property which is available at runtime.

Looking to move that direction for a couple reasons:
a) They have pretty good support at this point
b) They aren't tied to a specific library or build tool
c) They are available at runtime so we don't need to pre-generate styles for every theme we support for every component.

Sass variables are still fine for local variables that don't need to be changed. And honestly I'm still working out the details in my head so we won't be rejecting PRs using Sass vars etc. I'm just getting my feet wet here because I do think it will make overwrites/themes/extensions much more clean.

terminus-ui/input/src/input.component.html
benjamincharity: I believe so.. I went to a codepen and tried all the different versions and this one is the only one that seemed to actually disable the input for me. Our demos would disable the form field but the actual `<input>` was not being disabled. This seemed to fix that.

PR #1765, opened 2019-10-30
1706 simple paginator

benjamincharity:
⚠️  NOTE: Must be merged after #1764 (it is based off of that code) ⚠️ 

Since it is branched from that PR, only the last of the 6 commits listed here need to be reviewed. Last commit can be found here: https://github.com/GetTerminus/terminus-ui/pull/1765/commits/11cff609ef183244c7d8123af505056639034c33

---

- Paginator
  - Support simple mode
  - Add new test helpers
  - Convert tests to use new helpers
  - Convert commented out tests to `test.todo`s

Screen Shot 2019-10-30 at 7 34 35 AM

### DIFF
--- a/demo/app/components/paginator/paginator.component.html
+++ b/demo/app/components/paginator/paginator.component.html
@@ -23,6 +23,13 @@
     Pagination is zero-based:
     <input type="checkbox" [(ngModel)]="zeroBased" (ngModelChange)="updatePages($event)">
   </label>
+
+  <br>
+
+  <label>
+    Simple mode:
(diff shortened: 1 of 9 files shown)


demo/app/components/chart/chart.component.ts
```diff
@@ -69,6 +69,7 @@ export class ChartComponent {
     /**
      * XY
      */
+    am4charts.XYChart
```
benjamincharity: This should not be here

atlwendy (approved):

PR #1767, opened 2019-10-30
feat(CSVEntry): allow consumers to pass in custom footer content

### LINKED ISSUE #1690: CSV Entry: Add ability for consumers to add their own buttons below the grid
Product has requested the ability to define custom buttons below the CSV entry table.

Example:

- [ ] Use `ng-content` to allow consumer to pass in rich content
- [ ] Lay out with flexbox row. (use empty span between left and right with the `flex` directive to push contents to each side)
- [ ] Add an `@Input` to control the button layout `footerDirection: 'ltr' | 'rtl' = 'ltr';`
  - This flag should use flex layout to flip the order of footer items.
  - By default our buttons start on the left and consumer buttons on the right

benjamincharity:
also now allows consumer to control footer direction

ISSUES CLOSED: #1690

csv

### DIFF
--- a/demo/app/components/csv-entry/csv-entry.component.html
+++ b/demo/app/components/csv-entry/csv-entry.component.html
@@ -1,12 +1,33 @@
 <ts-card tsVerticalSpacing>
   <h3 tsVerticalSpacing>CSV Entry</h3>
 
+  <div>
+    Footer direction:
+    <ul>
+      <li>
+        <label>
+          ltr
+          <input type="radio" name="direction" value="ltr" [(ngModel)]="footerDirection">
+        </label>
+      </li>
+      <li>
+        <label>
+          rtl
+          <input type="radio" name="direction" value="rtl" [(ngModel)]="footerDirection">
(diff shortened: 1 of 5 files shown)


atlwendy (approved):
Would be great to have integration tests...

benjamincharity:
> Would be great to have integration tests...

What could I test? I originally was going to add tests, but it seems I'd really just be testing Angular (`ng-content`) and FlexLayoutModule (`dir`).

atlwendy:
I guess it's more of a visual integration test? For example, when you select `rtl`, the order of the buttons maintains the same, but the whole div moves to the right. 
I didn't mean we need to have it here, more of a wish that we had true integration tests..

benjamincharity:
Ahh yes, totally agree. I was about to even try to test that in mode 'ltr' the divs bounding box was at a different location etc.. but then I remembered I'd be mocking what the browser returned anyway 🙄 

This is _exactly_ why we need visual tests.

PR #1773, opened 2019-11-05
fix(Table): only update sticky columns if they exist

benjamincharity:
Also debouncing the sticky column update call

### DIFF
--- a/terminus-ui/table/src/table.component.ts
+++ b/terminus-ui/table/src/table.component.ts
@@ -1,7 +1,8 @@
 import { Directionality } from '@angular/cdk/bidi';
 import { Platform } from '@angular/cdk/platform';
 import {
-  CDK_TABLE_TEMPLATE, CdkTable,
+  CDK_TABLE_TEMPLATE,
+  CdkTable,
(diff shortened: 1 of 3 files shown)


terminus-ui/table/src/table.component.ts
```diff
@@ -142,10 +149,11 @@ export class TsTableColumnsChangeEvent {
 })
 // tslint:disable-next-line no-any
 export class TsTableComponent<T = any> extends CdkTable<T> implements OnInit, AfterViewChecked, OnDestroy {
+
   /**
-   * Override the sticky CSS class set by the `CdkTable`
+   * Create a debounced function to update CDK sticky styles
    */
-  protected stickyCssClass = 'ts-table--sticky';
+  public debouncedStickyColumnUpdate = debounce(this.updateStickyColumnStyles, COLUMN_DEBOUNCE_DELAY);
```
atlwendy, line 156: is the debounce trying to make sure updating sticky column doesn't happen too often that they'd cancel each other?

atlwendy (approved):

PR #1777, opened 2019-11-05
fix(Button): deprecate incorrect spelling of collapsible

benjamincharity:
Also general cleanup

### DIFF
--- a/terminus-ui/button/src/button.component.scss
+++ b/terminus-ui/button/src/button.component.scss
@@ -88,7 +88,7 @@ $shadow-transition: box-shadow $icon-transition-duration-expand $g-material-shad
 .c-button {
   // Define theme styles
   $themes: primary accent warn;
-  $collapsable-radius: 2em;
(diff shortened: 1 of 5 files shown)


terminus-ui/button/src/button.component.scss
```diff
@@ -155,8 +155,9 @@ $shadow-transition: box-shadow $icon-transition-duration-expand $g-material-shad
     }
   }
 
-  // When collapsable AND disabled
-  &.c-button--collapsable {
+  // When collapsible AND disabled
+  &.c-button--collapsable,
+  &.c-button--collapsible {
```
atlwendy, line 160: `the difference between collapsable and collapsible is that collapsable is while collapsible is that can be collapsed.` lol

atlwendy (approved):

terminus-ui/button/src/button.component.scss
benjamincharity: 🤕

PR #1781, opened 2019-11-05
style(Chart): general cleanup

benjamincharity:
- match our formatting
- fix spelling

### DIFF
--- a/terminus-ui/chart/src/chart.component.ts
+++ b/terminus-ui/chart/src/chart.component.ts
@@ -55,11 +55,8 @@ export type TsChart
 /**
  * This is the chart UI Component
  *
- * FIXME: Once amCharts v4 is TypeScript scrict complient, we should set the `tsconfig.skipLibCheck` to true.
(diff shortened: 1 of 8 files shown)


terminus-ui/chart/src/chart.component.ts
```diff
@@ -102,11 +99,7 @@ export class TsChartComponent implements OnInit, OnChanges, OnDestroy {
    */
   @Input()
   public set visualization(value: TsChartVisualizationOptions) {
-    if (!value) {
-      return;
-    }
-
-    this._visualization = value;
+    this._visualization = value ? value : 'xy';
```
atlwendy, line 102: why did you set to random string `xy` instead of an empty string?

benjamincharity: It's one of the allowed chart visualization types and was already our default chart type. IMO this is just an easier way of reverting to the default when `undefined` is passed in.

PR #1783, opened 2019-11-05
style(Checkbox): general cleanup

benjamincharity:
Also fixed incorrect emitter type

### DIFF
--- a/terminus-ui/checkbox/src/checkbox.component.spec.ts
+++ b/terminus-ui/checkbox/src/checkbox.component.spec.ts
@@ -221,11 +221,11 @@ describe(`TsCheckboxComponent`, function() {
     describe(`set ngModel`, () => {
 
       test(`should set the private variable`, () => {
-        expect(component._isChecked).toEqual(false);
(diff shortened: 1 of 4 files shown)


terminus-ui/checkbox/src/checkbox.component.spec.ts
```diff
@@ -221,11 +221,11 @@ describe(`TsCheckboxComponent`, function() {
     describe(`set ngModel`, () => {
 
       test(`should set the private variable`, () => {
-        expect(component._isChecked).toEqual(false);
+        expect(component['_isChecked']).toEqual(false);
```
atlwendy, line 224: we still have `_` method call 😮

atlwendy (approved):

terminus-ui/checkbox/src/checkbox.component.spec.ts
benjamincharity: Honestly not even sure why we are checking the private var vs the getter.. but this component will be getting refactored _someday_..

PR #1784, opened 2019-11-05
fix(Demo): bump UI library version for demo

benjamincharity:
Also now removing all old demo files

### DIFF
--- a/tooling/ci/demos-deploy.sh
+++ b/tooling/ci/demos-deploy.sh
@@ -19,7 +19,12 @@ if [[ " ${releaseBranches[@]} " =~ " ${CIRCLE_BRANCH} " ]]; then
   # Clone the demo repo
   cd $HOME
   git clone git@github.com:GetTerminus/ui-demos-$CIRCLE_BRANCH.git
+
+  # Enter the demos repo
   cd ui-demos-$CIRCLE_BRANCH
+
+  # Remove old demo files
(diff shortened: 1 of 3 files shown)


tooling/ci/demos-deploy.sh
```diff
@@ -19,8 +19,13 @@ if [[ " ${releaseBranches[@]} " =~ " ${CIRCLE_BRANCH} " ]]; then
   # Clone the demo repo
   cd $HOME
   git clone git@github.com:GetTerminus/ui-demos-$CIRCLE_BRANCH.git
+
+  # Enter the new repo
   cd ui-demos-$CIRCLE_BRANCH
 
+  # Remove old demo files
+  find -E . -regex '.*\.(html|css|js)' -type f -delete
```
atlwendy, line 27: how do you differentiate between old and new files?

benjamincharity: It's all old. The only reason I don't just empty the entire directory is so that our favicon and 3rdpartylicenses file stays intact since I _think_ those aren't in the dist/ automatically

atlwendy: ah okay... it's confusing because the line above comment saying "Enter the new repo", so it's like new repo which contains all old unwanted files...

benjamincharity: Fair enough. Updated the comment to say `Enter the demos repo` for clarity.

PR #1797, opened 2019-11-25
Update compodoc config

benjamincharity:
- Move compodoc config to external file
- Change docs theme for better menu usability
- Docs generation on CI no longer outputs tons of logs
- Don't clean istanbul asset dirs
- Add dev docs around JSDoc tag usage
- Moved JSDom to dev dependency
- Added logo for generated docs
- Fixed compodoc tsconfig path so that no testing components or helpers are included
- Added `@internal` tags to items in `TsAutocomplete` and `TsButton` as examples
- PRs now require demos AoT build to pass


terminus-ui/autocomplete/src/autocomplete.component.ts
```diff
@@ -142,33 +143,44 @@ export class TsAutocompleteComponent implements OnInit,
...
+   *
+   * @internal
    */
   private document: Document;
 
   /**
    * Subject used to alert the parent {@link TsFormFieldComponent} when the label gap should be recalculated
    *
    * Implemented as part of TsFormFieldControl.
+   * @internal
```
atlwendy, line 169: I see sometimes you have a line break before `@internal` sometimes you don't.... does that matter?

terminus-ui/autocomplete/src/autocomplete.component.ts
```diff
@@ -180,26 +192,36 @@ export class TsAutocompleteComponent implements OnInit,
...
    * field. Instead, we pass it manually through the template with this reference.
+   *
+   * @internal
    */
   public selfReference = this;
 
   /*
    * Implemented as part of {@link TsFormFieldControl}
+   *
+   * @internal
```
atlwendy, line 203: why is this `@internal` showing as gray instead of red as elsewhere?

atlwendy (approved):

terminus-ui/autocomplete/src/autocomplete.component.ts
benjamincharity: It's not about the flag itself, but there should always be a line break after the JSDoc definition or the text will be generated on a single line

```typescript
/**
 * Line one
 * this will be generated on line one
 * me too
 */

/**
 * Line one
 *
 * Now I will be line #2 when generated
 */
```

atlwendy: I guess I meant does a line break before the flag needed?
```
/**
 * Line
 * @internal
 */
```
vs.
```
/**
 * Line
 *
 * @interal
 */
```

terminus-ui/autocomplete/src/autocomplete.component.ts
benjamincharity: I have no clue.. testing out docs generation shows that this item is still added to the docs so the flag isn't working as expected in this one spot..

Even changing it to `@ignore` does nothing.. It only seems to work when I create a whole separate JSDoc comment. Which is really odd since these all look identical to me:

Screen Shot 2019-11-26 at 7 52 37 AM

I guess I'm going to leave it as is since it appears that it is working most places ¯\\_(ツ)_/¯

terminus-ui/autocomplete/src/autocomplete.component.ts
benjamincharity: Nope. In your examples it makes no difference if the second line is a flag or plain text. Line #2 must _always_ have a line break.

```typescript
// good
/**
 * Line one
 *
 * Must have a line break for the next piece of docs.
 */

// good
/**
 * Line one
 *
 * @tagOne
 * @tagTwo
 */

// good
/**
 * Line one
 *
 * Line two
 * @tagOne
 */

// bad
/**
 * Line one
 * @tagOne
 * @tagTwo
 */

// bad
/**
 * Line one
 * Line two
 */
```

terminus-ui/autocomplete/src/autocomplete.component.ts
atlwendy: Wow... I had thought it might be a github glitch, but it seems doing something different... interesting..

benjamincharity: Yeah I'm stumped!

PR #1806, opened 2019-12-02
Cleanup cohort date

benjamincharity:
- cohort cleanup
- cohort test helpers now exposed under `/testing`
- added missing commit scopes
- cohort demo now shows selection value

### DIFF
--- a/terminus-ui/cohort-date-range/src/cohort-date-range.component.html
+++ b/terminus-ui/cohort-date-range/src/cohort-date-range.component.html
@@ -1,22 +1,21 @@
-<div>
-  <ts-date-range
-    class="ts-cohort-date-range__date-range"
-    [dateFormGroup]="dateRangeFg"
-    [isDisabled]="!allowCustomDates || isDisabled || disableDateRange"
-    (dateRangeChange)="cohortDateRangeChange($event)"
-  ></ts-date-range>
-  <ts-select
-    label="Select a date range"
-    class="ts-cohort-date-range__select"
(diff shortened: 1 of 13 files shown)


terminus-ui/cohort-date-range/src/cohort-date-range.component.html
```diff
@@ -1,22 +1,21 @@
...
-    [isDisabled]="isDisabled"
-    (selectionChange)="selectionChange($event)"
+<ts-date-range
+  class="ts-cohort-date-range__date-range"
+  [dateFormGroup]="dateRangeFormGroup"
+  [isDisabled]="!allowCustomDates || isDisabled || disableDateRange"
+  (dateRangeChange)="cohortDateRangeChange($event)"
+></ts-date-range>
+
+<ts-select
```
atlwendy, line 8: I guess at some point this would be updated to use `ts-selection-list`?

atlwendy (approved):

terminus-ui/cohort-date-range/src/cohort-date-range.component.html
benjamincharity: Ahh great callout! I thought these were all already converted. I'll do it in this PR 👍

benjamincharity: Went in to do this and realized why this wasn't converted.. currently our selection list doesn't support a non-typed version that the standard select does support.

PR #1820, opened 2019-12-05
1772 selection list truncation bug

benjamincharity:
- fix selection list truncation
- fix selection list disabled state
- add disabled toggle to demo

Screen Shot 2019-12-10 at 8 40 32 AM

### DIFF
--- a/terminus-ui/selection-list/src/selection-list.component.scss
+++ b/terminus-ui/selection-list/src/selection-list.component.scss
@@ -18,6 +18,12 @@ $transition-duration: 200;
       width: 94%;
     }
   }
+
+  &.ts-selection-list--disabled {
+    .ts-selection-list__input {
+      cursor: cursor(not-allowed);
+    }
+  }
 }
 
 .ts-selection-list__input-wrap {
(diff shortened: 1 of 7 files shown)


terminus-ui/selection-list/src/selection-list.component.scss
```diff
@@ -12,6 +12,18 @@ $transition-duration: 200;
 .ts-selection-list {
   cursor: cursor(pointer);
   display: block;
+
+  &.ts-selection-list--single {
+    .ts-selection-list__input {
+      width: 94%;
```
atlwendy, line 18: this width is very precise :)

benjamincharity:
@atlwendy FYI the selection/removal bug you sent me is a pre-existing bug so I split it into it's own ticket: #1850

atlwendy (approved):

terminus-ui/selection-list/src/selection-list.component.scss
benjamincharity: That's a good point. I could do this better by using `calc` to subtract the space used by the icon.

PR #1822, opened 2019-12-06
popover component

### LINKED ISSUE #1696: Popover
## Overview

We need a 'popover' component to support progressive disclosure of complex controls and rich UI. Examples are a rule-builder within a filter panel of a table, optional settings when adding a report to a dashboard etc.

A popover is a floating panel that can contain rich HTML content and is anchored to an existing DOM element. (see visual examples at the bottom)

## Requirements

The goal is to provide as much flexibility around showing/hiding and content while taking care of panel positioning and style for the consumer.

---

Leverage popper.js for the popper functionality: https://github.com/FezVrasta/popper.js

This will consist of 1 directive and 1 component:

```html
<button tsPopoverTrigger="popper1">Click me!</button>

<ts-popover #popper1>
  <h1>My Title</h1>
  <p>Other random content.</p>
</ts-popover>
```

### General

- [ ] Add popover.js as an **optional** dependency
- [ ] When the directive initializes it should verify that popper.js is available. If not we should throw an error so  the consumer knows what is needed.
- [ ] Add tests
- [ ] Add to usage docs
- [ ] Update primary readme with description/links/etc
- [ ] Update component count badge

### Trigger

All consumer settings are applied to `[tsPopoverTrigger]`. `<ts-popover>` is just a wrapper that allows us to consume the popover content.

- [ ] The trigger directive also acts as an input to accept a reference to the popover content.
  - [ ] `@Input('tsPopoverTrigger') popover`:
    - This reference is what is passed as the 'popper' when creating an instance: `new Popper(reference, PASS_IN_HERE, options)`
- [ ] Positions
  - Available positions: https://github.com/FezVrasta/popper.js/blob/master/packages/popper/src/methods/placements.js
  - https://github.com/FezVrasta/popper.js/blob/master/docs/_includes/popper-documentation.md#defaultsplacement
  - `@Input() position: TsPopoverPositions = 'bottom'`
  - Defaults to `bottom` (this is popper.js default also)
  - [ ] This should be a setter so we can verify the position passed in.
  - [ ] `TsPopoverPositions` can be created from https://github.com/FezVrasta/popper.js/blob/master/packages/popper/src/methods/placements.js
  - [ ] If the position is not an allowed position, we should throw a UI library error so the consumer gets information about what is wrong and why (See the Error Example below)
  - I don't believe we need to adjust any of the default popper.js modifiers
- The trigger itself is passed as the instance trigger: `new Popper(this.elementRef.nativeElement, popper, options)`
    - [ ] Creates a new popper `new Popper(reference, popper, options)`
        - https://github.com/FezVrasta/popper.js/blob/master/docs/_includes/popper-documentation.md#new-popperreference-popper-options
    - The only defaults we need to overwrite are the methods for `onCreate` and `onUpdate`
        - https://github.com/FezVrasta/popper.js/blob/master/docs/_includes/popper-documentation.md#defaultsoncreate
- [ ] The trigger should show the popover on `click` or if `ENTER` is pressed
- [ ] Should support events
  - [ ] `popoverCreated` emits the popover instance
    - Fired from `popoverInstance.onCreate`
  - [ ] `popoverUpdated` emits the popover instance
    - Fired from `popoverInstance.onUpdate`
  - [ ] `popoverShown` emits the popover instance
  - [ ] `popoverHidden` emits the popover instance
    - NOTE: We may need to track the shown/hidden status ourselves. I'm not sure popper.js offers this.
- Public methods
  - [ ] Public method `update` should internally call the update method on the popper instance
  - [ ] Public method `show` should show the popover
  - [ ] Public method `hide` should hide the popover
- [ ] Based on the passed in positions, create a popover with popover.js when the trigger is clicked
- [ ] Should allow close by clicking outside the popper
  - [ ] `@Input() hideOnBlur = true`
  - [ ] Default to `true`
  - [ ] Should subscribe to `touchend` on `document` to call `popperInstance.hide()` when true
    - [ ] Should use `TsDocumentService`
- [ ] `@Input id = this.uid`
  - User can define an ID
  - ID defaults to UID (see almost any other component)
  - ID is used for aria label on the popover `[attr.aria-labelledby]="id"`
  - Trigger sets an `id` property in the popover when it is passed to the trigger `@Input`
- [ ] Popover should close when the Escape key is pressed
  - [ ] Should listen for a key event on the popover to call `popperInstance.hide()`
- [ ] Should export itself as `tsPopoverTrigger`
- [ ] The popover should be destroyed in the component's destroy lifecycle
- [ ] Should support being initialized in the open state
  - [ ] `@Input() defaultOpened = false`
  - Default `false`

### Popover

- [ ] Popover accepts rich content via `ng-content`
- [ ] Should export itself as `tsPopover`
- [ ] Should have `@include elevation-element(menu)` mixin
- [ ] Should have `padding: spacing(default)`
  
### Demo

- [ ] Should have a single popover trigger
- [ ] Should have controls to dynamically choose any of the positions (select menu)
- [ ] Popover should have at least some HTML included for demonstration
  

---

##### Error example from the table:

```typescript
if (tsTableColumnAlignmentTypesArray.indexOf(column.alignment) < 0 && isDevMode()) {
  throw new TsUILibraryError(`TsCellDirective: "${column.alignment}" is not an allowed alignment.`
    + `See "TsTableColumnAlignment" type for available options.`);
}
```

---

### Popover Visual Examples

<details>
  <summary>Image examples:</summary>

  Screen Shot 2019-09-17 at 8 40 58 AM
  Screen Shot 2019-09-17 at 8 41 09 AM
  Screen Shot 2019-09-17 at 8 41 17 AM
  Screen Shot 2019-09-17 at 8 41 25 AM
</details>

### Resources

- https://popper.js.org
- https://github.com/FezVrasta/popper.js
- https://github.com/MrFrankel/ngx-popper
- https://stackblitz.com/edit/cdk-popover-example-p1
- https://github.com/angular/components/blob/87992c88be51f71c8baafd44c41421d551141ac0/src/cdk/overlay/position/flexible-connected-position-strategy.ts
- https://ant.design/components/popover/
- https://akveo.github.io/nebular/docs/components/popover/overview#nbpopoverdirective
- https://material.angular.io/cdk/overlay/api
- https://ng-semantic.herokuapp.com/#/elements/popup

atlwendy:
closes #1696


README.md
```diff
@@ -427,6 +429,7 @@ This project follows the [all-contributors](https://github.com/kentcdodds/all-co
 [src-pipes-titleCase]:    https://github.com/GetTerminus/terminus-ui/blob/release/terminus-ui/pipes/src/title-case/title-case.pipe.ts
 [src-pipes-truncate]:     https://github.com/GetTerminus/terminus-ui/blob/release/terminus-ui/pipes/src/truncate/truncate.pipe.ts
 [src-pipes]:              https://github.com/GetTerminus/terminus-ui/blob/release/terminus-ui/pipes/src/
+[src-popover]:            https://github.com/Getterminus/terminus-ui/popover/src/
```
benjamincharity: This URL has the incorrect path

demo/app/components/popover/popover.module.ts
```diff
@@ -0,0 +1,29 @@
...
+    FormsModule,
+    PopoverRoutingModule,
+    ReactiveFormsModule,
+    TsPopoverModule,
+    TsSelectionListModule,
+    TsOptionModule,
+  ],
+  declarations: [
+    PopoverComponent,
+  ],
```
benjamincharity, line 27: Missing exports array here

terminus-ui/popover/package.json
```diff
@@ -0,0 +1,13 @@
+{
+  "ngPackage": {
+    "lib": {
+      "entryFile": "src/public-api.ts",
+      "umdModuleIds": {
+        "@terminus/ngx-tools/browser": "terminus.ngxTools.browser",
+        "@terminus/ngx-tools/type-guards": "terminus.ngxTools.type-guards",
+        "@terminus/ngx-tools/utilities": "terminus.ngxTools.utilities"
+      }
```
benjamincharity, line 11: Missing ngxtools and popper here

 
Screen Shot 2019-12-09 at 10 20 05 AM

terminus-ui/popover/src/popover-options.ts
```diff
@@ -0,0 +1,69 @@
+export type TsPopoverPosition =
```
benjamincharity: Missing JSDoc comments for all items in this file

terminus-ui/popover/src/popover-options.ts
```diff
@@ -0,0 +1,69 @@
...
+  | 'left-start'
+  | 'right-start'
+  | 'top-end'
+  | 'bottom-end'
+  | 'left-end'
+  | 'right-end'
+  | 'auto'
+  | 'auto-start'
+  | 'auto-end';
+
```
benjamincharity: The `=` and the `;` should both be knocked down a line to match our existing types:

```typescript
export type Foo
  = 'foo'
  | 'bar'
;
```

terminus-ui/popover/src/popover-trigger.directive.ts
```diff
@@ -0,0 +1,302 @@
...
+  TsDocumentService,
+  untilComponentDestroyed,
+} from '@terminus/ngx-tools';
+import {
+  ControlValueAccessorProviderFactory,
+  TsUILibraryError,
+} from '@terminus/ui/utilities';
+import {
+  fromEvent,
+  Observable,
```
benjamincharity: I don't think this is used

terminus-ui/popover/src/popover-trigger.directive.spec.ts
```diff
@@ -0,0 +1,171 @@
...
+  createComponent as createComponentInner,
+  dispatchEvent,
+  dispatchKeyboardEvent,
+} from '@terminus/ngx-tools/testing';
+import * as testComponents from '@terminus/ui/popover/testing';
+import { TsPopoverTriggerDirective } from './popover-trigger.directive';
+import { TsPopoverModule } from './popover.module';
+
+describe(`popover trigger`, () => {
+  let fixture: ComponentFixture<any>;
```
benjamincharity: I've started following this pattern so we can remove the `any` in our test fixtures: https://github.com/GetTerminus/terminus-ui/blob/release/terminus-ui/cohort-date-range/testing/src/test-components.ts#L66-L69

terminus-ui/popover/src/popover.module.ts
```diff
@@ -0,0 +1,25 @@
...
+import { TsPopoverComponent } from './popover.component';
+
+export * from './popover.component';
+export * from './popover-trigger.directive';
+export * from './popover-options';
+
+@NgModule({
+  imports: [
+    CommonModule,
+    PlatformModule,
```
benjamincharity: I see this imported everywhere but am not sure I see why? What are we using this for?

terminus-ui/popover/src/popover.module.ts
```diff
@@ -0,0 +1,25 @@
+import { PlatformModule } from '@angular/cdk/platform';
+import { CommonModule } from '@angular/common';
+import { NgModule } from '@angular/core';
```
benjamincharity, line 3: Need line break before internal imports

terminus-ui/popover/src/popover.component.ts
```diff
@@ -0,0 +1,191 @@
...
+  ElementRef,
+  EventEmitter,
+  OnDestroy,
+  Output,
+  ViewChild,
+  ViewContainerRef,
+  ViewEncapsulation,
+} from '@angular/core';
+import { TsUILibraryError } from '@terminus/ui/utilities';
+import Popper from 'popper.js/dist/esm/popper.js';
```
benjamincharity, line 16: We should probably add a note here as to why this custom import is needed. I know 100% I'll forget by the time I need to touch this again.. 🤦‍♂

terminus-ui/popover/src/popover-options.ts
```diff
@@ -0,0 +1,69 @@
...
+}
+
+export type Trigger = 'click';
+
+export class Triggers {
+  public static CLICK: Trigger = 'click';
+}
+
+export interface PopoverOptions {
+  ariaRole?: string;
```
benjamincharity: We should namespace all these items with `TS/ts` like our other interfaces etc

terminus-ui/popover/src/popover-trigger.directive.spec.ts
```diff
@@ -0,0 +1,171 @@
...
+    let testComponent;
+    let popoverNativeElement;
+    let popoverTriggerDebugElement;
+    let buttonDebugElement;
+    let popoverTriggerInstance;
+    let popoverTrigger;
+    let popoverDebugElement;
+    let buttonInstance;
+    let popoverInstance;
+
```
benjamincharity: We should probably add types to all these undefined items. I believe as it is now, they will all by typed as `any`

terminus-ui/popover/src/popover-trigger.directive.spec.ts
```diff
@@ -0,0 +1,171 @@
...
+    function setup(component = testComponents.Basic) {
+      fixture = createComponent(component);
+      fixture.detectChanges();
+      testComponent = fixture.debugElement.componentInstance;
+      popoverTriggerDebugElement = fixture.debugElement.query(By.directive(TsPopoverTriggerDirective));
+      // popoverTriggerDebugElement = fixture.debugElement.query(By.css('.ts-popover-trigger'));
```
benjamincharity: old comments

terminus-ui/popover/src/popover-trigger.directive.spec.ts
```diff
@@ -0,0 +1,171 @@
...
+        fixture.componentInstance.popoverOnShown = jest.fn();
+        fixture.detectChanges();
+        fixture.componentInstance.hideOnBlur = true;
+        fixture.detectChanges();
+        dispatchEvent(buttonDebugElement, event);
+        fixture.detectChanges();
+        tick();
+        expect(fixture.componentInstance.popoverOnShown).toHaveBeenCalled();
+        const outsideElement = fixture.debugElement.query(By.css('.outside')).nativeElement as HTMLElement;
+        dispatchEvent(outsideElement, event);
```
benjamincharity, line 141: fwiw I can usually just call the click method on items directly unless I need to modify the event object. So `outsideElement.click()` should work.

(nothing you need to change, just an FYI that may be faster for you in the future)

terminus-ui/popover/src/popover-trigger.directive.spec.ts
```diff
@@ -0,0 +1,171 @@
...
+        expect(fixture.componentInstance.popoverOnHidden).toHaveBeenCalled();
+      }));
+    });
+
+    describe(`default opened`, () => {
+      test(`should open on load if defaultOpened set to true`, fakeAsync(() => {
+        setup(testComponents.DefaultOpen);
+        fixture.componentInstance.popoverOnShown = jest.fn();
+        tick();
+        expect(popoverDebugElement.nativeElement.getAttribute('x-placement')).toEqual('right');
```
benjamincharity, line 153: I don't think we're `expect`ing the correct thing here. We create a spy above but then check placement here

terminus-ui/popover/src/popover.component.ts
```diff
@@ -0,0 +1,191 @@
...
+      this.visibility = true;
+
+    } else {
+      this.opacity = 0;
+      this.displayType = 'none';
+      this.ariaHidden = 'true';
+      this.visibility = false;
+    }
+    // eslint-disable-next-line dot-notation
+    if (!this.CDR['destroyed']) {
```
benjamincharity, line 194: We can ignore the 'else' path here for test coverage

terminus-ui/popover/src/popover.component.ts
```diff
@@ -0,0 +1,191 @@
...
+   */
+  public scheduleUpdate(): void {
+    this.popoverInstance && this.popoverInstance.scheduleUpdate();
+  }
+
+  /**
+   * Based on visibility, adjust visual settings
+   * @param visibility - boolean
+   */
+  public toggleVisibility(visibility: boolean) {
```
benjamincharity: Missing return type.

terminus-ui/popover/src/popover.component.ts
```diff
@@ -0,0 +1,191 @@
...
+
+  /**
+   * When popover is hidden, destroy the instance, toggle visibility to false and emit onHidden event.
+   */
+  public hide() {
+    if (this.popoverInstance) {
+      this.popoverInstance.destroy();
+    }
+    this.toggleVisibility(false);
+    this.onHidden.emit();
```
benjamincharity, line 143: It doesn't seem like we're actually testing these two lines (only have coverage for the if statement)

terminus-ui/popover/src/popover.component.ts
```diff
@@ -0,0 +1,191 @@
...
+
+  constructor(
+    private viewRef: ViewContainerRef,
+    private CDR: ChangeDetectorRef,
+  ) {
+    /**
+     * Check whether popper.js has been properly imported.
+     */
+    if (!Popper) {
+      throw new TsUILibraryError('popper.js is not available to reference.');
```
benjamincharity: We should prefix this error message with the component name `TsPopoverComponent: My message`

terminus-ui/popover/src/popover.component.ts
```diff
@@ -0,0 +1,191 @@
...
+  @Output()
+  public readonly onUpdate = new EventEmitter();
+
+  /**
+   * Emit when it's hidden.
+   */
+  // tslint:disable-next-line:no-output-on-prefix
+  @Output()
+  public readonly onHidden = new EventEmitter();
+
```
benjamincharity: Missing type for the emitter content

terminus-ui/popover/src/popover.component.ts
```diff
@@ -0,0 +1,191 @@
...
+   */
+  public myId = '';
+
+  /**
+   * options needed for popper.js
+   */
+  public popoverOptions: PopoverOptions = {
+    position: TsPopoverPositions.Bottom,
+    trigger: Triggers.CLICK,
+  } as PopoverOptions;
```
benjamincharity: Curious why we need to coerce the type here using `as`? It seems we have a couple paths forward:

a) If these items are legitimate `PopoverOptions` they should be added to the interface
b) If `PopoverOptions` should allow other random items, we should make the type accept other random keys

terminus-ui/popover/src/popover.component.ts
```diff
@@ -0,0 +1,191 @@
...
+
+  /**
+   * Whether the element is visible
+   */
+  public visibility = false;
+
+  /**
+   * id set by trigger and pass to popover element
+   */
+  public myId = '';
```
benjamincharity: We usually only use the `my` prefix for demo items so users clearly see what items they should be defining. Internally this should probably just be `id`.

terminus-ui/popover/src/popover.component.ts
```diff
@@ -0,0 +1,191 @@
...
+
+  /**
+   * The opacity of popover element
+   */
+  public opacity = 0;
+
+  /**
+   * aria hidden attribute
+   */
+  public ariaHidden = 'true';
```
benjamincharity, line 59: We should add a note as to why we need to save this boolean as a string for future reference

terminus-ui/popover/src/popover.component.html
```diff
@@ -0,0 +1,15 @@
+<div
+  #popoverViewRef
+  class="popover__container"
+  [style.display]="displayType"
+  [style.opacity]="opacity"
```
benjamincharity: Any specific reason we are inlining these styles instead of adding/removing classes? I'd love to keep our inline styles as minimal as possible

terminus-ui/popover/src/popover-trigger.directive.ts
```diff
@@ -0,0 +1,302 @@
...
+ *              [defaultOpened]="defaultOpened"
+ *              (popoverOnShown)="popoverOnShown()"
+ *              (popoverOnHidden)="popoverOnHidden()"
+ * />
+ */
+@Directive({
+  selector: '[tsPopoverTrigger]',
+  host: { class: 'ts-popover-trigger' },
+  providers: [
+    ControlValueAccessorProviderFactory<TsPopoverTriggerDirective>(TsPopoverTriggerDirective),
```
benjamincharity: Do we need this? Usually this is just for items that are implementing a custom ControlValueAccessor

terminus-ui/popover/src/popover-trigger.directive.ts
```diff
@@ -0,0 +1,302 @@
...
+   */
+  private document: Document;
+
+  /**
+   * When no id provided, it falls back to uid
+   */
+  public readonly uid = `ts-popover-trigger-${nextUniqueId++}`;
+
+  /**
+   * Some default options for popper. For now we only support click, so it specifies click here. But could support more if needed.
```
benjamincharity: We should split this note into two lines. First line is the one sentence description and any explanations needed should be below.

```typescript
  /**
   * My thing.
   *
   * It does this. But not that.
   */
```

This will keep the 1 sentence name elevated in the docs for better skimming.

terminus-ui/popover/src/popover-trigger.directive.ts
```diff
@@ -0,0 +1,302 @@
...
+
+  /**
+   * Set position of where popover opens
+   */
+  @Input()
+  public set position(value: TsPopoverPosition) {
+    if (tsPopoverPositions.indexOf(value) < 0) {
+      throw new TsUILibraryError(`"${value}" is not an allowed position value.`);
+    }
+    this._position = isUnset(value) ? TsPopoverPositions.Bottom : value;
```
benjamincharity: Since `''` and `0` are not valid values we can simplify this to just

```typescript
this._position = value || TsPopoverPositions.Bottom;
```

terminus-ui/popover/src/popover-trigger.directive.ts
```diff
@@ -0,0 +1,302 @@
...
+      this.show();
+    }
+    if (this.popover) {
+      this.popover.myId = this.id;
+    }
+  }
+
+  /**
+   * Called on every change happens
+   * @param changes - SimpleChange
```
benjamincharity: I think we can better describe what is happening here. Currently we are just describing what the lifecycle event does.

Also, missing a line break before the params.

terminus-ui/popover/src/popover-trigger.directive.ts
```diff
@@ -0,0 +1,302 @@
...
+        untilComponentDestroyed(this)
+      ).subscribe(event => {
+        this.toggle();
+        event.stopPropagation();
+        event.preventDefault();
+      });
+  }
+
+  /**
+   * Called after the default change detector has completed checking all content
```
benjamincharity: I think we can better describe what is happening here. Currently we are just describing what the lifecycle event does.

terminus-ui/popover/src/popover-trigger.directive.ts
```diff
@@ -0,0 +1,302 @@
...
+      .subscribe(e => this.onPopoverUpdate(e));
+  }
+
+  /**
+   * Toggle between show and hide.
+   */
+  public toggle(): void {
+    this.isOpen ? this.hide() : this.show();
+    this.changeDetectorRef.markForCheck();
+    this.changeDetectorRef.detectChanges();
```
benjamincharity: Do we really need both of these?

terminus-ui/popover/src/popover-trigger.directive.ts
```diff
@@ -0,0 +1,302 @@
...
+  public static defaultOptions: PopoverOptions = {
+    trigger: Triggers.CLICK,
+    ariaRole: 'popover',
+    placement: TsPopoverPositions.Bottom,
+  };
+
+  /**
+   * Whether current popover is open or not.
+   */
+  public isOpen = false;
```
benjamincharity, line 82: I see us setting this flag, but is it ever used anywhere? Just wondering if we can use a better method to know if the popper is open or not. It seems that we should be able to get this information from either Popper.js or our popper component (since it knows when it's opened to fire the emitters). Just hoping for a less manual way to know since currently we set the flag without actually knowing if the popper was opened.

terminus-ui/popover/src/popover-trigger.directive.ts
```diff
@@ -0,0 +1,302 @@
...
+    this.isOpen = true;
+    this.popover.show(this.popover.popoverOptions);
+    this.popoverOnShown.emit(this.popover.popoverInstance);
+  }
+
+  /**
+   * Emit popoverOnUpdate when it's updated.
+   * @param event
+   */
+  private onPopoverUpdate(event): void {
```
benjamincharity: We only call this method in one place so we can probably just move the emitter up to where it's used

benjamincharity: If we do want to leave this method we need to add a type for the event param

benjamincharity: And add the missing space between the doc comment and the doc param lines

terminus-ui/popover/src/popover.component.md
```diff
@@ -0,0 +1,95 @@
...
+  - [Hide on Blur](#hide-on-blur)
+  - [Open on load](#open-on-load)
+- [Events](#events)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+## Basic info
+
+We wrap our popover trigger/component with a third party library, named [popper.js](https://github.com/popperjs/popper.js).
+`popper.js` is not a required import for ui library but if it's not imported and consumers try to use popover component, an error will show up.
```
benjamincharity: Good callout here. We should also add the install instructions and a link to the popper.js docs.

Also, this just reminded me, we should add Popper.js to the `ACKNOWLEDGEMENT.md` file

terminus-ui/popover/src/popover.component.md
```diff
@@ -0,0 +1,95 @@
...
+  [position]="top"
+>Click me!</button>
+
+<ts-popover #popper1>
+  <h1>My Title</h1>
+  <p>Other random content.</p>
+</ts-popover>
+```
+
+It defaults to `bottom`. For all available positions, please reference [here](https://github.com/FezVrasta/popper.js/blob/master/packages/popper/src/methods/placements.js).
```
benjamincharity: Let's keep all our links defined at the bottom of the file to follow existing patterns. Probably should change the link text to something like `please reference the [popper.js documentation]` etc

terminus-ui/popover/src/popover.component.md
```diff
@@ -0,0 +1,95 @@
...
+  [hideOnBlur]="false"
+>Click me!</button>
+
+<ts-popover #popper1>
+  <h1>My Title</h1>
+  <p>Other random content.</p>
+</ts-popover>
+```
+
+It defaults to `true`.
```
benjamincharity: Since we default to `true` we should probably flip these docs to document the `false` functionality since it closes on blur by default.

terminus-ui/popover/src/popover.component.scss
```diff
@@ -0,0 +1,77 @@
+@import './../../scss/helpers/color';
+@import './../../scss/helpers/shadows';
+@import './../../scss/helpers/spacing';
+
+$PLACEMENT_MARGIN: 5px;
```
benjamincharity: I don't believe SCSS files get isolated scope, so we should probably move this variable just inside the first SCSS scope (in this instance `popover__container`).

terminus-ui/popover/src/popover.component.scss
```diff
@@ -0,0 +1,77 @@
...
+    border-color: color(pure);
+    border-style: solid;
+    height: 0;
+    position: absolute;
+    width: 0;
+  }
+
+  &[x-placement^='top'] {
+    margin-bottom: $PLACEMENT_MARGIN;
+    .popover__arrow {
```
benjamincharity, line 35: No need to change it, but for future reference we have a mixin to help create these triangles easily

terminus-ui/popover/src/popover.component.scss
```diff
@@ -0,0 +1,77 @@
...
+    margin-right: $PLACEMENT_MARGIN;
+    .popover__arrow {
+      border-bottom-color: transparent;
+      border-right-color: transparent;
+      border-top-color: transparent;
+      border-width: 5px 0 5px 5px;
+      margin-left: 0;
+      margin-right: 0;
+      right: -5px;
+      top: calc(50% - 5px);
```
benjamincharity: We should probably be using our 5px variable so all these don't need to be manually updated if our sizing changes. (your editor should be yelling at you about these 'magic numbers')

benjamincharity (changes requested):
Looking great! Mostly non functional items in review 👍

A couple other callouts I couldn't attach to a file:

1. I believe we need to add popover to this file: https://github.com/GetTerminus/terminus-ui/blob/release/terminus-ui/tsconfig.spec.json (feel free to fix the 1 line with incorrect indention also 😬 )
2. We need to add Popover as a new scope for cz-config: https://github.com/GetTerminus/terminus-ui/blob/release/tooling/cz-config.js

terminus-ui/popover/src/popover.component.scss
atlwendy: I totally expected lint or editor would be yelling at these random numbers, but neither of them complained, so I didn't update lol

benjamincharity: Dang... you may need to verify your IDE is using our rules then cause I _definitely_ get yelled at 😂

demo/app/components/popover/popover.module.ts
atlwendy: What's the exports array you're referring here?

benjamincharity: The NgModule exports array. Check any other component's module file, they all have it

atlwendy: closing this out as demo doesn't have an exports array :)

terminus-ui/popover/src/popover-trigger.directive.spec.ts
atlwendy: Historically using `.click()` for some reason in some tests wouldn't fire off the click event, but worked fine with `dispatchEvent`. So I've been using that, maybe should try again `.click()`

terminus-ui/popover/src/popover-trigger.directive.spec.ts
atlwendy: That mock call isn't necessary. But expect is expecting correctly - default open set to true, popover opens on load.

terminus-ui/popover/src/popover-trigger.directive.ts
atlwendy: This flag is only used internally by trigger... mainly so toggle would know to toggle it to open or hide. I had considered  making it subscribing to the event emitter, but thought it's an overkill since it's not trying to keep track of status between component/directive. popover only opens on load when `defaultOpened` set to `true`, where `ngAfterContentCheck` will check. I think it works well for our current case, but if we consider supporting more complex case down the road, I agree that subscribing to an event will make more sense.

terminus-ui/popover/src/popover-trigger.directive.spec.ts
benjamincharity: Ah no worries.. just trying to save you some keystrokes!

terminus-ui/popover/src/popover-trigger.directive.spec.ts
benjamincharity: > But expect is expecting correctly - default open set to true, popover opens on load.

Sorry, can you clarify this a bit for me? When I read this test I read that it should verify the popover is open by default. But the actual test seems to be verifying the position rather than then existence? Shouldn't we just be seeing if some parent class/element exists in the dom?

atlwendy: Ah I see the confusion... so the popover element always exists. But display is hidden by default. When popover shows up, that means option set (position etc) and that `x-placement` attribute assigned. So by checking that `x-placement`, it not only makes sure popover display shows up, but also the position is set correctly (based on input). It probably could have worded better. I might change it to check `c-popover__visible` shown up first, then that placement set correctly.

benjamincharity: Ahhh I see

Yeah my vote would be to check for a visible class or something.. I could just see someone removing this test cause it looks like it is duplicating a positioning test etc

terminus-ui/popover/src/popover.component.ts
atlwendy: My coverage shows that those two lines are covered... although I don't think they are explicitly tested, but should be covered from running `.hide()`.

benjamincharity: > My coverage

Are you referring to your IDE coverage or the `/coverage` directory output? I was looking at the directory output, but something may have been updated. I can check again.

atlwendy: Both my IDE and `/coverage` output (I think webstorm reads out from that directory) show that these two lines are covered. The only line not covered in this file is that popper.js import checking and throwing UI library error.

benjamincharity: I see, I was seeing the uncovered branch rather than lines. I think we can just add the comment to ignore that else path for coverage 👍

terminus-ui/popover/src/popover-trigger.directive.ts
```diff
@@ -0,0 +1,304 @@
...
+  @Output()
+  public readonly popoverOnUpdate = new EventEmitter<TsPopoverComponent>();
+
+  /**
+   * Register the click to decide whether it's inside the element or out.
+   *
+   * @param targetElement
+   */
+  @HostListener('document:click', ['$event.target'])
+  public onClick(targetElement) {
```
benjamincharity: Missing type

terminus-ui/popover/src/popover-trigger.directive.ts
benjamincharity: ^^

terminus-ui/popover/src/popover-options.ts
```diff
@@ -0,0 +1,89 @@
...
+ * Define Trigger type.
+ */
+export type Trigger = 'click';
+
+/**
+ * A class of Triggers.
+ */
+export class Triggers {
+  public static CLICK: Trigger = 'click';
+}
```
benjamincharity: Still have two items that aren't namespaced

terminus-ui/popover/src/popover.component.ts
```diff
@@ -0,0 +1,199 @@
...
+   */
+  @ViewChild('popoverViewRef', { static: true })
+  public popoverViewRef!: ElementRef;
+
+  /**
+   * Emit when it's updated.
+   */
+  // tslint:disable-next-line:no-output-on-prefix
+  @Output()
+  public readonly onUpdate = new EventEmitter();
```
benjamincharity: Missing emitter type

terminus-ui/popover/src/popover.component.html
```diff
@@ -0,0 +1,17 @@
+<div
+  #popoverViewRef
+  class="popover__container"
+  [ngClass]="{
+    'c-popover__visible': visibility,
+    'c-popover__hidden': !visibility
```
benjamincharity: These classes are using the 'element' style rather than the 'modifier' style

eg

```
.car {}
.car__door {}
.car--blue {}
```

benjamincharity (changes requested):

terminus-ui/popover/src/popover-trigger.directive.ts
```diff
@@ -0,0 +1,295 @@
...
+  @Output()
+  public readonly popoverOnUpdate = new EventEmitter<TsPopoverComponent>();
+
+  /**
+   * Register the click to decide whether it's inside the element or out.
+   *
+   * @param targetElement
+   */
+  @HostListener('document:click', ['$event.target'])
+  public onClick(targetElement): void {
```
benjamincharity: Looks like we're still missing the param type here

benjamincharity (changes requested):

benjamincharity (approved):

PR #1876, opened 2019-12-17
fix(Chart): correctly dispose of the chart when needed

### LINKED ISSUE #1873: Chart: Dispose of chart before reinitializing in ngOnChanges
Currently we are re-initializing the chart if the visualization changes. We need to correctly dispose of the chart before calling `init` again.

- [ ] Fix issue
- [ ] Add tests

benjamincharity:
ISSUES CLOSED: #1873

- moved test components to `test` endpoint
- added test helpers
- refactor so chart is disposed of properly before initialization

### DIFF
--- a/terminus-ui/chart/src/chart.component.ts
+++ b/terminus-ui/chart/src/chart.component.ts
@@ -6,7 +6,6 @@ import {
   ElementRef,
   EventEmitter,
   Input,
-  isDevMode,
   NgZone,
   OnChanges,
   OnDestroy,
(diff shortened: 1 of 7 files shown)


terminus-ui/chart/src/chart.component.ts
```diff
@@ -173,9 +183,12 @@ export class TsChartComponent implements OnInit, OnChanges, OnDestroy {
   /**
    * Initialize a chart
    */
-  private init(type: TsChartVisualizationOptions): void {
+  private createChart(type: TsChartVisualizationOptions): void {
     this.zone.runOutsideAngular(() => {
-      // Create the appropriate chart using a chained ternary
+      if (this.chart) {
+        this.chart.dispose();
```
atlwendy, line 189: Does this ensure that old chart gets destroyed before creating new one?

atlwendy (approved):

terminus-ui/chart/src/chart.component.ts
benjamincharity: yeah, if the chart exists, we call the chart's dispose method.

PR #1895, opened 2020-01-09
fix(Chip): chip takes chip collection inputs into consideration..

### LINKED ISSUE #1850: Chip: Collection doesn't respect existing `isRemovable` when syncing state
#### 1. What is the expected behavior?

Changing the `isSelectable` flag should not affect `isRemovable`.

#### 2. What is the current behavior?

If the collection is set to *not* allow removal, and _then_ set to not allow selection, the collection syncs the chips state and overrides the `isRemovable` flag.

https://github.com/GetTerminus/terminus-ui/blob/release/terminus-ui/chip/src/chip-collection.component.ts#L659

#### 3. What are the steps to reproduce?

a) Go to our demos: https://getterminus.github.io/ui-demos-release/components/chip
b) Uncheck `allow removal`
c) Notice remove icons disappear
d) Uncheck `allow selection`
e) Notice the remove icons reappear

#### 4. Which versions of this library, Angular, TypeScript, & browsers are affected?

- UI Library: 15.6.1

atlwendy:
closes #1850

### DIFF
--- a/demo/app/components/chip/chip.component.html
+++ b/demo/app/components/chip/chip.component.html
@@ -10,20 +10,25 @@
       </select>
     </label>
     <br>
-    <label>
-      Allow removal:
-      <input type="checkbox" [(ngModel)]="removable">
-    </label>
     <br>
+    <div tsVerticalSpacing>
+      <label>
+        Chip Collection Disabled:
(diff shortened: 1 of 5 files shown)


benjamincharity (approved):
🥇

PR #1899, opened 2020-01-10
1893 cohortdaterange default cohort

benjamincharity:
- Selection list now supports static trigger
- Cohort now supports a default cohort
- Fixed lint staged error for SCSS files
- Fixed incorrect cohort docs
- Added a utilities SCSS file for utility classes

### DIFF
--- a/terminus-ui/cohort-date-range/src/cohort-date-range.component.ts
+++ b/terminus-ui/cohort-date-range/src/cohort-date-range.component.ts
@@ -12,25 +12,32 @@ import {
 import {
   AbstractControl,
   FormBuilder,
+  FormControl,
 } from '@angular/forms';
-import { untilComponentDestroyed } from '@terminus/ngx-tools';
 import { coerceDateProperty } from '@terminus/ngx-tools/coercion';
+import { untilComponentDestroyed } from '@terminus/ngx-tools/utilities';
 import { TsOption } from '@terminus/ui/option';
 import {
-  TsSelectChange,
-  TsSelectComponent,
-} from '@terminus/ui/select';
+  TsSelectionListChange,
+  TsSelectionListComponent,
+} from '@terminus/ui/selection-list';
 
 
 /**
- * Represents date cohort object that passed in
+ * Represents the range for a {@link TsDateCohort}
+ */
+export interface TsDateCohortRange {
+  start: string | Date;
+  end: string | Date;
+}
+
+/**
+ * Represents the date cohort object that passed in
  */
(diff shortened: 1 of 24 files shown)


terminus-ui/cohort-date-range/src/cohort-date-range.component.ts
```diff
@@ -125,24 +133,61 @@ export class TsCohortDateRangeComponent implements OnInit, OnDestroy {
...
+    if (activeCohort.length) {
+      this.cohortControl.setValue(activeCohort);
+      this.setDateRangeValues(activeCohort[0].range);
+    }
+  }
+  public get cohorts(): ReadonlyArray<TsDateCohort> {
+    return this._cohorts;
+  }
+  private _cohorts: TsDateCohort[];
+  private originalCohorts: ReadonlyArray<TsDateCohort>;
```
atlwendy, line 190: this `originalCohorts` doesn't need a default value?

terminus-ui/scss/helpers/_layout.scss
```diff
@@ -120,5 +119,6 @@
   padding: 0;
   pointer-events: none;
   position: absolute;
+  visibility: hidden;
```
atlwendy, line 122: wonder how this mixin ever worked without setting visibility...

atlwendy (approved):
such a long route to just enable setting a default value! didn't realize `ts-select` didn't support values being objects.

benjamincharity:
> such a long route to just enable setting a default value! didn't realize ts-select didn't support values being objects.

Definitely surprised me.. I spent forever looking thinking I was missing it. But I think maybe Dean added that ability to selection-list ¯\_(ツ)_/¯

terminus-ui/cohort-date-range/src/cohort-date-range.component.ts
benjamincharity: I don't believe so. It's set anytime the cohorts are set and is only used if the `allowUserInput` setter is called.

terminus-ui/scss/helpers/_layout.scss
benjamincharity: It _almost_ worked in my case without it since there is such a tiny tiny area that is visible. You could _just barely_ see a pixel of stuff when I tried, so I just went ahead and added it.

The reason it didn't originally have that is because setting visibility to hidden removes the element from the accessibility tree which goes against the primary use of the mixin (to hide things visually that need to be there for accessibility).

But getting this out is more important since we don't actually follow strict accessibility here.

PR #1934, opened 2020-01-16
fix(SelectionList): selected option should be scrolled into view when panel opened

### LINKED ISSUE #1738: SelectionList: If the selected option exists in the options collection it should be scrolled into view when opened
When opening the selection list, it should scroll the selected option into view if it exists. If multiple selections have been made, the last one should be scrolled into view.

- [ ] Extend/use functionality to scroll selected option into view
  - Can possibly find example code in `TsSelect`
- [ ] Add test

atlwendy:
closes #1738

### DIFF
--- a/terminus-ui/selection-list/src/selection-list.component.spec.ts
+++ b/terminus-ui/selection-list/src/selection-list.component.spec.ts
@@ -974,4 +974,20 @@ describe(`TsSelectionListComponent`, function() {
   //   Waiting for https://github.com/jsdom/jsdom/issues/317 to land
   test.todo(`should select all existing text when the input is focused`);
 
+  describe(`should scroll populated item into view`, () => {
+    let fixture: ComponentFixture<testComponents.SeededSingleSelect>;
+    let instance: TsSelectionListComponent;
+
+    beforeEach(() => {
+      fixture = createComponent(testComponents.SeededSingleSelect);
+      fixture.detectChanges();
+      instance = getSelectionListInstance(fixture);
+    });
+    test(`should have panel open with the preset value`, fakeAsync(() => {
+      instance.trigger.handleFocus();
+      tick(100);
+      expect(instance.panel.keyManager.activeItemIndex).toEqual(6);
+    }));
+  });
+
 });
--- a/terminus-ui/selection-list/src/selection-list.component.ts
+++ b/terminus-ui/selection-list/src/selection-list.component.ts
@@ -36,6 +36,7 @@ import {
 } from '@terminus/ui/chip';
 import { TsFormFieldControl } from '@terminus/ui/form-field';
 import {
+  getOptionScrollPosition,
   TS_OPTION_PARENT_COMPONENT,
   TsOptgroupComponent,
   TsOptionComponent,
@@ -59,7 +60,10 @@ import {
   TsSelectionListPanelComponent,
   TsSelectionListPanelSelectedEvent,
 } from './selection-list-panel/selection-list-panel.component';
-import { TsSelectionListTriggerDirective } from './selection-list-panel/selection-list-trigger.directive';
+import {
+  SELECTION_LIST_PANEL_MAX_HEIGHT,
+  TsSelectionListTriggerDirective,
+} from './selection-list-panel/selection-list-trigger.directive';
 
 
 // Unique ID for each instance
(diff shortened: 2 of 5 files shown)


terminus-ui/selection-list/src/selection-list.component.spec.ts
```diff
@@ -974,4 +974,21 @@ describe(`TsSelectionListComponent`, function() {
...
+    beforeEach(() => {
+      fixture = createComponent(testComponents.SeededSingleSelect);
+      fixture.detectChanges();
+      instance = getSelectionListInstance(fixture);
+    });
+    test(`should have panel open with the preset value`, () => {
+      instance.trigger.handleFocus();
+      setTimeout(() => {
+        expect(instance.panel.keyManager.activeItemIndex).toEqual(6);
+      }, 20);
```
benjamincharity: I don't believe need timeouts in our tests, we should be able to use fakeAsync and tick(20)

benjamincharity: Also, not sure the `beforeEach` is gaining us anything, but ¯\_(ツ)_/¯

terminus-ui/selection-list/src/selection-list.component.ts
```diff
@@ -607,6 +611,14 @@ export class TsSelectionListComponent implements
     this.trigger.selectionListPanel.opened.pipe(untilComponentDestroyed(this)).subscribe(() => {
       this.opened.emit();
     });
+    // Wire up listeners for panel opened event
+    this.opened.pipe(untilComponentDestroyed(this)).subscribe(() => {
+      // NOTE: setTimeout is added to resolve a timing issue. Promise.resolve().then does not work in this case.
+      // Without setTimeout when opened event emitted, panel is still undefined.
+      setTimeout(() => {
+        this.scrollActiveOptionIntoView();
+      }, 0);
```
benjamincharity: `0` doesn't need to be defined as that is the default value

terminus-ui/selection-list/testing/src/test-components.ts
```diff
@@ -309,6 +309,36 @@ export class Seeded {
   }
 }
 
+@Component({
+  template: `
+    <ts-selection-list
+      [formControl]="myCtrl"
+      [reopenAfterSelection]="keepOpen"
```
benjamincharity: this is `false` by default

terminus-ui/selection-list/testing/src/test-components.ts
```diff
@@ -309,6 +309,36 @@ export class Seeded {
...
+    </ts-selection-list>
+  `,
+})
+export class SeededSingleSelect {
+  public states: State[] = STATES.slice();
+  public myCtrl = new FormControl([STATES[6]]);
+  public keepOpen = false;
+  public formatter: TsSelectionListFormatter | undefined;
+  public setFormatter() {
+    this.formatter = v => (v as State).population;
```
benjamincharity: Do we use this dynamic formatter setter anywhere? Or can we just statically define the formatter func?

Also, the `as State` coercion shouldn't be necessary if we type the param:

```typescript
formatter(state: State) {
  return state.population;
}
```

terminus-ui/selection-list/src/selection-list.component.ts
```diff
@@ -927,4 +939,22 @@ export class TsSelectionListComponent implements
     this.inputElement.nativeElement.value = '';
   }
 
+  /**
+   * Scroll the active option into view
+   */
+  private scrollActiveOptionIntoView(): void {
+    const ctrlValue = this.selectionListFormControl.value || [];
```
benjamincharity: Is it possible that the control value is not an array? I know we default to an array during initialization..

terminus-ui/selection-list/src/selection-list.component.ts
```diff
@@ -927,4 +939,22 @@ export class TsSelectionListComponent implements
...
+    this.trigger.selectionListPanel.scrollTop = getOptionScrollPosition(
+      index,
+      this.trigger.itemHeight,
+      this.trigger.selectionListPanel.scrollTop,
+      SELECTION_LIST_PANEL_MAX_HEIGHT,
+    );
+    // Only update the focus when it's not autocomplete
+    if (!this.allowUserInput) {
+      this.trigger.selectionListPanel.keyManager.setActiveItem(index);
+    }
```
benjamincharity: Doesn't look like this is tested:

Screen Shot 2020-01-16 at 10 28 14 AM

benjamincharity (changes requested):
Looks pretty good. Tested on a single select and it worked great!

It doesn't seem to work when there are multiple selections though:
selectionlistscroll

benjamincharity (approved):

PR #1957, opened 2020-01-23
fix(SelectionList): form control value updated properly when set to empty array

### LINKED ISSUE #1956: ts-select-list is not properly clearing previously selected values
#### 1. What is the expected behavior?

Using setValue or patchValue on a form control which is assigned to a ts-select-list component with an empty array as the argument should clear all selections.

#### 2. What is the current behavior?

Previously selected options will disappear visually, and the form control value will also be empty. However, once a new value is selected, all previously selected values will now be selected along with the new value.

#### 3. What are the steps to reproduce?

With a ts-select-list that either allows multiple selections or allows single selection and user input, select one or more values and then programmatically clear selections from the associated form control using either setValue or patchValue. In the case of the single select, the previous value will still be present. In the case of the multi select, simply make a new selection.

Providing a reproduction is the *best* way to share your issue.

a) Fork our starter repo: https://github.com/GetTerminus/ui-stackblitz-starter
b) Replicate your issue.
c) Add a link to your replication or attach a zip of your project to this issue.
d) Outline any steps needed to reproduce the issue.

#### 4. Which versions of this library, Angular, TypeScript, & browsers are affected?

- UI Library: v15.10.5 (might be all of v15?)
- Angular: 8+
- TypeScript: 3.5.3
- Browser(s): Chrome

atlwendy:
closes #1956

### DIFF
--- a/terminus-ui/selection-list/src/selection-list.component.spec.ts
+++ b/terminus-ui/selection-list/src/selection-list.component.spec.ts
@@ -848,6 +848,36 @@ describe(`TsSelectionListComponent`, function() {
     }));
   });
 
+  test(`should be able to clear all options programmatically`, fakeAsync(() => {
+    const fixture = createComponent(testComponents.Seeded);
+    fixture.componentInstance.allowDuplicates = true;
+    fixture.detectChanges();
+    let chips = getAllChipInstances(fixture);
+    expect(chips.length).toEqual(1);
+    let instance = getSelectionListInstance(fixture);
+    instance.selectionListFormControl.setValue([]);
+    tick(1000);
+    fixture.detectChanges();
+    chips = getAllChipInstances(fixture);
+    expect(chips.length).toEqual(0);
+
+    const input = getSelectionListInput(fixture);
+    const states = fixture.componentInstance.states;
+    typeInElement(states[5].name.substring(0, 2), input);
+    tick(1000);
+    fixture.detectChanges();
+
+    const opt = getOptionElement(fixture, 0, 5);
+    opt.click();
+    tick(1000);
+    fixture.detectChanges();
+
+    chips = getAllChipInstances(fixture);
+    expect(chips.length).toEqual(1);
+    instance = getSelectionListInstance(fixture);
+    expect(instance.selectionListFormControl.value).toEqual([states[5]]);
+  }));
+
   describe(`clearPreviousSelectedOption`, () => {
     let fixture: ComponentFixture<testComponents.Seeded>;
     let trigger: TsSelectionListTriggerDirective;
--- a/terminus-ui/selection-list/src/selection-list.component.ts
+++ b/terminus-ui/selection-list/src/selection-list.component.ts
@@ -561,11 +561,9 @@ export class TsSelectionListComponent implements
           .subscribe(newValue => {
             // istanbul ignore else
             if (newValue) {
-              if (!newValue[0]) {
-                return;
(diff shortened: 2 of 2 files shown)


terminus-ui/selection-list/src/selection-list.component.ts
```diff
@@ -561,12 +561,10 @@ export class TsSelectionListComponent implements
           .subscribe(newValue => {
             // istanbul ignore else
             if (newValue) {
-              if (!newValue[0]) {
-                return;
+              if (newValue[0]) {
+                this.searchQuery = this.displayFormatter(newValue[0]);
```
benjamincharity, line 565: With this in place, the value will get set twice. Once here and once on line 571. Seems like we probably just want to remove this internal `if` on line 564?

benjamincharity (dismissed):

terminus-ui/selection-list/src/selection-list.component.spec.ts
```diff
@@ -875,6 +875,27 @@ describe(`TsSelectionListComponent`, function() {
...
+      typeInElement(fixture.componentInstance.states[3].name.substring(0, 3), input);
+      tick(1000);
+      fixture.detectChanges();
+
+      const opt = getOptionElement(fixture, 0, 0);
+      opt.click();
+      tick(1000);
+      fixture.detectChanges();
+
+      const chips = getAllChipInstances(fixture);
```
benjamincharity: Out of curiosity, any reason we check the number of selected options at the beginning but check chips at the end?

terminus-ui/selection-list/src/selection-list.component.ts
atlwendy: Here is only to set the `searchQuery` instead of form control value, and line 571 sets the form control value, right?
We cannot remove the `if` statement, because when it's an empty array, you can't set format on that. This change is just to let empty array go through. Previously, if new value is an empty array, the stream stops, so `selectionListFormControl` value never gets set properly.

terminus-ui/selection-list/src/selection-list.component.spec.ts
atlwendy: The latter one was simply trying to mock how the bug was caught. Guess we could also check chips at beginning too..

terminus-ui/selection-list/src/selection-list.component.ts
benjamincharity: Maybe I'm unclear on the extent of the bug you are solving? The issue calls out the value of the form control is not updated correctly. I get why removing the return was needed to solve that, but I'm unclear why the searchQuery was edited?

atlwendy: The searchQuery needs to be there for all the new value except when it's an empty array. Previously, because when it's an empty array, the stream returns, so no `if` condition needed for setting searchQuery. Now we don't want to stop the stream just because it's an empty array, thus this `if` is needed because there is no point to set formatter on an empty array. 
The issue is about setting an empty array on a form control.

benjamincharity (changes requested):
Test is not passing

atlwendy:
@benjamincharity I noticed the test is flaky on ci, sometimes it passed sometimes it failed. It's been passing locally consistently. Thinking tighten that up somehow, so it could be more consistent on ci. I'll tag you once that's done.

benjamincharity (approved):

PR #1958, opened 2020-01-24
feat(Table): now dynamically sizing all columns

### LINKED ISSUE #1805: Table: Create column layout engine
This is required due to the different ways ux/eng want the table to lay out (sometimes filling full width even when content doesn't require it).

#### Requirements

- [x] Table must fill full space no matter the number of columns
  - This is reflected by the table borders and header/summary rows
- [x] Columns must be resizable
- [x] Consumers now must pass in widths for all columns
- [x] ~Scrollbar styles should be added only when content is larger than available space~
  - This is now the consumer's responsibility
- [x] Column resizing should be switched to `width` vs `max-width`

### LINKED ISSUE #1813: Table: Remove column borders
UX wants the column borders removed. Row borders and table borders should remain.

benjamincharity:
Table now dynamically sizes columns.

BREAKING CHANGE:
- TsColumn definition change
- Outer border change
- Style changes

ISSUES CLOSED: #1805
ISSUES CLOSED: #1813 

---

NOTE: Test coverage drops slightly in this release. This is just to get it out asap to unblock feature teams. I will continue working on this post-release until coverage is back up.

---

### `TsColumn`

`TsColumn` definitions now require a width to be set:

```diff
const columns: TsColumn[] = [
  {
    name: 'Column One',
+   width: 100,
  },
];
```

### Outer border

The table no longer defines it's own outer border. See usage docs.

### Persistent scrollbars

The table no longer controls the scrollbars. See usage docs.

### DIFF
--- a/demo/app/components/table/table.component.scss
+++ b/demo/app/components/table/table.component.scss
@@ -1,5 +1,7 @@
 .example-container {
-  height: 400px;
+  //height: 400px;
+  overflow-y: scroll;
+  width: 100%;
 }
 
 .truncate {
--- a/demo/app/components/table/table.component.ts
+++ b/demo/app/components/table/table.component.ts
@@ -7,6 +7,8 @@ import {
   AfterViewInit,
   ChangeDetectorRef,
   Component,
+  OnDestroy,
+  OnInit,
   ViewChild,
 } from '@angular/core';
 import {
@@ -17,6 +19,7 @@ import {
   DomSanitizer,
   SafeHtml,
 } from '@angular/platform-browser';
+import { untilComponentDestroyed } from '@terminus/ngx-tools';
 import {
   TsPaginatorComponent,
   TsPaginatorMenuItem,
@@ -26,7 +29,8 @@ import {
   TsColumn,
   TsTableColumnsChangeEvent,
   TsTableComponent,
-  TsTableDataSource, TsTableDensity,
+  TsTableDataSource,
+  TsTableDensity,
 } from '@terminus/ui/table';
 import {
   merge,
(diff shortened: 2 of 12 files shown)


demo/app/components/table/table.component.scss
```diff
@@ -1,5 +1,7 @@
 .example-container {
-  height: 400px;
+  //height: 400px;
```
atlwendy, line 2: I guess height here isn't needed any more?

demo/app/components/table/table.component.ts
```diff
@@ -6,7 +6,7 @@ import { HttpClient } from '@angular/common/http';
 import {
   AfterViewInit,
   ChangeDetectorRef,
-  Component,
+  Component, OnDestroy, OnInit,
```
atlwendy: Multi line.. but it's on demo, so nbd.

demo/app/components/table/table.component.ts
```diff
@@ -88,63 +89,72 @@ export class ExampleHttpDao {
...
-      name: 'body',
-      width: '260px',
-      control: new FormControl(true),
-    },
+    // {
+    //   display: 'Body',
+    //   name: 'body',
+    //   width: '260px',
+    //   control: new FormControl(true),
+    // },
```
atlwendy, line 138: Does it need to be removed?

terminus-ui/table/src/cell.ts
```diff
@@ -194,108 +220,114 @@ export class TsHeaderCellDirective extends CdkHeaderCell implements AfterViewIni
...
    */
-  private triggerDragging(event: MouseEvent): void {
-    const diff = event.clientX - this.startOffsetX;
-    const newWidth = TsHeaderCellDirective.determineWidth(this.startWidth, diff);
-    this.resized.emit(new TsHeaderCellResizeEvent(this, `${newWidth}px`));
+  private onResizeColumn(event: MouseEvent): void {
+    this.startOffsetX = event.clientX;
+    this.startWidth = this.cellWidth;
+
+    fromEvent<MouseEvent>(document, 'mousemove')
```
atlwendy, line 286: I guess because it's subscribed from an event outside angular zone, this would be still out of the zone as well (so no need to run outside angular any more)

atlwendy (approved):
looks good

demo/app/components/table/table.component.scss
benjamincharity: Just demo styles so if someone wants to see how it'll look with a constrained height they can uncomment it, but I didn't need it for anything on this ticket.

demo/app/components/table/table.component.ts
benjamincharity: eh I'll still fix it 😬

demo/app/components/table/table.component.ts
benjamincharity: I have been leaving it in since it's the only column that has really long content. So if we need to check how really long content effects columns etc we can just uncomment without needing to go see what the GitHub API can return.

PR #2004, opened 2020-02-04
chore: update eslint spec rules

benjamincharity:
- use newer version of our eslint rules package
- disable rules for spec and mock files
- fix a few styling issues
- remove JS files from gitignore
- remove cli flag for eslint ignore in favor of eslint ignore file


.eslintrc.js
```diff
@@ -3,28 +3,42 @@ module.exports = {
...
+      "env": {
+        "jest": true,
+      },
+      "rules": {
+        "dot-notation": "off",
+        "guard-for-in": "off",
+        "line-comment-position": "off",
+        "no-console": "off",
+        "no-magic-numbers": "off",
+        "no-underscore-dangle": "off",
```
atlwendy, line 40: Finally! 👏

.gitignore
```diff
@@ -5,8 +5,6 @@ node_modules/
 out-tsc/
 debug.log
 npm-debug.log
-src/**/*.js
-terminus-ui/**/*.js
 src/**/*.css
```
atlwendy, line 8: should we not ignore generated `js` files at all? (instead of having few exceptions)

atlwendy (approved):

.gitignore
benjamincharity: Isn’t that what this change does? Remove any rules to ignore js files?

atlwendy: Right... I just want to make sure we're intentional on watching ALL js files, including the ones that're generated from typescript..

PR #2016, opened 2020-02-11
Use calc value instead of a magic percentage for selectionlist input width

### LINKED ISSUE #1905: SelectionList: Improve fragile CSS
This CSS could be better written:

https://github.com/GetTerminus/terminus-ui/blob/release/terminus-ui/selection-list/src/selection-list.component.scss#L16-L20

Rather than using a magic percentage, we should use `calc` to subtract the spacing needed for the icon.

atlwendy:
closes #1905

### DIFF
--- a/terminus-ui/selection-list/src/selection-list.component.scss
+++ b/terminus-ui/selection-list/src/selection-list.component.scss
@@ -15,7 +15,9 @@
 
   &.ts-selection-list--single {
     .ts-selection-list__input {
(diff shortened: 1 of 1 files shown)


benjamincharity:
@atlwendy I think this PR includes more than it should...

atlwendy:
oh crap... mis committed..

benjamincharity (approved):

PR #2017, opened 2020-02-11
1963 button css property conversion

### LINKED ISSUE #1963: Button: Convert to css variables
Move as many items into CSS variables as possible to enable feature teams to easily overwrite styles easily.

- Spacing (margin/padding/width/height/etc)
- Color
  - Handle themes by changing variable values rather than generating full sets of classes for each theme.
- Typography
- Animation

NOTE: We should refactor the overall styles to use custom properties rather than overwriting styles for every button type and state.

benjamincharity:
NOTE: The large number of files changed is due to updating sass import paths.

This is the first component to move to custom CSS properties based off of our design tokens package. Very much a proof of concept.

Closes #1963

- Build now includes design tokens
- Button component styles now based on design tokens
- Nested all sass helpers into dirs to cleanup
- Lint our config files
- Shortened sass import paths so that webstorm can correctly find the file
- Removed outdated sass jsdoc items
- Removed empty sass spec files


atlwendy (approved):
I guess the rules you have now in button css are based on the design token conventions, correct?

benjamincharity:
> I guess the rules you have now in button css are based on the design token conventions, correct?

Mostly.. it's still early so I expect as I transition more components over, they will inform structure and naming changes. Still working out the best paths.

atlwendy:
> they will inform structure and naming changes.

Automatically?

benjamincharity:
> Automatically?

hah I wish.. I could just retire at that point.

I just mean that as I convert over more complex components I expect to learn from that process just as I learned from converting the button component. As I find gaps and weak spots in our setup, I'll be adjusting how we define/name/export tokens.

Issue #2025, opened 2020-02-13
CSV Entry Reset Table


atlwendy:
Doesn't seem to consider single column table when resetting.

![1csv](https://user-images.githubusercontent.com/377552/74455140-e2f1be80-4e52-11ea-9511-a6164beeabda.gif)

- [ ] Reseting a full width table should not add extra columns
- [ ] Add test

benjamincharity:
Closed in another PR

PR #2026, opened 2020-02-14
fix(CsvEntry): make container scrollbar not showing up in firefox

### LINKED ISSUE #1689: CSV: Scrollbars issues in Firefox
It seems that in Firefox, the scrollbars are always present. The content may be slightly larger than the window.

When rows are added or removed the scrollbars flash in and out.

atlwendy:
closes #1689 #2025

### DIFF
--- a/terminus-ui/csv-entry/src/csv-entry.component.scss
+++ b/terminus-ui/csv-entry/src/csv-entry.component.scss
@@ -40,6 +40,8 @@
   .c-csv-entry__scroll {
     // Odd height used to ensure that the table cuts off mid-cell so the user knows there is more content
     max-height: 28.5em;
+    // To remove a horizontal scrollbar showing in firefox
+    overflow-x: -moz-hidden-unscrollable;
     overflow-y: auto;
   }
 
--- a/terminus-ui/csv-entry/src/csv-entry.component.ts
+++ b/terminus-ui/csv-entry/src/csv-entry.component.ts
@@ -126,6 +126,7 @@ const DEFAULT_VALIDATION_MESSAGES_MAX = 6;
   exportAs: 'tsCSVEntry',
 })
 export class TsCSVEntryComponent implements OnInit, OnDestroy {
+  private originalColumnCount = DEFAULT_COLUMN_COUNT;
   /**
    * Define the default component ID
    */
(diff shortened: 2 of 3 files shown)


terminus-ui/csv-entry/src/csv-entry.component.ts
```diff
@@ -653,7 +653,6 @@ export class TsCSVEntryComponent implements OnInit, OnDestroy {
   public resetTable(): void {
     this.clearAllRows();
     this.clearHeaderCells();
-    this.columnCount = DEFAULT_COLUMN_COUNT;
```
benjamincharity, line 656: This breaks the use case where a user pastes more columns than the default, and then clicks to reset the table. 

A test should _definitely_ have caught this, so we need to add something in.

terminus-ui/csv-entry/src/csv-entry.component.scss
```diff
@@ -40,6 +40,7 @@
   .c-csv-entry__scroll {
     // Odd height used to ensure that the table cuts off mid-cell so the user knows there is more content
     max-height: 28.5em;
+    overflow-x: -moz-hidden-unscrollable;
```
benjamincharity, line 44: We should probably add a note about why this is needed. Since we don't officially support anything other than chrome I could see this getting deleted down the line.

benjamincharity (changes requested):
These two bug fixes should be different commits so that they both appear in our changelog.

terminus-ui/csv-entry/src/csv-entry.component.ts
atlwendy: I tested it out in demo, it doesn’t break. The reason is that if there’s no columnCount provided, default value is used and if columnCount is provided, that value is used here for reset.
I see resetTable function is covered, so thought there’s no need to add additional tests.

terminus-ui/csv-entry/src/csv-entry.component.scss
atlwendy: Ditto... a comment would definitely be nice. Was adding it but probably got removed when doing clean up (tried bunch of different things). Although I’d say this name is very Firefox specific :)

terminus-ui/csv-entry/src/csv-entry.component.ts
atlwendy: Ah I see your use case, where you can actually expand the column number on fly (via copy/paste). And reset is supposed to reset it to original set up.

benjamincharity (approved):

benjamincharity:
@atlwendy looks like we're not up to date with `release`

PR #2030, opened 2020-02-18
refactor to fix docs generation issue

benjamincharity:


benjamincharity:
@atlwendy just so you know about this issue in case you hit it later.. ^

atlwendy (approved):
Interesting... would not had thought this broke the doc generation.

PR #2034, opened 2020-02-18
fix(CsvEntry): Prefix instance uid for cell ids

### LINKED ISSUE #2021: CSVEntry: Cell IDs need a custom prefix
CSVEntry cells all have custom IDs, but those IDs are the same between table instances. So if two CSVEntry components are used on the same page, IDs conflict

- [ ] Use the component UID (or a shortened version) to prefix column names
  - NOTE: This will likely require tests to be updated since our tests currently are searching for specific IDs

atlwendy:
closes #2021

### DIFF
--- a/terminus-ui/csv-entry/src/csv-entry.component.spec.ts
+++ b/terminus-ui/csv-entry/src/csv-entry.component.spec.ts
@@ -108,6 +108,7 @@ describe(`TsCSVEntryComponent`, function() {
     (event.clipboardData as any) = { getData: jest.fn().mockReturnValue(stringValue) };
     return event;
   };
+  let instancePrefix: string;
   let ENTER_EVENT: KeyboardEvent;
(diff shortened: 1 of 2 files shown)


terminus-ui/csv-entry/src/csv-entry.component.spec.ts
```diff
@@ -279,7 +280,8 @@ describe(`TsCSVEntryComponent`, function() {
     hostComponent = fixture.componentInstance;
     component = hostComponent.component;
     fixture.detectChanges();
-    firstHeaderCell = fixture.debugElement.query(By.css('#r_-1Xc_0')).nativeElement;
+    instancePrefix = fixture.nativeElement.querySelectorAll('[id^="ts-csv-entry"]')[0].getAttribute('id');
```
benjamincharity, line 283: Ah smart!

benjamincharity (approved):

PR #2037, opened 2020-02-21
fix(Input): force the value to a string before sanitization attempt

deanterm:
We should make sure the value is a string type before attempting to sanitize. there is a potential to have primitive values such as a number. even though this isn't expected. this will prevent errors.


benjamincharity (approved):

benjamincharity:
Appreciate the included note 🙏

PR #2053, opened 2020-03-02
chore: Restructure into an angular project

### LINKED ISSUE #2041: Move UI library repo into nested Angular CLI projects
This move is needed so that we can manage both the demos and the regression testing source within our library project.

--- 

Structure should be:

```
dist/
node_modules/
projects/
  demo/
    src/
  visual-regression/
    src/
  library/ 
    autocomplete/
    etc...
```

- [ ] Set up whole new cli project (NOTE: make sure we are using angular 8 not 9)
- [ ] Move ui library files over
- [ ] Move demo project over
- [ ] Add regression project
- [ ] Verify library build works
  - [ ] Verify all paths in gulp and ngpackagr files
- [ ] Verify demo build works
- [ ] Verify regression build works
- [ ] Verify Jest is used for all tests (jasmine/karma is not present)
- [ ] Verify TS paths work

---

**NOTE**

When you create the new cli repo originally, I’d use the flags to **a)** don’t create git repo **b)** don’t make initial commit, **c)** don’t create initial project.

This is better for us since we don’t want a ‘default project’ outside the projects folder. Then you just generate both projects via the ng g application MyProject command so they are all equally nested

See https://angular.io/cli/generate for all flags etc

Then, this is the easiest way to switch to Jest that I’ve found: https://medium.com/angular-in-depth/angular-cli-ng-test-with-jest-in-3-minutes-v2-1060ddd7908d

atlwendy:
closes #2041


benjamincharity (changes requested):
I think we're mainly good to go here, only two small notes:

a) Demo needs to include our design tokens (correct import etc is a few days back in our private channel)
b) We need to rebase to solve the last conflict

benjamincharity:
@atlwendy I also noticed a couple `.DS_store` files that are being tracked that should not be. It looks like we're missing our `.gitignore` file?

angular.json
```diff
@@ -142,64 +190,182 @@
...
+    "library": {
       "projectType": "application",
+      "schematics": {
+        "@schematics/angular:component": {
+          "style": "scss"
+        }
+      },
+      "root": "projects/library",
+      "sourceRoot": "projects/library/src",
+      "prefix": "app",
```
benjamincharity: I believe this should be `ts`

angular.json
```diff
@@ -142,64 +190,182 @@
...
+            }
           }
         }
       }
     },
-    "packages-e2e": {
-      "root": "",
-      "sourceRoot": "",
+    "library": {
       "projectType": "application",
```
benjamincharity: It looks like we may have generated an application project vs a library project?

angular.json
```diff
@@ -142,64 +190,182 @@
...
+            "main": "projects/library/src/main.ts",
+            "polyfills": "projects/library/src/polyfills.ts",
+            "tsConfig": "projects/library/tsconfig.app.json",
+            "aot": false,
+            "assets": [
+              "projects/library/src/favicon.ico",
+              "projects/library/src/assets"
+            ],
+            "styles": [
+              "projects/library/src/styles.scss"
```
benjamincharity: This file doesn't exist.. maybe due to generating an app vs a library?

angular.json
```diff
@@ -142,64 +190,117 @@
...
+            }
           }
         }
       }
     },
-    "packages-e2e": {
-      "root": "",
-      "sourceRoot": "",
-      "projectType": "application",
+    "library": {
```
benjamincharity, line 250: @atlwendy  this is still not the correct config for a CLI library. We should generate a library with the CLI and then move our files into that library project so that we can be sure everything is wired up correctly.

benjamincharity (changes requested):

benjamincharity (approved):

PR #2065, opened 2020-03-06
2046 menu style

benjamincharity:
- fix disabled menu cursor styles
- fix styles for buttons nested inside menu
- convert link and menu styles to use design tokens where possible

### DIFF
--- a/projects/library/button/src/button.component.scss
+++ b/projects/library/button/src/button.component.scss
@@ -92,39 +92,53 @@ ts-button {
     transition: var(--button-color-transition), var(--button-background-transition), var(--button-shadow-transition);
   }
 
+  // TODO: Sass area of possible themes should be available to all components from a global location
+  // Define all possible themes
+  $themes: primary accent warn;
+  // Loop through using each theme name to create styles
+  @each $theme in $themes {
+    // Theme class sets the colors for standard buttons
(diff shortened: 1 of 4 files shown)


projects/library/button/src/button.component.scss
```diff
@@ -92,39 +92,53 @@ ts-button {
     transition: var(--button-color-transition), var(--button-background-transition), var(--button-shadow-transition);
   }
 
+  // TODO: Sass area of possible themes should be available to all components from a global location
+  // Define all possible themes
+  $themes: primary accent warn;
+  // Loop through using each theme name to create styles
+  @each $theme in $themes {
+    // Theme class sets the colors for standard buttons
```
atlwendy, line 100: Similar pattern I saw in chip component as well... is it something that can be refactored later?

atlwendy (approved):

projects/library/button/src/button.component.scss
benjamincharity: Yeah, you'll likely see it in most components that use themes. It's possible we could abstract it out, but I'm not really sure how easy it would be since the theme implementation is very component specific. (ie what elements within a component get styled by themes)

Open to any ideas though 👍

PR #2067, opened 2020-03-06
fix(Table): correctly set any possible column alignment

### LINKED ISSUE #2066: Table: Cell alignment only working for 'right'
As called out here: https://github.com/GetTerminus/terminus-ui/issues/767#issuecomment-595465677

The correct alignment classes are not getting set; only the right alignment works

benjamincharity:
ISSUES CLOSED: #2066

- fix table alignment not being set

### DIFF
--- a/projects/library/table/src/cell.ts
+++ b/projects/library/table/src/cell.ts
@@ -60,7 +60,7 @@ export function setColumnAlignment(column: TsColumnDefDirective, renderer: Rende
         + `See "TsTableColumnAlignment" type for available options.`);
     }
 
-    renderer.addClass(elementRef.nativeElement, 'ts-cell--align-right');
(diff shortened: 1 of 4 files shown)


projects/library/table/src/cell.ts
```diff
@@ -60,7 +60,7 @@ export function setColumnAlignment(column: TsColumnDefDirective, renderer: Rende
         + `See "TsTableColumnAlignment" type for available options.`);
     }
 
-    renderer.addClass(elementRef.nativeElement, 'ts-cell--align-right');
+    renderer.addClass(elementRef.nativeElement, `ts-cell--align-${column.alignment}`);
```
atlwendy, line 63: How did we not see this? lol

atlwendy (approved):

PR #2083, opened 2020-03-19
chore: Run a single E2E test to verify the AoT build loads

### LINKED ISSUE #1212: CI: Run a single E2E test to verify the AoT build loads
We have had two AoT bugs slip through our CI process. This is because those specific items weren't errors at build time, but were errors when the app was actually loaded. We should use an E2E framework to boot up the app and verify it doesn't error.

- [ ] Add single E2E test to the CI flow
  - eg Load page and check for text.

atlwendy:
closes #1212


.circleci/config.yml
```diff
@@ -69,6 +73,11 @@ jobs:
           key: yarn-packages-{{ checksum "yarn.lock" }}
           paths:
             - ~/.cache/yarn
+      - save_cache:
+          name: Save Cypress Cache
+          key: cache-{{ checksum "package.json" }}
```
benjamincharity: Pretty sure the checksum should be based off the yarn.lock file instead?

.circleci/config.yml
```diff
@@ -51,10 +54,11 @@ jobs:
       - restore_cache:
           name: Restore Yarn Cache
           keys:
-            - terminusui-yarn-deps-{{ .Branch }}-{{ checksum "yarn.lock" }}
-            - terminusui-yarn-deps-{{ .Branch }}
-            - terminusui-yarn-deps-release
-            - terminusui-yarn-deps-
+            - yarn-packages-{{ .Branch }}-{{ checksum "yarn.lock" }}
```
benjamincharity, line 56: With this change, have you started to see Circle finding the yarn cache?

.circleci/config.yml
```diff
@@ -13,6 +15,7 @@ references:
     working_directory: ~/ci-build
     docker:
       - image: circleci/node:12.16.1
+      - image: cypress/base:10
```
benjamincharity: I thought the purpose of orbs was so we don't need a custom docker image for everything? Is that not the case?

.circleci/config.yml
```diff
@@ -109,6 +118,19 @@ jobs:
...
           paths:
             - ./dist/library/*
+  integration_test:
+    <<: *defaults
+    steps:
+      - *attach_workspace
+      - restore_cache:
+          name: Restore Cypress Cache
+          keys:
+            - cache-{{ checksum "package.json" }}
```
benjamincharity: Does this need fallback paths like the yarn cache restore?

.circleci/config.yml
```diff
@@ -183,6 +205,9 @@ workflows:
       - build:
           requires:
             - setup
+      - integration_test:
```
benjamincharity, line 208: The publish demo step should rely on this step so that if integration test fails publishing is blocked.

cypress.json
```diff
@@ -0,0 +1 @@
+{}
```
benjamincharity, line 1: Is this really required?

cypress/plugins/index.js
```diff
@@ -0,0 +1,21 @@
+/// <reference types="cypress" />
```
benjamincharity, line 1: Seems to be three files here with only comments or empty exports; can we remove them?

cypress/integration/spin-up.js
```diff
@@ -0,0 +1,13 @@
+describe(`Demo`, () => {
+  it(`should load demo page`, () => {
```
benjamincharity, line 2: I'm guessing Cypress requires `it` instead of `test`? 😞

tooling/ci/install-required-packages-for-integration.sh
```diff
@@ -0,0 +1,18 @@
...
+
+#
+#
+# Install all the necessary packages for spinning up a server
+#
+
+. ~/.bashrc
+
+sudo apt-get install libgtk2.0-0 libgtk-3-0 libnotify-dev libgconf-2-4 libnss3 libxss1 libasound2 libxtst6 xauth xvfb
+npm install --save-dev start-server-and-test
```
benjamincharity, line 12: Any reason we're using `npm` here vs `yarn`?

tooling/ci/install-required-packages-for-integration.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+
+#
+#
+# Install all the necessary packages for spinning up a server
+#
+
+. ~/.bashrc
+
+sudo apt-get install libgtk2.0-0 libgtk-3-0 libnotify-dev libgconf-2-4 libnss3 libxss1 libasound2 libxtst6 xauth xvfb
```
benjamincharity, line 10: ~Curious why we have to install all this stuff manually if we're using a Docker image specifically for Cypress?~

Curious why we have to install all this stuff manually if we're using an orb specifically for Cypress?

benjamincharity (changes requested):
Excited for this one! Marking as changes requested, but most of my comments are questions so may not need changes.

.circleci/config.yml
atlwendy: Yes. cache was found with the change.

.circleci/config.yml
atlwendy: yep. docker is not needed any more. It was added during experimental stage trying to make things work..

.circleci/config.yml
atlwendy: It could have fallback path, but it doesn't have to have one. I'm adding it..

cypress/plugins/index.js
atlwendy: All these are most placeholders for more to come. It's currently no new content added given we're running a super simple test. I'd like to keep them cause when we add like regression test, these files would need to be updated.

.circleci/config.yml
benjamincharity: Nice!

cypress/plugins/index.js
benjamincharity: That's good with me 👍 (I was just having flashbacks to all those sass spec files that were added, never used, and finally removed 😂 )

tooling/ci/install-required-packages-for-integration.sh
atlwendy: the reason is that `yarn` doesn't work here..  saw some post online reported the similar issues..

benjamincharity: Boooooo ;p 

We should make a note in that file probably so no one tries to refactor it in a few months

tooling/ci/install-required-packages-for-integration.sh
atlwendy: This I actually didn't find much reference. If I don't have these added, ci complains about it. Theoretically it should have been included by orb. I guess I could dig more into it, but not sure whether it'd be worth the effort..

benjamincharity: Definitely not worth blocking this

cypress.json
atlwendy: A placeholder as well...

benjamincharity (approved):

atlwendy:
This is out of dated. #2129 replaces this.

PR #2087, opened 2020-03-25
chore(Paginator): Switch select with selection list and add no validation mode

### LINKED ISSUE #2032: Paginator: convert select to selectionlist and enable no validation mode
- [ ] Paginator is using old TsSelect control, this should be converted to a SelectionList
- [ ] Paginator has too much space below the controls for validation. We should enable no validation mode for the control.

atlwendy:
closes #2032


benjamincharity (changes requested):
Looks good. Just need a rebase 👍

benjamincharity (approved):

PR #2129, opened 2020-04-08
chore(test): Run a single E2E test to verify the AoT build loads

### LINKED ISSUE #1212: CI: Run a single E2E test to verify the AoT build loads

atlwendy:
closes #1212


.github/workflows/ci-push.yml
```diff
@@ -192,3 +192,51 @@ jobs:
...
+        id: cache_node_modules
+        uses: actions/cache@v1
+        with:
+          path: node_modules
+          key: ${{ runner.os }}-node-${{ hashFiles('**/yarn.lock') }}
+        if: env.SHOULD_RUN == 'true'
+      - if: env.SHOULD_RUN == 'true' && steps.cache_node_modules.outputs.cache-hit != 'true'
+        run: yarn install --frozen-lockfile --non-interactive
+      - if: env.SHOULD_RUN == 'true'
+        run: yarn add start-server-and-test
```
benjamincharity, line 227: Any reason we are installing here and not adding as a devDependency?

lint-staged.config.js
```diff
@@ -46,16 +47,14 @@ module.exports = {
...
  */
-function removeDirectories(files) {
-  const match = micromatch.not(files, [
-    ...IGNORE,
-    '**/testing/**',
-    '**/demo/**',
-  ]);
-  return match.join(' ');
-}
+const removeDirectories = files => micromatch.not(files, [
```
benjamincharity, line 55: Why were changes needed here?

benjamincharity (changes requested):
Couple small questions inline.

One larger question here: Do you think we should wait on using Cypress for the release flow until we have a better handle on it? I'm worried that it will kill our automated releases and I'll have to sort through partially released versions. Thoughts?

.github/workflows/ci-push.yml
atlwendy: I thought about that, but this is more just for starting server on ci... no other place it'd be needed. I'm okay either way if you prefer one over the other...

lint-staged.config.js
atlwendy: It was returning string, but I'd want it to be array, so can directly plug into the function you defined where it targets all ts and js files. And it doesn't seem it was referenced anywhere else?

benjamincharity (approved):

PR #2162, opened 2020-04-15
chore(Demo): update page title on route change

### LINKED ISSUE #1903: Demos: Should update the window title on page change
Currently the window title does not change when navigating around the demos. A better experience would be to update the window title on navigation. See https://angular.io/guide/set-document-title

- [ ] Update the window title on navigation

benjamincharity:
Closes #1903

### DIFF
--- a/projects/demo/src/app/app.component.ts
+++ b/projects/demo/src/app/app.component.ts
@@ -1,4 +1,19 @@
-import { Component } from '@angular/core';
+import {
+  Component,
+  OnDestroy,
+  OnInit,
+} from '@angular/core';
+import { Title } from '@angular/platform-browser';
+import {
+  ActivatedRoute,
+  NavigationEnd,
+  Router,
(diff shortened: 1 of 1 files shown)


projects/demo/src/app/app.component.ts
```diff
@@ -1,10 +1,51 @@
...
+    this.router.events.pipe(
+      untilComponentDestroyed(this),
+      filter(event => event instanceof NavigationEnd),
+      map(() => {
+        let child = this.activatedRoute.firstChild;
+        while (child?.firstChild) {
+          child = child.firstChild;
+        }
+        if (child?.snapshot.data.name) {
+          return child.snapshot.data.name;
```
atlwendy, line 41: Didn't know there's `snapshot` attribute from `activatedRoute`...

atlwendy (approved):

PR #2174, opened 2020-04-20
First batch of regression tests

### LINKED ISSUE #2092: Visual regression for first batch of components
First batch of components is, Button Input Selectionlist and Card.

atlwendy:
closes #2092


.github/workflows/ci-pr.yml
```diff
@@ -0,0 +1,113 @@
+name: CI Pull Request
+on: pull_request
```
benjamincharity, line 2: Should this run for all pr events? Or just some? ie create and edit vs review added etc

.github/workflows/ci-pr.yml
```diff
@@ -0,0 +1,113 @@
...
+        with:
+          path: node_modules
+          key: ${{ runner.os }}-node-${{ hashFiles('**/yarn.lock') }}
+        if: env.SHOULD_RUN == 'true'
+      - if: env.SHOULD_RUN == 'true' && steps.cache_node_modules.outputs.cache-hit != 'true'
+        run: yarn install --frozen-lockfile --non-interactive
+      - if: env.SHOULD_RUN == 'true'
+        run: yarn add start-server-and-test
+      - if: env.SHOULD_RUN == 'true'
+        run: export PERCY_TOKEN=fb4bc79a6546164e53d208391f04b7fd84fb0afb0673fa47d285b36565b61d8c
```
benjamincharity: I'd guess this should an env variable so everyone doesn't know our key

cypress/integration/visual-regression/button.js
```diff
@@ -0,0 +1,21 @@
...
+  beforeEach(() => {
+    cy.visit('http://localhost:4200/components/button');
+  });
+
+  describe(`button`, () => {
+    it(`should load the page`, () => {
+      cy.get('body').contains('Themes');
+      cy.get('#ts-button-id-1 > .mat-button-wrapper > .c-button__content').contains('Primary');
+      cy.get('.c-button__content > .ts-icon > .c-icon').click();
+      cy.percySnapshot('button overview');
```
benjamincharity, line 11: Making sure I understand what's happening.. `cy.get` seems like a standard expect. Guessing `cy.percySnapshot` is basically checking a current snapshot against a previous via ID?

If that is the case, are there benefits to using one over the other or both at the same time? ie can the first 3 checks ever not be true if the last _is_ true?

cypress/integration/visual-regression/card.js
```diff
@@ -0,0 +1,56 @@
...
+
+  describe(`border and color`, () => {
+    it(`should set border color accordingly`, () => {
+      cy.get(':nth-child(5) > .ng-valid').select('bottom');
+      cy.get('#ts-card-0').should('have.class', 'c-card--border-bottom');
+      cy.percySnapshot('bottom-border');
+    });
+
+    it(`should set right color`, () => {
+      cy.get(':nth-child(4) > .ng-valid').select('warn');
```
benjamincharity: This type of selector seems fragile.. if we added another instance in place #2 would it break all these references?

package.json
```diff
@@ -66,6 +66,8 @@
     "vr:test": "jest --config jest.visual-regression.config.js",
     "vr:test:ci": "yarn run vr:test --runInBand --coverage",
     "vr:test:ci:local": "yarn run vr:test --coverage",
+    "vr:test:regression": "yarn start-server-and-test vr:start http://localhost:4200 vr:percy:run",
+    "vr:percy:run": "export PERCY_TOKEN=fb4bc79a6546164e53d208391f04b7fd84fb0afb0673fa47d285b36565b61d8c && percy exec -- cypress run --spec 'cypress/integration/visual-regression/*.js'",
```
benjamincharity: This token should probably not be exposed to the world.. Is this here to allow local runs?

cypress/integration/visual-regression/card.js
benjamincharity: Would we be able to add IDs to test components?

projects/visual-regression/src/app/components/input/input.component.html
```diff
@@ -0,0 +1,96 @@
...
+
+  <ts-input
+    [(ngModel)]="textareaModel"
+    label="My textarea"
+    name="My textarea"
+    hint="So much space! How will you use it?"
+    [isTextarea]="true"
+  ></ts-input>
+
+  <strong>ngModel value:</strong>
```
benjamincharity, line 76: Should our visual tests be checking model values?

projects/visual-regression/src/app/components/selection-list/selection-list.component.html
```diff
@@ -0,0 +1,111 @@
+
+  <div class="page">
```
benjamincharity: Indention is off a bit here for a few lines

projects/visual-regression/src/app/components/selection-list/selection-list.component.ts
```diff
@@ -0,0 +1,170 @@
...
+    } else if (demo === 'simple') {
+      this.simpleQuery$.next(query);
+    }
+  }
+
+  public formatter(value: State): string {
+    return value.name;
+  }
+
+  public duplicate(e: TsSelectionListChange, demo: DEMOS): void {
```
benjamincharity: Do we need all these log functions?

.github/workflows/ci-pr.yml
atlwendy: ah meant to have it run on open pr (create) as well as code push. Didn't think adding review would trigger this too. Will be more specific.

projects/visual-regression/src/styles.scss
```diff
@@ -1 +1,27 @@
-@import '~@terminus/design-tokens/css/library-design-tokens.css';/* You can add global styles to this file, and also import other style files */
+@import '~@terminus/ui/helpers';
+@import '~@terminus/ui/terminus-ui.css';
+@import '~@terminus/design-tokens/css/library-design-tokens.css';
+
+html {
+  background-color: color(pure);
+  font-family: $type__family--base;
```
benjamincharity, line 7: Don't forget you can use our CSS custom property tokens for this too ;p (def not a needed change)

projects/visual-regression/src/styles.scss
```diff
@@ -1 +1,27 @@
...
+  margin: 0;
+}
+
+hr,
+.page {
+  margin: 1rem;
+}
+
+
+// stylelint-disable-next-line selector-type-no-unknown, scss/selector-nest-combinators
```
benjamincharity: Looks like VR app already has it's own stylelint config, so feel free to disable some rules like this that we may not care about in apps like this.

projects/visual-regression/src/tsconfig.app.json
```diff
@@ -0,0 +1,12 @@
+{
```
benjamincharity, line 1: Curious why this was needed? Seems like it's one folder too deep and almost matches the one already existing?

benjamincharity (changes requested):
Well, I can leave one-off comments _until_ I make my first review comment.. so many of my questions get lumped into the review ¯\_(ツ)_/¯

Another question: Does cypress have anything built in to help only run tests on modules that were changed?

Most of my questions are due to my unfamiliarity with Cypress. It may be easier to just jump on a call and walk through some of these.

.github/workflows/ci-pr.yml
benjamincharity: Yeah I was surprised at how many events they do offer for PRs. Definitely useful for other stuff like adding/removing labels etc

projects/visual-regression/src/tsconfig.app.json
atlwendy: Hmm.. I'm not sure actually. It was there in `demo/src` folder, so it was copied to visual regression project folder.

cypress/integration/visual-regression/button.js
atlwendy: `cy.get().contains()` is standard expect. For cypress, if one expect fails, the rest of expects in the same block gets skipped (same as jest). So the point is that, if some simple check fails, there's no need to take a snapshot and compare them as they're pretty much guaranteed to be different.

cypress/integration/visual-regression/card.js
atlwendy: 👍 yeah thought about that as todo as I was trying to get the functionality working. Those selectors are from cypress - they output a selector if you point anywhere on the page.

projects/visual-regression/src/tsconfig.app.json
benjamincharity: Ohh I bet that was some cleanup we missed during The Big Migration™️

projects/visual-regression/src/app/components/input/input.component.html
atlwendy: Do you mean the actual model value or what's displayed on UI? It's already checking UI value..

benjamincharity: Meaning, I'm not sure why `visual regression` tests are checking model values. That seems like something our standard tests should cover leaving only visual items for these tests.

atlwendy: Ah, wasn't trying to test the actual model values. Was just thinking presentation in UI might be compared (updated versus benchmark). Not an important piece and certainly can move out.

atlwendy:
> Another question: Does cypress have anything built in to help only run tests on modules that were changed?

Cypress doesn't provide that out of the box. I dug a little and it seems to be doable by writing a couple scripts to make it only run on changed modules. But the possible problem that I see is that if for example, there are 20 images stored in percy and this run contains only 10 images, percy will report build failed because of those missing images. Well I think we could always go in and mark them as "approved". It just seems less automated.

.github/workflows/ci-pr.yml
benjamincharity: Are we leaving this to trigger on all events for now?

atlwendy: By default, it's triggered on `created`, `synchronize` and `reopened` events and that's what we want, right?

>By default, a workflow only runs when a pull_request's activity type is opened, synchronize, or reopened. To trigger workflows for more activity types, use the types keyword.

atlwendy: I guess I could make it more explicit by spelling out all those events..

projects/visual-regression/src/app/components/input/input.component.html
benjamincharity: yeah, imo it's extra to look at and doesn't add any value

benjamincharity (approved):
Just need a rebase and I think we're good

.github/workflows/ci-pr.yml
benjamincharity: ah I see yeah I forgot those are the exact events here. Disregard!

PR #2184, opened 2020-04-27
feat(Table): support footer rows

benjamincharity:
Although we were originally asked specifically to not implement support for a footer row, Retargeting now needs it.

Here is that support.

NOTE: Once this lands in v17 I will pluck the commits over to `release`.

Screen Shot 2020-04-27 at 3 54 21 PM

### DIFF
--- a/projects/library/table/src/table.component.spec.ts
+++ b/projects/library/table/src/table.component.spec.ts
@@ -18,6 +18,10 @@ import {
   getTableInstance,
   TestData,
 } from '@terminus/ui/table/testing';
+import {
+  getFooterCells,
+  getFooterRows,
+} from '../testing/src/test-helpers';
 
 import {
   TsCellDirective,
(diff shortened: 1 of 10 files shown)


projects/library/table/src/table.component.spec.ts
```diff
@@ -18,6 +18,9 @@ import {
   getTableInstance,
   TestData,
 } from '@terminus/ui/table/testing';
+import {
+  getFooterCells, getFooterRows,
+} from '../testing/src/test-helpers';
```
atlwendy, line 24: err... separate lines?

atlwendy (approved):
looks good

projects/library/table/src/table.component.spec.ts
benjamincharity: Yup, I realized I missed the alignment class so I'll fix this when I push that up.

Issue #2182, opened 2020-04-23
Search: Should support noValidationOrHint mode


benjamincharity:
As input does.

- [ ] Pass flag through to form field

PR #2195, opened 2020-04-30
search component

### LINKED ISSUE #2181: Search: Button alignment is wrong
- [ ] Fix alignment
- [ ] Add simple visual regression test

atlwendy:
Search button lines up with search input box
Support noValidationOrHint mode
Convert out of dated specs to integration tests
closes #2181 #2182

### DIFF
--- a/projects/library/search/src/search.component.html
+++ b/projects/library/search/src/search.component.html
@@ -18,6 +18,7 @@
     [hint]="inputHint"
     [isFocused]="isFocused"
     name="query"
+    [noValidationOrHint]="noValidationOrHint"
     pattern="{{ inputPatternRegex }}"
     [validateOnChange]="autoSubmit"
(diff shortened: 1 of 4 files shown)


projects/library/search/src/search.component.html
```diff
@@ -13,10 +13,12 @@
     [label]="inputLabel"
     [prefixIcon]="icon"
     [isClearable]="userCanClear"
+    [isDisabled]="isDisabled"
```
benjamincharity, line 16: heh literally just hit this 2 days ago 🥳

benjamincharity (approved):

projects/library/search/src/search.component.html
atlwendy: It's funny that how the old specs all passed, but this line was never added in the template 😂

PR #2217, opened 2020-05-04
Add custom reset to csv demo

benjamincharity:


projects/demo/src/app/components/csv-entry/csv-entry.component.ts
```diff
@@ -1,5 +1,6 @@
-import { Component } from '@angular/core';
+import { Component, ViewChild } from '@angular/core';
```
atlwendy, line 1: ..... just in demo though

projects/demo/src/app/components/csv-entry/csv-entry.component.ts
```diff
@@ -17,6 +18,12 @@ export class CSVEntryComponent {
   public myFile;
   public footerDirection: 'ltr' | 'rtl' = 'ltr';
 
+  @ViewChild(TsCSVEntryComponent, {
+    static: true,
+    read: false,
```
atlwendy, line 23: Haven't used this `read` option yet....

atlwendy (approved):

projects/demo/src/app/components/csv-entry/csv-entry.component.ts
benjamincharity: Yeah when we got rid of TSLint I relaxed some style-based rules in the demo and VR apps

projects/demo/src/app/components/csv-entry/csv-entry.component.ts
benjamincharity: I haven't really needed it before, but it's what hub is using and they are saying this doesn't work, so creating a reproduction they can see

PR #2222, opened 2020-05-05
chore(Select): add custom class to select overlay panel

IraErshova:
Add a custom class to select overlay panel


benjamincharity:
Closing as I believe this is no longer needed.

PR #2249, opened 2020-05-28
chore(Popover): upgrade popperjs to v2

### LINKED ISSUE #2227: Packages: Update popper
Looks like v2 has dropped:

https://popper.js.org/docs/v2/migration-guide/

atlwendy:
closes #2227

### DIFF
--- a/package.json
+++ b/package.json
@@ -138,7 +138,7 @@
     "@terminus/ui": "18.2.2",
     "date-fns": "2.12.0",
     "ngx-perfect-scrollbar": "9.0.0",
-    "popper.js": "1.16.1",
+    "@popperjs/core": "^2.0.0",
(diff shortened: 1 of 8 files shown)


benjamincharity (changes requested):
Looks good 👍 looks like we just need to bump the popper range in the library `projects/library/package.json` to `"popper.js": "^2.0.0"`

benjamincharity (changes requested):
The wrong package.json was changed.

The root `package.json` is for our repo and the UI demo application - here we use fixed versions.

The `package.json` file located here: `projects/library/package.json` is the `package.json` that gets shipped with the library. Here we want to use larger ranges so that consumers can update at their own pace.

benjamincharity (approved):

